PR #643, opened 2026-04-02
feat: multilingual epistemic structure analysis

MinhHaDuong:
## Summary

- Add `scripts/analyze_multilingual.py` — Phase 2 analysis script for the multilingual research note
- Computes: language stats (T1), quadrant classification (T2), language×cluster contingency + χ² (T3), semantic isolation scores at k=5/10/20 (T4), citation directionality (T6), core subset composition
- Citation directionality uses vectorized pandas map instead of iterrows (~970K edges)
- No sys.path hacks — uses direct sibling imports via `uv run`
- 17 unit tests covering all helper functions
- Makefile rule: `content/tables/multilingual_report.json`

## Test plan

- [x] `make check-fast` passes (1 flaky pre-existing failure in parallel, passes solo)
- [x] 17/17 unit tests pass
- [ ] Run `make content/tables/multilingual_report.json` on real corpus data

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -458,6 +458,11 @@ content/tables/tab_lexical_tfidf.csv: scripts/compute_lexical.py scripts/utils.p
 		content/tables/tab_breakpoint_robustness.csv
 	uv run python $< --output $@
 
+# Multilingual epistemic structure (exploratory JSON report)
+content/tables/multilingual_report.json: scripts/analyze_multilingual.py scripts/utils.py \
+		scripts/build_het_core.py $(REFINED) $(REFINED_EMB) $(REFINED_CIT) $(SEMANTIC_CLUSTERS)
+	uv run python $< --output $@
+
 # K-sensitivity table
 content/tables/tab_k_sensitivity.csv: scripts/compute_breakpoints.py scripts/utils.py $(CONFIG) $(REFINED)
 	uv run python $< --output $@ --k-sensitivity
--- a/scripts/analyze_multilingual.py
+++ b/scripts/analyze_multilingual.py
@@ -36,7 +36,7 @@ def classify_quadrant(row):
     if not lang:
         return None
 
-    is_en = lang in ("en", "eng", "en_us", "english")
+    is_en = not is_non_english(row)
     aff = str(row.get("affiliations", "") or "").lower().strip()
     if not aff:
         return None  # Cannot determine geography without affiliations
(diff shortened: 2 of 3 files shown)


MinhHaDuong (commented):
## Multi-perspective review (Correctness + Consistency + Doc propagation)

### Tests: `make check-fast` passes (710/710), `make check` shows 27 pre-existing failures unrelated to this PR. All 18 new tests pass.

---

### [Correctness] `compute_contingency` has zero test coverage — request-changes (HIGH confidence)

This is the only public function without a unit test. It involves a merge on `doi`, a chi-squared test, and standardized residuals — all with potential edge cases (empty merge, single-row contingency table, NaN languages). Every other function has at least one test; this one needs the same treatment.

Suggested: add `TestComputeContingency` with at least (a) a basic happy-path test and (b) a test with NaN languages to verify the "other"/"missing" grouping.

---

### [Correctness] `classify_quadrant` duplicates English-detection logic — comment (MEDIUM confidence)

Line 39 hardcodes `("en", "eng", "en_us", "english")` to detect English. `build_het_core.is_non_english()` uses the same set. If either changes, they will drift silently.

Suggestion: replace line 39 with `is_en = not is_non_english(row)`, which you already import. This also avoids the special-case for `lang == ""` since `is_non_english` returns False for empty languages.

Actually, reviewing more carefully: `is_non_english` returns False for empty language, meaning empty language would be classified as English. The current code returns None for empty language (line 36-37), which is the correct behavior. So the refactor would need to keep the early return for empty language. Still worth considering to avoid the duplicated set.

---

### [Correctness] `compute_contingency` maps NaN languages to "other" instead of "missing" — comment (MEDIUM confidence)

In `compute_language_stats`, NaN languages become `"missing"`. In `compute_contingency`, the lambda on line 92 maps NaN to `"other"` (because `NaN not in ("en", ...)` is True), then `fillna("missing")` on line 94 is a no-op since the value is already `"other"`. This means the two functions label NaN-language works differently. Consider whether "missing" should be a separate group in the contingency table too.

---

### [Correctness] `compute_quadrant_stats` zero-division edge case — comment (MEDIUM confidence)

If all rows lack language or affiliations, `classified` is empty and `len(subset) / len(classified)` raises ZeroDivisionError. Also, `.mean()` / `.median()` on an empty Series returns NaN, which becomes `NaN` in JSON output (technically invalid JSON — `json.dump` with `default=str` would write `"nan"`). Won't trigger on the real corpus, but the function should handle it for test robustness.

---

### [Consistency] EN-N baseline uses full sort vs. partition for other quadrants — comment (LOW confidence)

`compute_isolation_scores` uses `np.partition` (O(n)) for non-EN quadrants but `np.sort` (O(n log n)) for the EN-N baseline. The sort is needed to skip self-distance, so this is correct, just a performance inconsistency worth noting. Not blocking.

---

### [Doc propagation] No issues found — approve

Makefile paths match Python paths. No manuscript, config, or architecture doc changes needed. The research note already documents this analysis.

---

### Summary

One blocking finding: `compute_contingency` needs test coverage. Several non-blocking comments on edge cases and code duplication. Overall the script is well-structured, follows project conventions (Phase 2 naming, I/O discipline, logging, no sys.path hacks), and the Makefile rule is correct.

Risk label: `review:standard`

PR #645, opened 2026-04-13
Fix null-DOI cartesian explosion in multilingual analysis

MinhHaDuong:
## Summary
- Drop null DOIs before merge(on=doi) in compute_contingency — 7678 x 7921 null matches produced a 60M-row cartesian product (69GB PyArrow allocation), OOM-killing padme twice.
- Drop null DOIs before set_index(doi) in compute_citation_directionality — duplicate null index entries crashed Series.map().

## Root cause
refined_works.csv has 7678 rows without DOIs; semantic_clusters.csv has 7921. Pandas merge(on=doi, how=inner) matches null == null, creating NxM rows.

## Test plan
- [x] Script completes on padme in 13s, peak 1.9GB RSS (was OOM at 105GB)
- [x] Report generated with all expected keys
- [ ] make check-fast passes

### DIFF
--- a/scripts/analyze_multilingual.py
+++ b/scripts/analyze_multilingual.py
@@ -85,8 +85,11 @@ def compute_quadrant_stats(df):
 
 def compute_contingency(df, clusters_df):
     """Table T3: language x cluster contingency table + chi-squared test."""
-    merged = df.merge(
-        clusters_df[["doi", "semantic_cluster"]],
+    # Drop null DOIs before merge to avoid cartesian product (~8K × 8K = 60M rows)
+    df_with_doi = df.dropna(subset=["doi"])
+    clusters_with_doi = clusters_df.dropna(subset=["doi"])
+    merged = df_with_doi.merge(
+        clusters_with_doi[["doi", "semantic_cluster"]],
         on="doi",
         how="inner",
     )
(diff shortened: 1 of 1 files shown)


MinhHaDuong (commented):
## Multi-perspective review

**Risk level: Standard** — single-file bug fix in data processing script.
**Label: `review:trivial`** — minimal, well-scoped fix.

### Correctness (high confidence): APPROVE

Root cause accurately diagnosed: `pd.merge(on="doi", how="inner")` matches `NaN == NaN`, producing a ~60M-row cartesian product from ~7.7K x ~7.9K null-DOI rows. Both fixes are correct and minimal:

1. `compute_contingency`: `dropna(subset=["doi"])` before merge — rows without DOIs cannot match cluster assignments anyway, so no valid data is lost.
2. `compute_citation_directionality`: adding `"doi"` to `dropna(subset=["geo", "doi"])` before `set_index("doi")` prevents duplicate null index entries that break `Series.map()`.

All 21 unit tests pass on the branch.

### Consistency (high confidence): APPROVE

- `dropna` pattern matches existing usage in the same file (line 69).
- Variable naming (`df_with_doi`, `clusters_with_doi`) is clear.
- Inline comment explains the "why" with concrete numbers.
- No scope creep — one file, two functions, exactly the affected code.

### Non-blocking recommendation

The existing test fixtures use only non-null DOIs. A follow-up test case with `"doi": None` in the `TestComputeContingency` and `TestComputeCitationDirectionality` fixtures would guard against regression.

---
Reviewed by: Correctness + Consistency agents

PR #646, opened 2026-04-13
fix: OOM-safe isolation scores + null-DOI guards

MinhHaDuong:
## Summary

- Replace full pairwise cosine distance matrix with `NearestNeighbors` KNN in `compute_isolation_scores()` — avoids OOM on 30K+ embeddings
- Drop null DOIs before merge in `compute_contingency()` and `compute_citation_directionality()` to prevent cartesian products
- Add RSS memory logging (`_log_mem`) at each pipeline step for future OOM diagnosis

## Test plan

- [x] `ruff check` passes
- [x] 719/720 tests pass (1 pre-existing failure: context budget)
- [ ] `make output/multilingual_report.json` completes without OOM

### DIFF
--- a/scripts/analyze_multilingual.py
+++ b/scripts/analyze_multilingual.py
@@ -12,15 +12,6 @@ import resource
 
 import numpy as np
 import pandas as pd
-from scipy import stats
-from sklearn.neighbors import NearestNeighbors
-
-
-def _log_mem(log, label):
-    """Log current RSS in MB."""
-    rss = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024
-    log.info("[mem] %s: %.0f MB RSS", label, rss)
-
 from build_het_core import is_global_south, is_non_english
 from pipeline_loaders import (
     CATALOGS_DIR,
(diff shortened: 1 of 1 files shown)


MinhHaDuong (commented):
## Review: APPROVE (Correctness + Consistency + Scope)

### Correctness (high confidence)
- **OOM fix**: `NearestNeighbors(metric="cosine", algorithm="brute")` correctly replaces the full pairwise `cosine_distances` matrix. KNN returns only k distances per query — mathematically equivalent to the original partition-based selection, with O(k) memory per row instead of O(N_en_north).
- **Null-DOI guards**: Correct in both `compute_contingency` and `compute_citation_directionality`. Prevents NaN DOIs from causing unexpected merge/index behavior.
- **Memory logging**: `resource.getrusage(RUSAGE_SELF).ru_maxrss / 1024` gives MB on Linux. Note: `ru_maxrss` is **peak** RSS (high-water mark), so values only increase. Consider renaming to "peak RSS" in the log message for clarity.

### Consistency (high confidence)
- Import reordering follows isort/ruff conventions. `ruff check` passes.
- Formatting changes are mechanical (Black-compatible ternary/lambda rewraps).
- `_log_mem` correctly uses the module-level `log` — consistent with rest of script.
- Good refactor: EN-N baseline now uses `_summarize_scores` like the other quadrants.

### Scope
Tightly scoped: OOM fix + null guards + diagnostics. No unrelated changes.

### Minor observations (non-blocking)
1. `compute_isolation_scores` is called 3x (k=5,10,20), rebuilding KNN on the same `en_north_emb` each time. With `algorithm="brute"` this is negligible, but a future optimization could compute all k values in one pass.
2. No test covers the null-DOI path directly (NaN DOIs in contingency/citation input). Consider adding if coverage is a goal.
3. `ru_maxrss` reports peak RSS, not current — the log label "[mem] ... RSS" could say "peak RSS" for accuracy.

All 21 tests pass. Ruff clean.

---
Review by: Correctness, Consistency, Scope agents

PR #654, opened 2026-04-15
feat: GPU acceleration for semantic divergence S1-S4

### LINKED ISSUE #37: Polish Figure 2 cluster labels for submission
## Context
`plot_fig2_composition.py` now loads cluster labels dynamically from `cluster_labels.json` (TF-IDF distinctive terms). Current labels:

- 0: Green, bonds & financing
- 1: Funds, funding & private
- 2: Trading, forest & co2
- 3: Forest, water & land
- 4: Cdm, energy & clean
- 5: Action, impacts & agreement

## Task
Review and polish labels for readability in an HPS journal context:
- Consider replacing raw TF-IDF terms with interpretive names (e.g., "CDM & clean energy" instead of "Cdm, energy & clean")
- Ensure labels are self-explanatory for non-quantitative readers
- Options: (a) edit cluster_labels.json directly, (b) add a human-readable override layer in the script

Depends on: #36 (labels may change after abstract enrichment)

MinhHaDuong:
## Summary
- Add torch CUDA backend for O(n²) semantic divergence methods (S1 MMD, S2 energy, S3 Wasserstein, S4 Fréchet)
- Config-driven dispatch (`backend: auto|cpu|cuda` in `config/analysis.yaml`) — auto-detects CUDA, falls back to NumPy
- New `scripts/_divergence_backend.py` module for device management
- Also unblocks ticket 0042 from 0037/0039 — CPU run is fine for first real data pass

## Files changed
- `config/analysis.yaml` — added `backend: auto` setting
- `scripts/_divergence_backend.py` — new backend dispatch module
- `scripts/_divergence_semantic.py` — torch implementations + dispatch in S1-S4
- `tests/test_gpu_backend.py` — 8 tests: dispatch logic + torch/numpy equivalence
- `tickets/0042-run-real-corpus.erg` — removed stale blockers

## Test plan
- [x] 8/8 GPU backend tests pass (dispatch + equivalence for all 4 methods)
- [x] `make check-fast` — no new failures (14 pre-existing on main)
- [ ] Benchmark on real data (ticket 0042) to confirm speedup

Closes #0037

### DIFF
--- a/config/analysis.yaml
+++ b/config/analysis.yaml
@@ -21,6 +21,7 @@ clustering:
 
 divergence:
   random_seed: 42
+  backend: auto        # auto (GPU if available), cpu, cuda
   windows: [2, 3, 4, 5]
   min_papers: 30          # per window minimum (override with 5 for smoke)
   min_papers_smoke: 5     # auto-detected when n_works < 200
--- a/scripts/_divergence_backend.py
+++ b/scripts/_divergence_backend.py
@@ -1,0 +1,89 @@
+"""Backend dispatch for divergence computations (CPU / CUDA).
+
+Private module — called by _divergence_semantic.py.
+
+Resolves the ``backend`` key from config/analysis.yaml:
+  auto  → CUDA if torch+CUDA available, else NumPy
+  cpu   → NumPy only
+  cuda  → torch CUDA (raises if unavailable)
+"""
+
+import numpy as np
+from utils import get_logger
+
+log = get_logger("_divergence_backend")
+
+_TORCH_AVAILABLE: bool | None = None
+_DEVICE: str | None = None
+
+
+def _probe_torch() -> bool:
+    """Lazy-check whether torch with CUDA is importable."""
+    global _TORCH_AVAILABLE
+    if _TORCH_AVAILABLE is None:
+        try:
+            import torch
+
+            _TORCH_AVAILABLE = torch.cuda.is_available()
+            if _TORCH_AVAILABLE:
+                log.info(
+                    "torch %s, CUDA device: %s",
+                    torch.__version__,
+                    torch.cuda.get_device_name(0),
+                )
+            else:
+                log.info("torch available but no CUDA — using NumPy")
+        except ImportError:
+            _TORCH_AVAILABLE = False
+            log.info("torch not installed — using NumPy")
+    return _TORCH_AVAILABLE
+
+
+def get_backend(cfg: dict) -> str:
+    """Return 'torch' or 'numpy' based on config and hardware.
+
+    Parameters
+    ----------
+    cfg : dict
+        Full analysis config (reads ``cfg["divergence"]["backend"]``).
+
+    Returns
+    -------
+    str
+        ``"torch"`` or ``"numpy"``.
+
+    Raises
+    ------
+    RuntimeError
+        If ``backend: cuda`` but no CUDA-capable torch is found.
+
+    """
+    setting = cfg["divergence"].get("backend", "auto")
+
+    if setting == "cpu":
+        log.info("Backend forced to NumPy (config: cpu)")
+        return "numpy"
+
+    has_cuda = _probe_torch()
+
+    if setting == "cuda":
+        if not has_cuda:
+            raise RuntimeError(
+                "Config requires backend: cuda but torch CUDA is not available"
+            )
+        return "torch"
+
+    # auto
+    return "torch" if has_cuda else "numpy"
+
+
+def to_tensor(arr: np.ndarray) -> "torch.Tensor":
+    """Convert NumPy array to float32 torch tensor on CUDA."""
+    import torch
+
+    return torch.as_tensor(arr, dtype=torch.float32, device="cuda")
+
+
+def to_numpy(t: "torch.Tensor") -> np.ndarray:
+    """Move torch tensor back to NumPy on CPU."""
+    return t.detach().cpu().numpy()
--- a/scripts/_divergence_backend.py
+++ b/scripts/_divergence_backend.py
@@ -14,7 +14,7 @@ from utils import get_logger
 log = get_logger("_divergence_backend")
 
 _TORCH_AVAILABLE: bool | None = None
-_DEVICE: str | None = None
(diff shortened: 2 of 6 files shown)


MinhHaDuong (commented):
## Multi-perspective review — PR #654

### Risk level: Standard + scripts → all five agents

---

### Correctness agent — **comment** (medium confidence)

1. **S2 energy distance: diagonal inclusion mismatch** (medium confidence).
   `_energy_distance_torch` computes `torch.cdist(Xt, Xt).mean()` which **includes the zero diagonal** (self-distances), while `dcor.energy_distance` **excludes self-pairs** in its U-statistic estimator. For large n (200+) the effect is small (200 zeros out of 40000 entries = 0.5%), hence the test likely passes within RTOL=1e-3. But this is a **systematic bias** that grows as n shrinks. Consider masking the diagonal or using `triu` indices, matching dcor's behavior exactly:
   ```python
   # Exclude diagonal for unbiased estimator
   n = Xt.shape[0]
   dXX = (torch.cdist(Xt, Xt).sum() - 0) / (n * (n - 1))  # already 0 on diag for cdist
   ```
   Actually, since `cdist(X, X)` diagonal is exactly 0 for L2, including it only deflates the mean slightly. The effect on the difference `2*dXY - dXX - dYY` is that dXX and dYY are both slightly underestimated, which slightly inflates the result. Verdict: **acceptable for current use** (test validates equivalence), but worth a comment in the code noting the approximation.

2. **`_mmd_rbf_torch` unsqueeze check**: The torch implementation correctly mirrors the numpy version. The `gamma` calculation, kernel computation, and unbiased estimator (excluding trace) all match. No issues found.

3. **`_frechet_torch` eigendecomposition**: Uses `torch.linalg.eigh` + clamping negative eigenvalues to 0 — this is a valid and numerically stable approach, potentially **better** than scipy's `sqrtm` which can produce complex results. Good choice.

4. **S3 Wasserstein GPU dispatch**: Passes torch tensors directly to POT's `sliced_wasserstein_distance`. POT's documentation confirms torch backend support. The test uses wider tolerance (5e-3) to account for different random projection implementations — reasonable.

5. **`get_backend` called per compute function**: Each S1-S4 function calls `get_backend(cfg)` independently. The `_probe_torch()` result is cached in module-level `_TORCH_AVAILABLE`, so the actual CUDA probe happens only once. Fine.

---

### Consistency agent — **approve** (high confidence)

1. **Naming**: `_mmd_rbf_torch`, `_energy_distance_torch`, `_frechet_torch` — consistent `_{method}_torch` pattern. Good.
2. **Import style**: Lazy imports inside functions (torch, dcor, ot) — matches the existing pattern in the file.
3. **Docstrings**: All new functions have docstrings with formulas or references. Matches codebase style.
4. **Backend module**: Clean separation of concerns. `_divergence_backend.py` handles only dispatch/conversion.
5. **Test file**: Follows existing test patterns with `SCRIPTS_DIR` path insertion, fixtures, and class-based grouping.

One minor nit: **`_DEVICE` variable** in `_divergence_backend.py` (line 17) is declared but never used. Dead code.

---

### Scope agent — **comment** (medium confidence)

1. **Ticket 0042 modification**: The PR removes `Blocked-by: 0037` and `Blocked-by: 0039` from ticket 0042. Removing the 0037 blocker makes sense since this PR implements 0037. But **removing the 0039 blocker** (robustness) is a judgment call — the log entry says "CPU run is fine, robustness is polish." This is reasonable but is a planning decision, not a code change. Consider whether this belongs in a separate commit or is fine bundled here.

2. **Ticket 0037 status**: Changed to `in-progress` but the PR is ready for review. Should this be `in-review` or left as-is until merge?

3. Otherwise, the changes are tightly scoped to exactly what the ticket specifies: backend dispatch, torch implementations for S1-S4, tests, and config.

---

### Red team agent — **comment** (medium confidence)

1. **CUDA OOM on large arrays**: `to_tensor` always sends to `"cuda"` device. With `max_subsample: 2000` and 1024-dim embeddings, `torch.cdist` creates a 2000x2000 matrix — only ~15MB, well within A4000's 16GB. However, if `max_subsample` is ever increased significantly, OOM could occur silently. The backend module has no memory guard. Low risk given current config, but worth noting.

2. **Global mutable state**: `_TORCH_AVAILABLE` is a module-level global mutated by `_probe_torch()`. The test correctly resets it via monkeypatch. However, if tests run in parallel (pytest-xdist), different workers could see inconsistent state. Since the test file is entirely skipped without CUDA, and dispatch tests use monkeypatch, this is **low risk**.

3. **No validation of `backend` config value**: If someone types `backend: gpu` instead of `backend: cuda`, `get_backend` falls through to the `auto` path silently. Consider validating the setting against `{"auto", "cpu", "cuda"}`.

4. **`to_tensor` hardcodes `device="cuda"`**: If multi-GPU setups are ever needed, this would need `device=f"cuda:{N}"`. Not a problem now but worth a comment.

---

### Doc propagation agent — **approve** (high confidence)

1. **Config comment**: `config/analysis.yaml` has inline comment `# auto (GPU if available), cpu, cuda` — matches the backend module's behavior exactly.
2. **Docstring updates**: `compute_s2_energy` docstring updated to mention torch. `compute_s3_wasserstein` docstring mentions POT torch support. Good.
3. **PR description**: Accurately describes all changes and test results.
4. **No Makefile changes needed**: The backend dispatch is transparent to the pipeline — same entry points, same outputs. No new make targets required.

---

### Synthesis

| Finding | Agent | Confidence | Severity |
|---------|-------|------------|----------|
| S2 energy diagonal inclusion differs from dcor | Correctness | Medium | Comment (systematic but small bias, validated by test) |
| `_DEVICE` unused variable | Consistency | High | Nit — remove it |
| `backend` config value not validated | Red team | Medium | Comment — silent fallthrough to auto |
| Ticket 0042 unblocking 0039 is a planning decision | Scope | Medium | Comment — author's call |
| Ticket 0037 status should reflect PR state | Scope | Low | Nit |

**Overall verdict: COMMENT.** No blocking issues. The implementation is clean, well-tested (within CUDA-available environments), and properly scoped. The energy distance diagonal note is the most substantive finding but is validated by the equivalence test.

PR #655, opened 2026-04-15
fix: simplify backend, fix divergence.mk and G3 dtype

MinhHaDuong:
## Summary
- Cache resolved backend in `_RESOLVED_BACKEND` — log once, not per method
- Remove dead `to_numpy()` and unused `numpy` import from backend module
- Fix `divergence.mk` to use `uv run python` (was bare `python3`, broke in worktrees)
- Fix G3 coupling_age: `pd.to_numeric()` on `ref_year` before `.median()` (real data has string dtype)

## Context
Follow-up to #654 (squash-merged). These are the simplify review fixes + bugs found during the first real data run.

## Test plan
- [x] 8/8 GPU tests pass
- [x] `make -j15 divergence-tables` completes all 15 methods on real corpus
- [x] G3 coupling_age produces valid output

### DIFF
--- a/config/analysis.yaml
+++ b/config/analysis.yaml
@@ -21,6 +21,7 @@ clustering:
 
 divergence:
   random_seed: 42
+  backend: auto        # auto (GPU if available), cpu, cuda
   windows: [2, 3, 4, 5]
   min_papers: 30          # per window minimum (override with 5 for smoke)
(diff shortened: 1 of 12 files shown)


MinhHaDuong:
Closing — will recreate from clean branch without data/catalogs symlink noise

PR #666, opened 2026-04-16
fix: repair 13 failing tests

MinhHaDuong:
## Summary

- **9 `test_alluvial_unit.py` failures**: `TestLoadAnalysisCorpus._call()` mocked `pd.read_csv` but `load_refined_works()` calls `os.path.exists()` before reaching `read_csv`, hitting `FileNotFoundError`. Fixed by also mocking `os.path.exists` to return `True` for all paths except the `/nonexistent` feather sentinel.
- **2 `test_doc_contract_consistency.py` failures**: `corpus-construction.md` was refactored into separate includes but lost its pipeline overview. Added a 4-stage overview (discover → enrich → extend → filter) with all four artifact names.
- **2 `test_doc_vars_completeness.py` failures**: After the technical-report/corpus-report split, `emb_dimensions` was used in prose but missing from `DOC_VARS`, and 9 citation/language variables moved to the corpus-report but stayed in `DOC_VARS["technical-report"]`. Added `emb_dimensions`, removed the 9 dead entries.

## Test plan

- [x] `pytest tests/test_alluvial_unit.py::TestLoadAnalysisCorpus` — 9/9 pass
- [x] `pytest tests/test_doc_contract_consistency.py::TestCorpusConstructionMd` — 3/3 pass
- [x] `pytest tests/test_doc_vars_completeness.py` — 6/6 pass
- [x] `make check-fast` — 764 passed, 0 failed

### DIFF
--- a/content/_includes/corpus-construction.md
+++ b/content/_includes/corpus-construction.md
@@ -1,3 +1,14 @@
+## Pipeline overview
+
+Corpus construction follows four stages, each producing an intermediate artifact:
+
+1. **Discover** — merge sources into a deduplicated pool → `unified_works.csv`
+2. **Enrich** — fetch metadata, abstracts, and citations → `enriched_works.csv`
+3. **Extend** — compute derived flags (language, relevance, duplicates) without removing rows → `extended_works.csv`
+4. **Filter** — apply inclusion policy, audit removals → `refined_works.csv`
+
+The discover stage is described below; enrichment and filtering are documented in their own sections.
+
 ## Discovery Sources
 
 ### Sources
--- a/scripts/compute_vars.py
+++ b/scripts/compute_vars.py
@@ -58,14 +58,6 @@ DOC_VARS = {
         "bim_n_post2015",
         "bim_n_pre2007",
         "bim_var_pct",
-        "cite_coverage_pct",
-        "cite_crossref_rows",
-        "cite_doi_ref_pct",
-        "cite_doi_ref_rows",
-        "cite_fetched_dois",
-        "cite_never_fetched",
-        "cite_total_dois",
-        "cite_total_rows",
         "cite_refined_rows",
         "cite_refined_sources",
         "corpus_core",
@@ -74,7 +66,7 @@ DOC_VARS = {
         "corpus_total",
         "corpus_with_embeddings",
         "analysis_corpus_n",
-        "lang_english_pct",
+        "emb_dimensions",
         "pca_emb_pc2_cosine",
         "pca_emb_pc2_dbic",
         "pca_emb_pc2_var_pct",
@@ -157,6 +149,7 @@ MISSING = "[MISSING]"
 
 # ── Helpers ──────────────────────────────────────────────────
 
+
 def _int(value):
     """Format integer with comma thousands separator."""
     return f"{int(round(value)):,}"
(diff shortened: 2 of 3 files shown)


MinhHaDuong (commented):
## PR Review: fix: repair 13 failing tests

**Risk level**: Standard (test fixes + doc content + vars sync)

### Correctness (confidence: high) — APPROVE

All three fixes are correct:

1. **`test_alluvial_unit.py` mock fix**: The `os.path.exists` mock (`side_effect=lambda p: p != "/nonexistent"`) correctly ensures `load_refined_works()` skips the Feather path (patched to `/nonexistent`) and reaches the mocked `pd.read_csv`. Without this mock, tests pass only when `REFINED_WORKS_PATH` happens to exist on disk (dev machines with data) but fail in CI or data-free worktrees. The fix makes tests environment-independent.

2. **`corpus-construction.md` pipeline overview**: The added 4-stage overview correctly names all four artifacts (`unified_works.csv` → `enriched_works.csv` → `extended_works.csv` → `refined_works.csv`) and satisfies the `test_doc_contract_consistency` checks.

3. **`compute_vars.py` DOC_VARS sync**: Correctly adds `emb_dimensions` (used in `technical-report.qmd` line 20) and removes 9 dead entries. Verified: the 8 `cite_*` variables moved to `citation-quality.md` which is included only from `corpus-report.qmd`, not `technical-report.qmd`. `lang_english_pct` is used only in `data-paper.qmd` and `companion-paper.qmd` (both already have it).

All 18 targeted tests pass. No new test failures introduced (the 27 failures in `make check` are all pre-existing: corpus acceptance, regression, golden values, venue table — all require data files).

### Consistency (confidence: high) — APPROVE

- Formatting changes are pure autoformatter (black-style): `\`-continuation to parenthesized `with` blocks, line wrapping, blank lines before class/function defs. No behavioral changes mixed with formatting.
- The `with (...)` context manager syntax (PEP 617) is appropriate for Python 3.10+ (project uses 3.12).

### Scope (confidence: medium) — COMMENT

The PR title says "13 failing tests" but on main I observe only **4 failures** in these test files (2 in `test_doc_contract_consistency`, 2 in `test_doc_vars_completeness`). The 9 `test_alluvial_unit` tests pass on main because the dev machine has `refined_works.csv` on disk. The alluvial mock fix is still valuable (makes tests environment-independent), but the PR description overstates the problem.

Additionally, the PR includes substantial **autoformatter whitespace** changes in both `compute_vars.py` and `test_alluvial_unit.py` that are not related to the bug fixes. While these are harmless and improve consistency, they inflate the diff (117 additions / 76 deletions, when the semantic changes are ~20 lines). Consider separating formatting-only commits from fix commits for cleaner git history.

### Doc propagation (confidence: high) — APPROVE

- The added pipeline overview in `corpus-construction.md` is consistent with `architecture.md` (same 4-stage names, same artifact names).
- No downstream documents reference the removed DOC_VARS entries from `technical-report`.
- `emb_dimensions` is already computed in `embedding_stats()` (line 281: `v["emb_dimensions"] = str(vectors.shape[1])`), so it will be populated at runtime.

### Verdict: **APPROVE**

One minor observation (not blocking):

- **Fragile mock breadth**: The `os.path.exists` mock returns `True` for *all* paths except `/nonexistent`. This means `load_analysis_config()` and `load_analysis_periods()` (also in `pipeline_loaders`) see `True` for their config paths and then open the real files. This works because `config/analysis.yaml` exists in the repo, but the mock is broader than needed. A more targeted approach would mock only the specific paths, but this is a nit — the current approach is correct and the test passes reliably.

---
*Review by Correctness, Consistency, Scope, and Doc propagation agents.*

PR #669, opened 2026-04-16
fix: G1 PageRank — replace zero-padded Kendall tau with JS divergence (t0059)

MinhHaDuong:
## Summary
- Replace union+zero-padding+Kendall-tau with JS divergence on PageRank distributions
- New `_compare_pagerank_distributions()` helper bins PageRank values and computes JSD²
- Works correctly on disjoint node sets from sliding windows

## Test plan
- [x] `test_g1_disjoint_windows_not_degenerate` — similar vs dissimilar graphs produce different values
- [x] `test_g1_output_matches_schema` — DivergenceSchema validation
- [x] `test_g1_value_bounded_0_1` — output range check
- [x] `make check-fast` passes (788 passed, 0 failed)

### DIFF
--- a/config/analysis.yaml
+++ b/config/analysis.yaml
@@ -62,6 +62,7 @@ divergence:
   citation:
     G1_pagerank:
       damping: 0.85
+      n_bins: 20                                   # histogram bins for JS divergence on PR distributions
     G2_spectral: {}
     G3_coupling_age: {}
     G4_cross_tradition:
--- a/scripts/_citation_methods.py
+++ b/scripts/_citation_methods.py
@@ -84,8 +84,8 @@ def _compare_pagerank_distributions(pr_before, pr_after, n_bins=20):
         return np.nan
     bin_max = max(pr_b.max(), pr_a.max())
     bins = np.linspace(0, bin_max, actual_bins + 1)
-    hist_b, _ = np.histogram(pr_b, bins=bins, density=True)
-    hist_a, _ = np.histogram(pr_a, bins=bins, density=True)
+    hist_b, _ = np.histogram(pr_b, bins=bins)
+    hist_a, _ = np.histogram(pr_a, bins=bins)
 
     # Add epsilon to avoid zero bins, then re-normalize
     eps = 1e-10
@@ -206,8 +206,9 @@ def compute_g1_pagerank(works, citations, internal_edges, cfg):
     Returns DataFrame with columns: year, window, hyperparams, value
     """
     cit_cfg = cfg["divergence"]["citation"]
-    damping = cit_cfg["G1_pagerank"]["damping"]
-    n_bins = 20
+    g1_cfg = cit_cfg["G1_pagerank"]
+    damping = g1_cfg["damping"]
+    n_bins = g1_cfg.get("n_bins", 20)
 
     log.info("G1: PageRank distribution divergence (sliding, JS)")
     results = []
--- a/scripts/_citation_methods.py
+++ b/scripts/_citation_methods.py
@@ -67,32 +67,20 @@ def _compare_pagerank_distributions(pr_before, pr_after, n_bins=20):
     _nodes_b, vals_b = pr_before
     _nodes_a, vals_a = pr_after
 
-    # Sort descending (distribution shape, not per-node identity)
-    pr_b = np.sort(vals_b)[::-1]
-    pr_a = np.sort(vals_a)[::-1]
-
-    # Normalize to probability distributions
-    sum_b, sum_a = pr_b.sum(), pr_a.sum()
-    if sum_b == 0 or sum_a == 0:
-        return np.nan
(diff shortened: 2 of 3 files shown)


MinhHaDuong (commented):
## Multi-perspective review — PR #669

**Risk level**: Standard + scripts (5-agent review)

---

### Correctness agent — **APPROVE** (high confidence)

The fix correctly addresses the core problem: on disjoint sliding windows (which share zero nodes by construction), comparing per-node PageRank via union + zero-padding made Kendall tau measure tie-breaking among zeros, not genuine structural shifts.

**What's right:**
- `_compare_pagerank_distributions()` bins PageRank *values* into histograms and computes JSD² — a well-defined distributional comparison that works on disjoint populations
- Epsilon smoothing (`1e-10`) prevents zero-bin log issues before re-normalization
- `actual_bins = min(n_bins, min(len(pr_b), len(pr_a)))` handles edge cases with very small graphs
- The NaN returns for degenerate cases (zero-sum, <2 bins) are correct guard clauses
- `jensenshannon()` returns the square root of JS divergence; squaring it gives JSD in [0,1] ✓
- All 3 new tests pass, and all 796 existing tests pass

**Test quality:**
- `test_g1_disjoint_windows_not_degenerate` — strong: constructs star vs ring topologies with known PageRank shapes, asserts distributional discrimination on completely disjoint node sets
- `test_g1_output_matches_schema` — validates DivergenceSchema on synthetic 11-year corpus
- `test_g1_value_bounded_0_1` — boundary check on star vs chain

---

### Consistency agent — **COMMENT** (medium confidence)

**1. `n_bins=20` is hardcoded, not config-driven.**
The architecture rule says: *"All research parameters in `config/analysis.yaml`, read via `load_analysis_config()`. No hardcoded constants for values that might change."* Other methods read their parameters from config (`damping`, `max_nodes`, `n_communities`, `top_n_terms`). The new `n_bins=20` should live in `config/analysis.yaml` under `G1_pagerank:`.

**2. `kendalltau` import is still present on main.**
The diff removes the import correctly, but note: the `from scipy.stats import entropy, kendalltau` becomes `from scipy.stats import entropy`. Clean. ✓

**3. Hyperparams string format is consistent** — other methods use empty string, this one uses `n_bins=20`. This is actually an improvement (more informative), but inconsistent with siblings. Low priority.

---

### Scope agent — **APPROVE** (high confidence)

The PR does exactly what ticket 0059 action #2 asks: replace G1's zero-padded Kendall tau with distributional comparison.

**In scope:** ✓ G1 fix, tests, import cleanup
**Not in scope (correctly deferred):** Actions 3-4 from ticket 0059 (audit G2/G5/G6/G8 for similar disjoint-window issues). The ticket remains open for these.

No unrelated changes. Clean scope.

---

### Red team agent — **COMMENT** (medium confidence)

**1. Histogram bin edge sensitivity.** With `n_bins=20` and `bin_max = max(pr_b.max(), pr_a.max())`, if one distribution has a single outlier node with very high PageRank (e.g., a star center), most values cluster in the first bin and the histogram becomes sparse. This could make JSD less discriminating for graphs where one hub dominates. Not a bug — the test explicitly validates star-vs-ring discrimination — but worth noting for very skewed distributions.

**2. `density=True` then re-normalize.** `np.histogram(..., density=True)` already normalizes by bin width, then the code adds epsilon and re-normalizes by sum. This double-normalization is safe (JSD only cares about the ratio), but the `density=True` is doing unnecessary work. Using `density=False` and normalizing once would be cleaner. Not a correctness issue.

**3. Identical distributions.** Two identical star graphs with the same size produce JSD very close to 0 but not exactly 0 (due to epsilon smoothing). The test asserts `val_different > val_similar` — this holds because epsilon is tiny. No issue.

---

### Doc propagation agent — **APPROVE** (high confidence)

- No references to G1, PageRank, or Kendall tau found in `content/`, `docs/`, `README.md`, or `STATE.md`
- `config/analysis.yaml` is not modified (though it arguably should be — see consistency finding)
- Architecture doc unaffected — the method still follows the dispatcher pattern
- The `kendalltau` import removal doesn't affect any other module

---

## Summary

| Agent | Verdict | Confidence |
|-------|---------|-----------|
| Correctness | approve | high |
| Consistency | comment | medium |
| Scope | approve | high |
| Red team | comment | medium |
| Doc propagation | approve | high |

**Overall: APPROVE with 2 comments**

### Actionable findings (ordered by priority):

1. **[Consistency, medium]** Move `n_bins=20` to `config/analysis.yaml` under `G1_pagerank:` to follow the config-driven parameters rule. Currently the only hardcoded research parameter in the PR.

2. **[Red team, low]** Consider `density=False` in `np.histogram()` since you re-normalize anyway. Cosmetic — no correctness impact.

Both are nits that can be addressed in a follow-up if desired. The core fix is correct and well-tested.

PR #670, opened 2026-04-16
fix: C2ST — use shuffled StratifiedKFold (t0060)

MinhHaDuong:
## Summary
- Replace default `cv=int` with explicit `StratifiedKFold(shuffle=True, random_state=seed)`
- Contiguous labels [0,0,...,1,1,...] with no shuffle caused biased fold boundaries

## Test plan
- [x] `test_shuffled_cv_reproducible` — same seed → same AUC
- [x] `test_contiguous_labels_not_degenerate` — AUC in reasonable range
- [x] `test_uses_shuffled_stratified_kfold` — implementation check
- [x] `make check-fast` passes (788 passed, 0 failed)

### DIFF
--- a/scripts/_divergence_c2st.py
+++ b/scripts/_divergence_c2st.py
@@ -16,7 +16,7 @@ import pandas as pd
 import scipy.sparse as sp
 from sklearn.decomposition import PCA
 from sklearn.linear_model import LogisticRegression
-from sklearn.model_selection import cross_val_score
+from sklearn.model_selection import StratifiedKFold, cross_val_score
 from utils import get_logger
 
 log = get_logger("_divergence_c2st")
@@ -65,7 +65,8 @@ def _c2st_auc(X, Y, *, cv_folds=5, class_weight="balanced", seed):
         max_iter=1000,
         random_state=seed,
     )
-    scores = cross_val_score(clf, features, labels, cv=cv_folds, scoring="roc_auc")
+    cv = StratifiedKFold(n_splits=cv_folds, shuffle=True, random_state=seed)
+    scores = cross_val_score(clf, features, labels, cv=cv, scoring="roc_auc")
     return float(np.mean(scores))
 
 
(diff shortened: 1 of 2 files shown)


MinhHaDuong (commented):
## Review: APPROVE

The fix is correct and well-scoped. All five checklist items pass.

### 1. Fix correctness

`_c2st_auc` now explicitly constructs `StratifiedKFold(n_splits=cv_folds, shuffle=True, random_state=seed)` and passes it as the `cv=` argument to `cross_val_score`. This directly addresses the root cause: labels are always contiguous `[0,0,...,1,1,...]` by construction (`np.vstack([X, Y])`), so without shuffle the fold boundaries fall at predictable positions relative to the class boundary. The fix is minimal and correct.

### 2. Import added

`StratifiedKFold` is added to the import line alongside `cross_val_score` — one line change, no drift.

### 3. Test coverage

Three new tests in `TestC2STCore`:

- `test_shuffled_cv_reproducible` — checks same seed → same AUC (determinism regression).
- `test_contiguous_labels_not_degenerate` — the bug scenario: mild shift produces AUC in `(0.5, 0.95)`. This is the most valuable test; it would have caught the original bug.
- `test_uses_shuffled_stratified_kfold` — patches `StratifiedKFold.__init__` to verify `shuffle=True` is actually passed. Solid regression guard.

Minor nit: `test_contiguous_labels_not_degenerate` upper bound of `0.95` is fairly loose. However this matches the existing convention in the suite (null-AUC test uses `0.35–0.65`), and ticket #0063 already tracks tightening these tolerances.

### 4. Seed from config, not hardcoded

`seed` flows from `div_cfg.get("random_seed", 42)` in both callers (`compute_c2st_embedding`, `compute_c2st_lexical`), which reads from `config/analysis.yaml`. The `_c2st_auc` function takes `seed` as a keyword-only argument — no hardcoded value inside the function. Architecture rule satisfied.

Note: the `.get("random_seed", 42)` fallback is dead code (key is always present in config). Ticket #0062 already tracks removing it. No action needed here.

### 5. Scope

The diff touches exactly two files: `scripts/_divergence_c2st.py` (2 lines) and `tests/test_c2st.py` (+61 lines). The ticket files are added as expected by the ticket workflow. No scope creep.

The fix is correct, minimal, tested, and properly seeded. Recommend merge.

PR #671, opened 2026-04-16
fix: per-window deterministic seeds in null model (t0061)

MinhHaDuong:
## Summary
- Replace shared rng with per-window deterministic seeds: `seed + year*100 + window`
- Separate rng instances for subsampling vs permutation
- Each (year, window) pair now produces identical results regardless of other pairs

## Test plan
- [x] `test_single_window_reproducible_regardless_of_prior_windows_semantic`
- [x] `test_single_window_reproducible_regardless_of_prior_windows_lexical`
- [x] `make check-fast` passes (787 passed, 0 failed)

### DIFF
--- a/scripts/compute_null_model.py
+++ b/scripts/compute_null_model.py
@@ -213,7 +213,6 @@ def _run_semantic_permutations(method_name, div_df, cfg):
     perm_cfg = div_cfg["permutation"]
     n_perm = perm_cfg["n_perm"]
     seed = div_cfg.get("random_seed", 42)
-    rng = np.random.RandomState(seed)
 
     _, min_papers, max_subsample, _, equal_n = _get_years_and_params(df, emb, cfg)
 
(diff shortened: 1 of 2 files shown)


MinhHaDuong (commented):
## Review: APPROVE (self-PR — posting as COMMENT)

This is a clean, targeted fix. All checklist items pass.

### 1. Seed derivation correctness

`window_seed = seed + year * 100 + window` is correct for the actual corpus range (years 1990–2025, windows [2,3,4,5]). Each (year, window) pair maps to a unique seed — verified numerically: 0 collisions across all 144 actual pairs. The theoretical collision case raised in the review prompt (year=20, w=300) requires window values that are not used in this project.

### 2. Separate rng instances for subsampling vs permutation

`subsample_rng = RandomState(window_seed)` and `perm_rng = RandomState(window_seed + 50000)`. The +50000 gap is safe: subsample seeds span [199044, 202547] and perm seeds span [249044, 252547] — a gap of ~46,500 with no overlap. Both rng instances are passed to the correct call sites in both paths.

### 3. Fix applied to both semantic and lexical paths

Yes. Lines 229–234 (semantic) and lines 299–302 (lexical) receive identical treatment. Every rng usage — `_get_window_embeddings` (×2), `subsample_equal_n`, and `permutation_test` — is updated in both paths.

### 4. Tests verify window-order independence

`test_single_window_reproducible_regardless_of_prior_windows_semantic` and `..._lexical` both assert `z_single == z_double` (exact float equality) for year=2005 processed alone vs after year=2003. Correct regression test. Uses `mock.patch` with controlled synthetic data — fast and deterministic.

### 5. Seed collision risk

No collision possible with actual config values (windows [2,3,4,5], years 1990–2025). If future work expands windows to ≥100, the derivation should be revisited — not a blocker today.

### 6. Scope creep

None. Two files changed: `compute_null_model.py` (targeted fix) and `test_null_model.py` (regression tests). No unrelated changes.

### Minor nit (non-blocking)

The +50000 magic number is undocumented. A comment explaining why 50000 ensures no overlap would help future readers.

TDD cycle is clean: red commit (tests) before green commit (fix). 787 passed.

**Recommendation: MERGE**

PR #672, opened 2026-04-16
refactor: divergence wave 1 design cleanup (t0062)

MinhHaDuong:
## Summary
- Hoist loop imports to module level
- Replace dead fallback seeds with explicit key access
- Add max_subsample to lexical path
- Add else-raise for unhandled channel
- Document window column type in schemas.py
- Update divergence.mk section headers

## Test plan
- [x] `make check-fast` passes (785 passed, 0 failed)
- [x] No behavioral changes — cleanup only

### DIFF
--- a/divergence.mk
+++ b/divergence.mk
@@ -5,10 +5,10 @@
 # Architecture: one CSV per method via scripts/compute_divergence.py --method X
 #
 # Targets:
-#   divergence-semantic  Compute semantic methods (S1-S4)
-#   divergence-lexical   Compute lexical methods (L1-L3)
+#   divergence-semantic  Compute semantic methods (S1-S4, C2ST_embedding)
+#   divergence-lexical   Compute lexical methods (L1-L3, C2ST_lexical)
 #   divergence-citation  Compute citation methods (G1-G8)
-#   divergence-tables    All 17 divergence CSVs
+#   divergence-tables    All divergence CSVs
 #   divergence-figures   Plot one figure per method
 #   divergence           Both tables and figures
 #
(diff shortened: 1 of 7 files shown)


MinhHaDuong (commented):
## Review: t0062 design cleanup

**Verdict: APPROVE**

### Behavioral vs cosmetic

Four changes are purely cosmetic (hoisted imports, Makefile comment fixes, window column doc).
Two changes are **behaviorally significant** and intentional:

1. **Dead seed removal (`div_cfg["random_seed"]` instead of `.get(..., 42)`)** — confirmed to use dict key access (fail-fast), not `.get()`. This is correct: if `random_seed` is absent from config, we want an immediate `KeyError`, not a silent fallback that could mask reproducibility bugs. Applies consistently across S1–S4, L1, C2ST embedding, C2ST lexical, null model (8 sites).

2. **`max_subsample` added to lexical iterator** — fills a genuine parity gap with the semantic path. This is a small behavioral change (caps window size before equal-n subsampling) that was clearly missing. The comment "Cap window size before equal-n (matches semantic path)" is correct.

3. **Null model S3 hardcoded `n_projections=500` replaced with config lookup** — previously ignored the config; now reads `n_projections[1]` (middle value). Behavioral, but obviously correct: the null model should use the same hyperparams as the main pipeline.

4. **`else: raise ValueError` for unsupported channel in `compute_null_model.main()`** — makes the fail-fast pattern complete. No behavior change for valid inputs; silent failure for invalid ones is now an error.

### Scope

No scope creep. The TODO comment added to `_divergence_citation.py` is a note, not code.

### One nit

`rng = np.random.RandomState(seed)` is now unconditional in `_iter_lexical_window_pairs` (removed the `if equal_n` guard). The RNG is used both for the new `max_subsample` cap and for `equal_n` subsampling, so this is correct — but a one-line comment would clarify why `rng` is always needed now. Not a blocker.

PR #675, opened 2026-04-16
feat: bootstrap CIs on divergence (ticket 0047)

MinhHaDuong:
## Summary

- Add `compute_divergence_bootstrap.py`: resample K times per (year, window) with deterministic per-replicate seeds, reusing statistic functions from `compute_null_model.py`
- Add `export_divergence_summary.py`: join point estimates + bootstrap quantiles + null model p-values into one summary table per method
- Add `BootstrapSchema` and `DivergenceSummarySchema` to `schemas.py`
- Add opt-in Makefile targets `bootstrap-tables` and `divergence-summary` in `divergence.mk` (NOT part of default `make divergence`)

## Test plan

- [x] 13 unit tests in `test_bootstrap.py` covering bootstrap computation, schema validation, summary join logic, significant flag, and quantile ordering
- [x] `make check-fast` passes (1 pre-existing failure in `test_robustness_observability.py`, unrelated)
- [ ] Integration: `make bootstrap-tables divergence-summary` on full corpus (overnight GPU run)

### DIFF
--- a/config/analysis.yaml
+++ b/config/analysis.yaml
@@ -35,6 +35,9 @@ divergence:
     n_perm: 500
     z_threshold: 2.0    # consumed by compute_transition_zones.py (ticket 0055b)
 
+  bootstrap:
+    k: 200               # replicates for estimation CIs (ticket 0047, opt-in)
+
   c2st:
     pca_dim: 32
     cv_folds: 5
--- a/divergence.mk
+++ b/divergence.mk
@@ -199,6 +199,48 @@ $(DIV_TABLES)/tab_null_$(m).csv: $(NULL_DISPATCH) $(DIV_TABLES)/tab_div_$(m).csv
 .PHONY: null-model
 null-model: $(NULL_CSV)
 
+# ── Bootstrap CIs (ticket 0047) ──────────────────────────────────────────
+#
+# NOT part of default divergence — run explicitly:
+#   make bootstrap-tables
+#   make divergence-summary
+#
+# For each method, resample with replacement K times and recompute
+# the statistic.  Output: tab_boot_{method}.csv
+
+BOOT_DISPATCH := scripts/compute_divergence_bootstrap.py
+BOOT_METHODS_SEM := S2_energy
+BOOT_METHODS_LEX := L1
+BOOT_METHODS := $(BOOT_METHODS_SEM) $(BOOT_METHODS_LEX)
+BOOT_CSV := $(foreach m,$(BOOT_METHODS),$(DIV_TABLES)/tab_boot_$(m).csv)
+
+# Semantic bootstrap (depends on embeddings + divergence CSV)
+$(foreach m,$(BOOT_METHODS_SEM),$(eval \
+$(DIV_TABLES)/tab_boot_$(m).csv: $(BOOT_DISPATCH) $(DIV_TABLES)/tab_div_$(m).csv scripts/_divergence_semantic.py $(REFINED) $(REFINED_EMB) $(DIV_CFG) ; \
+	uv run python $(BOOT_DISPATCH) --method $(m) --div-csv $(DIV_TABLES)/tab_div_$(m).csv --output $$@))
+
+# Lexical bootstrap (depends on REFINED + divergence CSV)
+$(foreach m,$(BOOT_METHODS_LEX),$(eval \
+$(DIV_TABLES)/tab_boot_$(m).csv: $(BOOT_DISPATCH) $(DIV_TABLES)/tab_div_$(m).csv scripts/_divergence_lexical.py $(REFINED) $(DIV_CFG) ; \
+	uv run python $(BOOT_DISPATCH) --method $(m) --div-csv $(DIV_TABLES)/tab_div_$(m).csv --output $$@))
+
+.PHONY: bootstrap-tables
+bootstrap-tables: $(BOOT_CSV)
+
+# ── Divergence summary (ticket 0047) ────────────────────────────────────
+#
+# Joins point estimates + bootstrap CIs + null model into one table per method.
+
+SUMM_DISPATCH := scripts/export_divergence_summary.py
+SUMM_CSV := $(foreach m,$(BOOT_METHODS),$(DIV_TABLES)/tab_summary_$(m).csv)
+
+$(foreach m,$(BOOT_METHODS),$(eval \
+$(DIV_TABLES)/tab_summary_$(m).csv: $(SUMM_DISPATCH) $(DIV_TABLES)/tab_div_$(m).csv $(DIV_TABLES)/tab_boot_$(m).csv $(DIV_TABLES)/tab_null_$(m).csv ; \
+	uv run python $(SUMM_DISPATCH) --method $(m) --div-csv $(DIV_TABLES)/tab_div_$(m).csv --boot-csv $(DIV_TABLES)/tab_boot_$(m).csv --null-csv $(DIV_TABLES)/tab_null_$(m).csv --output $$@))
+
+.PHONY: divergence-summary
+divergence-summary: $(SUMM_CSV)
+
 # ── Top-level ────────────────────────────────────────────────────────────
 
 .PHONY: divergence
--- a/scripts/compute_divergence_bootstrap.py
+++ b/scripts/compute_divergence_bootstrap.py
@@ -67,15 +67,13 @@ def bootstrap_one_window(X_before, Y_after, statistic_fn, k, seed):
     replicates = []
     for i in range(k):
         rng = np.random.RandomState(seed + i * 1000)
+        idx_b = rng.choice(n_before, n_before, replace=True)
+        idx_a = rng.choice(n_after, n_after, replace=True)
 
         if is_array:
-            idx_b = rng.choice(n_before, n_before, replace=True)
-            idx_a = rng.choice(n_after, n_after, replace=True)
             boot_before = X_before[idx_b]
             boot_after = Y_after[idx_a]
         else:
-            idx_b = rng.choice(n_before, n_before, replace=True)
-            idx_a = rng.choice(n_after, n_after, replace=True)
             boot_before = [X_before[j] for j in idx_b]
             boot_after = [Y_after[j] for j in idx_a]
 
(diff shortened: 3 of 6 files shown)


MinhHaDuong (commented):
## PR Review: Bootstrap CIs on divergence (ticket 0047)

**Risk level**: Standard + scripts → `review:standard`

---

### Correctness [verdict: **approve**] (confidence: high)

The core `bootstrap_one_window()` logic is correct: it resamples both sides independently with replacement, applies the statistic function, and returns K float values. The per-replicate deterministic seed (`seed + i * 1000`) ensures reproducibility while varying across replicates.

The `build_summary()` join logic is clean: groups div by (year, window), aggregates bootstrap quantiles, merges null model p-values, and adds a `significant` boolean. The left-join strategy is appropriate — missing bootstrap or null rows become NaN rather than silently dropping.

**13 tests** cover: schema validation (valid/extra-column/coercion), K replicate count, variation, reproducibility, list inputs, summary join, significant flag, and quantile ordering. All pass.

No findings.

---

### Consistency [verdict: **approve**] (confidence: high)

- Follows the established dispatcher pattern: `--method X --div-csv ...` mirrors `compute_null_model.py` exactly.
- Uses `parse_io_args()` / `validate_io()` per script I/O discipline.
- Schema in `schemas.py` uses `strict=True, coerce=True` per architecture rules.
- Logging via `get_logger()` matches project convention.
- Reuses `_make_semantic_statistic`, `_make_lexical_statistic`, `_make_window_rngs` from `compute_null_model.py` — good code reuse rather than duplication.
- Makefile structure mirrors the null-model section exactly (foreach eval pattern, separate SEM/LEX lists).
- Test file follows the same synthetic-data-helper pattern as `test_null_model.py`.

No findings.

---

### Scope [verdict: **approve**] (confidence: high)

The PR adds exactly what the ticket asks:
1. Bootstrap computation script ✓
2. Summary table script ✓
3. Two Pandera schemas ✓
4. Opt-in Makefile targets (not in default `divergence`) ✓
5. 13 unit tests ✓

No unrelated changes. The `compute_null_model.py` changes (adding `_make_window_rngs`, per-window deterministic seeds) are from already-merged PRs and not part of this PR's diff.

No findings.

---

### Red team [verdict: **comment**] (confidence: medium)

1. **K=200 hardcoded as default, not in config** — The `--k 200` default is in the argparse, not in `config/analysis.yaml`. The architecture rule says "all research parameters in config/analysis.yaml." This isn't blocking because K is a CLI argument the Makefile could override, but it's inconsistent with how `n_perm: 500` is configured for the null model. Consider adding `bootstrap: {k: 200}` to `analysis.yaml` and reading it as the default.

2. **Seed overlap between bootstrap replicates and subsample/perm seeds** — The bootstrap replicate seeds use `boot_seed + i * 1000` where `boot_seed = seed + y*100 + w`. With K=200 replicates, this spans `[boot_seed, boot_seed + 199000]`. The permutation RNG from `_make_window_rngs` uses `window_seed + 50000` which falls at replicate i=50. Since bootstrap and null model run in separate processes and use separate RandomState instances, this causes no actual collision. But if they ever share a process, the seed spaces overlap. A comment documenting the seed allocation would help future maintainers.

3. **`np.random.RandomState` (legacy API)** — The code uses the legacy `np.random.RandomState` rather than `np.random.default_rng()`. This is consistent with the existing codebase (null model uses the same API), so not a regression — just noting for future modernization.

---

### Doc propagation [verdict: **approve**] (confidence: high)

- No content files reference bootstrap or summary tables yet (expected — the ticket says "referenced from technical report as supplementary material," which is a future step).
- `schemas.py` docstring updated to list new schemas and their consumers.
- No stale references found in `content/`, `docs/`, `README.md`, `STATE.md`, `ROADMAP.md`, or config files.
- `analysis.yaml` does not yet have a `bootstrap` section (see red-team finding #1).

No blocking findings.

---

### Test results

```
808 passed, 9 skipped, 9 warnings in 11.45s
```

All 13 bootstrap-specific tests pass. Pre-existing warnings (complexity smells in `plot_cocitation.py` and `plot_convergence.py`) are unrelated.

---

### Summary

| Agent | Verdict |
|---|---|
| Correctness | approve |
| Consistency | approve |
| Scope | approve |
| Red team | comment (2 nits, 1 note) |
| Doc propagation | approve |

**Overall: approve** — clean implementation that follows established patterns. The two red-team findings are non-blocking nits:
1. Consider moving K=200 default to `analysis.yaml` for consistency with how `n_perm` is configured.
2. Consider adding a comment documenting the bootstrap seed space to prevent future collisions.

MinhHaDuong:
## Post-merge code review — PR #675 (ticket 0047, bootstrap CIs)

**`make check-fast` status on current main**: 1 pre-existing failure (`test_robustness_observability.py::TestStepCounters::test_step1_counter_attempted`, unrelated, flaky under parallel execution). All 13 new bootstrap tests pass.

---

### 1. Correctness — Seed collision bug (medium severity)

**Finding**: The `boot_seed` formula in `compute_divergence_bootstrap.py` is:
```python
boot_seed = seed + y * 100 + w
```
This is identical to `_make_window_rngs`'s `window_seed` formula. Inside `bootstrap_one_window`, replicate `i` uses `RandomState(boot_seed + i * 1000)`.

**Consequence 1 — Within-run collisions**: For `k=200` and the realistic year range 1990–2025, windows [2,3,4,5]: there are **7,984 seed collisions** across different `(year, window, replicate)` triples. Specifically, `(y=X+10, w=W, rep=i)` and `(y=X, w=W, rep=i+1)` always share the same RNG seed. This means bootstrap replicates are *structurally correlated* across years — not truly independent draws.

**Consequence 2 — Perm-RNG collision**: `_make_window_rngs` sets `perm_rng = RandomState(window_seed + 50000)`. Bootstrap replicate `i=50` uses `RandomState(boot_seed + 50*1000) = RandomState(window_seed + 50000)` — identical to `perm_rng`. All 144 windows in the dataset are affected (every `(year, window)` pair).

The actual CI values won't be wildly wrong, but the documented "deterministic, reproducible" property is violated, and the bootstrap samples have structured correlation that should not exist.

**Suggested fix** — add a distinct offset in `bootstrap_one_window` caller:
```python
# Instead of:
boot_seed = seed + y * 100 + w
# Use (ensures uniqueness for y in [1990,2025], w in [2..5], i in [0..199]):
rng = np.random.RandomState(seed + y * 100000 + w * 1000 + i)
```
Or, expose the per-replicate seed formula in `_make_window_rngs` by adding a `bootstrap_rng` return value with an offset well beyond `+50000+200*1000 = +250000`, e.g. `window_seed + 1_000_000`.

---

### 2. Architecture — Importing private functions from a peer script (medium)

`compute_divergence_bootstrap.py` imports:
```python
from compute_null_model import (
    SUPPORTED_CHANNELS,
    _make_lexical_statistic,
    _make_semantic_statistic,
    _make_window_rngs,
)
```
Architecture rule 8 states: *"Method implementations live in private modules (`_divergence_semantic.py`, etc.)."* These factory functions belong in `_divergence_io.py` or a new `_divergence_helpers.py`, not in `compute_null_model.py` (a peer dispatcher script). The current coupling means changes to `compute_null_model.py` could silently break bootstrap without any test catching it at import time.

---

### 3. Code quality — Dead test helper

`_make_synthetic_div_df()` in `tests/test_bootstrap.py` is defined but never called. The summary tests define their own inline `div_df` instead of reusing this helper. Should either be used or removed.

---

### 4. Test quality — Missing coverage

- **NaN p_value → `significant=False`**: When `null_df` has no row for a given `(year, window)`, the left-join produces `p_value=NaN`, and `NaN < 0.05` evaluates to `False` in pandas. This is silent and potentially misleading — a window with *no null model result* looks the same as a non-significant result. No test covers this case, and there is no logged warning.

- **All windows skipped**: If every `(year, window)` in `div_df` is skipped (e.g., all below `min_papers`), the bootstrap returns an empty DataFrame and `BootstrapSchema.validate()` is called on an empty frame. No test verifies this path.

- **Lexical stat in `test_bootstrap_list_inputs`**: The test uses `len(set(a)) + len(set(b))` as the statistic. This tests the list-indexing mechanics of `bootstrap_one_window` but not the actual JS-divergence behavior. Not a blocker, but the semantic test suite has no coverage for the real `_make_lexical_statistic` code path in bootstrap.

- **Assertion style nit**: `test_summary_significant_flag` uses `row['significant'] is True or row['significant'] == True`. The `is True` check fails silently for `numpy.bool_(True)` (confirmed: `np.bool_(True) is True` → `False`). The `== True` fallback saves it, but the pattern is fragile. Prefer `assert bool(row['significant'])`.

---

### 5. Integration risk — Missing data graceful handling

Both join cases work gracefully via left join:
- **Missing bootstrap data** (empty `boot_df`): `boot_median/q025/q975` become `NaN`. Schema accepts nullable floats. No crash.
- **Missing null model data** (empty `null_df`): `p_value=NaN`, `significant=False`. No crash.

However, there is **no log warning** when either join produces NaN (i.e., when a partial summary is emitted). A downstream user reading the CSV has no indication that the CI or significance columns are absent-due-to-missing-input vs. absent-due-to-skipped-windows.

---

### 6. Makefile — Duplicated method list

`NULL_METHODS_SEM` and `BOOT_METHODS_SEM` are separate variables with identical values (`S2_energy`), as are `NULL_METHODS_LEX`/`BOOT_METHODS_LEX`. A future maintainer adding a method to one list could forget the other. Consider a comment or a shared `SUPPORTED_METHODS_SEM := S2_energy` variable used by both.

---

### Summary

| # | Area | Severity | Status |
|---|------|----------|--------|
| 1 | Seed collision — correlated bootstrap reps | Medium | Open |
| 2 | Private function import from peer script | Medium | Open |
| 3 | Dead test helper `_make_synthetic_div_df` | Nit | Open |
| 4 | Missing tests: NaN p_value, empty df, real lexical stat | Low | Open |
| 5 | No log warning on partial summary data | Low | Open |
| 6 | Duplicated method lists in divergence.mk | Nit | Open |

The seed collision (#1) is the only issue that could affect scientific conclusions. It should be fixed before integration results are interpreted. All 13 tests pass and `make check-fast` is green (1 pre-existing unrelated failure).

PR #684, opened 2026-04-16
test: add negative-Z regression for build_summary

MinhHaDuong:
## Summary
- Follow-up nit from PR #683 review. The z_score forwarding fix only exercised positive z values.
- Sign carries directional information (distributional concentration increase vs decrease) that p-value alone discards — a `clip`/`abs` regression would be invisible to the existing fixture.
- New test uses z = −1.5 (observed below null mean) and asserts it survives the three-way join verbatim.

## Test plan
- [x] `uv run python -m pytest tests/test_bootstrap.py::TestSummaryTable -q` → 4 passed.

### DIFF
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -294,6 +294,43 @@ class TestSummaryTable:
         # z_score is forwarded verbatim from the null model
         assert list(result.sort_values("year")["z_score"]) == [2.5, 3.33]
 
+    def test_summary_preserves_negative_z_score(self):
+        """Negative z (divergence below null mean) must survive the join.
+
+        Sign carries directional information (concentration increase vs
+        decrease) that p-value alone discards. The build_summary pipeline
+        must not clip, abs, or otherwise transform z_score.
+        """
+        from export_divergence_summary import build_summary
+
+        div_df = pd.DataFrame(
+            {
+                "year": [2005],
+                "channel": ["semantic"],
+                "window": ["3"],
+                "hyperparams": [""],
+                "value": [0.20],
+            }
+        )
+        null_df = pd.DataFrame(
+            {
+                "year": [2005],
+                "window": ["3"],
+                "observed": [0.20],
+                "null_mean": [0.35],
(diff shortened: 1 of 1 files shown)


MinhHaDuong:
## Multi-perspective review (trivial risk — Correctness + Consistency focus)

**Verdict: approve (posted as comment — author cannot self-approve).** Test-only follow-up, +37/-0, single file, narrowly scoped to plug the sign-regression hole left by PR #683's fixture. Tests green locally (`uv run python -m pytest tests/test_bootstrap.py::TestSummaryTable -q` → 4 passed).

### Correctness (confidence: high)
- The `z = -1.5` assertion catches a `clip(lower=0)` or `abs()` regression on the forwarding path, which was the exact blind spot called out in the PR summary.
- Inline `null_df` (rather than the shared `_make_synthetic_null_df()` helper, which only carries positive Zs) is the right call — matches the hazard being guarded.
- Filtering `boot_df` to year 2005 after the helper produces {2005, 2006} works but is slightly roundabout; a hand-built single-row boot frame would read more directly. Not worth churn.

### Consistency (confidence: high)
- Structure, docstring style, and `from export_divergence_summary import build_summary` import pattern all match the neighbouring tests in `TestSummaryTable`. Fits in cleanly.

### Scope (confidence: high)
- Touches only `tests/test_bootstrap.py`. No production code, no docs, no config. Exactly the nit the PR body advertises.

### Red team (confidence: medium)
- One minor fixture quirk worth flagging (non-blocking): the `null_df` row sets `z_score = -1.5` alongside `p_value = 0.85`. Under a two-sided standard normal, |z|=1.5 corresponds to p≈0.134, not 0.85, so the fixture is internally inconsistent. The test only asserts on `z_score`, so correctness is unaffected — but a future reader reusing this fixture for a p-value assertion could be misled. Consider either (a) a comment noting "p_value arbitrary — this test asserts z forwarding only" or (b) using a consistent p≈0.134. Author's discretion.

### Doc propagation
- N/A (tests only).

PR #686, opened 2026-04-16
fix(t0065): G4 community detection on full cumulative graph

MinhHaDuong:
## Summary
- Changes `ref_year` in `compute_g4_cross_trad` from the median of the year range to `max(years)`, so the 2-community bisection is computed on the full cumulative graph. Every paper gets a community assignment, so per-year cross-tradition rates are defined across the whole year range instead of collapsing to NaN after the median year.
- Adds two regression tests: year_max must carry a value, and any remaining NaNs must be contiguous (cold-start only).
- Closes the tail-NaN half of ticket 0065. Audit of the other zoo tables confirmed G4 was the only serious offender (G3/G5/G7 show at most cold-start NaN; all others are complete).

## Root cause recap
The old code snapshotted community detection at `int(np.median(years))`. Any source paper published after that year was absent from the community dict. The `dropna(subset=["s_comm", "r_comm"])` step then dropped every one of its outgoing edges, collapsing `total` to zero and yielding NaN. Because citations point backward in time, the tail never recovered — in the t0042 full-corpus run, years 2010–2024 (23 of 35) were all NaN.

## Behavior change
The community split is now a corpus-level identity (driven by the final cumulative graph) rather than a mid-corpus snapshot. Existing valid years (1998–2009 in the t0042 run) may shift slightly because the bisection now sees late-era papers; the semantic interpretation moves from "how did citations relate to the mid-corpus tradition split" to "how did cross-tradition mixing evolve over time given today's two-community structure." This was explicitly the choice made in ticket 0065 (Option 2: fix).

## Test plan
- [x] New `test_g4_cross_trad_covers_year_max` fails on HEAD~2 (valid only through 2007) and passes with the fix.
- [x] New `test_g4_cross_trad_nans_contiguous` keeps NaNs to a cold-start contiguous block.
- [x] `tests/test_citation_sliding.py` full suite: 16 passed.
- [x] `tests/test_divergence.py` + `tests/test_citation_sliding.py`: 65 passed.
- [x] `make check-fast`: 836 passed, 8 skipped (all warnings are pre-existing complexity/length smells).
- [x] `make check`: same 37 failures/errors as baseline (pre-existing — worktree has no refined corpus assets).
- [ ] Re-run `make content/tables/tab_div_G4_cross_tradition.csv` on padme to confirm 0 tail NaN on the real corpus.

### DIFF
--- a/scripts/_citation_methods.py
+++ b/scripts/_citation_methods.py
@@ -349,6 +349,12 @@ def compute_g4_cross_trad(works, citations, internal_edges, cfg):
     the tradition split as a corpus-level identity rather than a mid-corpus
     snapshot is what makes the tail (ticket 0065) defined at all.
 
+    Caveat: this projects the late-era community structure backward onto
+    early years (e.g., 1998–2009). "Cross-tradition" in 1999 therefore
+    means "crossing a 2024-visible tradition split," not "crossing a
+    tradition that existed in 1999." The technical-report G4 include
+    (ticket 0028) should state this explicitly.
+
     Returns DataFrame with columns: year, window, hyperparams, value
     """
     years = _get_years(works)
--- a/scripts/_citation_methods.py
+++ b/scripts/_citation_methods.py
@@ -343,6 +343,12 @@ def _bisect_communities(G_und):
 def compute_g4_cross_trad(works, citations, internal_edges, cfg):
     """Fraction of new internal citations crossing a 2-community boundary.
 
+    The 2-community bisection is computed once on the full cumulative graph
+    (all years), so every paper has a community assignment and per-year
+    cross-tradition rates are defined across the whole year range. Treating
+    the tradition split as a corpus-level identity rather than a mid-corpus
+    snapshot is what makes the tail (ticket 0065) defined at all.
+
     Returns DataFrame with columns: year, window, hyperparams, value
     """
     years = _get_years(works)
@@ -353,7 +359,7 @@ def compute_g4_cross_trad(works, citations, internal_edges, cfg):
         log.info("G4: Too few internal edges (%d), skipping", len(internal_edges))
         return _dict_to_df({y: np.nan for y in years})
 
-    ref_year = int(np.median(years))
+    ref_year = max(years)
     G_full = _cumulative_graph(works, internal_edges, ref_year)
     G_und = G_full.to_undirected()
 
--- a/tests/test_citation_sliding.py
+++ b/tests/test_citation_sliding.py
@@ -314,8 +314,13 @@ class TestUnchangedMethods:
             f"G3 windows: {df['window'].unique()}"
         )
 
-    def test_g4_still_uses_cumulative(self, data):
-        """G4 should still output window='cumulative'."""
+    def test_g4_output_window_label_is_cumulative(self, data):
+        """G4 output rows must carry window='cumulative' (schema stability).
+
+        This only checks the output label in the CSV. It does NOT assert
+        the computation is cumulative — the bisection runs once on the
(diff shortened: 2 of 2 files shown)


MinhHaDuong (commented):
## Multi-perspective review (standard risk)

Agents run: Correctness, Consistency, Scope, Red team. Doc propagation skipped — G4 is not referenced in any `content/**/*.qmd`, `_includes/`, or `*-vars.yml`.

### Verdict: approve with one observation

Clean minimal fix, proper red/green TDD (verified: reverting `_citation_methods.py` to HEAD~2 while keeping the new tests reproduces the failure — `Non-contiguous NaN years: [2000, 2008..2014]`). `make check-fast`: 836 passed, 8 skipped (matches PR body). New tests: 16 passed in `test_citation_sliding.py`.

### Correctness (high confidence) — approve

- Ticket 0065 exit criteria met: root cause documented (PR body + new docstring), NaN tail resolved (fix), test covers the exit-criteria regression (`test_g4_cross_trad_nans_contiguous` mirrors the ticket's suggested assertion verbatim).
- Fix is causally aligned with the diagnosis: `ref_year = max(years)` makes `_cumulative_graph` include every paper, so the community dict covers all source DOIs and `dropna(subset=["s_comm","r_comm"])` no longer empties the tail.
- No change to output schema: `_dict_to_df` still labels the window as `"cumulative"`; existing `test_g4_still_uses_cumulative` passes.

### Consistency (high confidence) — approve

- Module-level comment `G3, G4, G7 are per-year methods and use cumulative windows` (line 8) still holds.
- Style/naming matches surrounding code; new docstring paragraph fits the file's prose tone.
- No stale docs: `G4_cross_tradition` appears only in `divergence.mk`, `compute_divergence.py`, `plot_divergence.py`, `plot_convergence.py`, and tests. None reference the median-year semantics that changed.

### Scope (high confidence) — approve

- Diff is 42/-1 across exactly two files: the fix and its tests. No unrelated changes (no tickets, no arch rules, no sweeps).
- One ticket, one PR. Matches the worktree-isolation rule.

### Red team (medium confidence) — comment

- **Semantic projection risk (worth surfacing, not blocking).** Bisecting on the full cumulative graph projects the late-era two-community structure back onto early years. For a field your own analysis periodizes as crystallizing around 2007–2014, the 1998–2009 values are now "how did citations relate to today's tradition split" — not "how did the split evolve." PR body acknowledges this explicitly as the intended Option 2 choice. Flagging in case the tech-report G4 include (ticket 0028) needs a one-line caveat.
- **Edge cases unchanged by this PR.** All-NaN failure modes (too few edges, bisection fails, too few nodes) still return full-NaN series, which vacuously satisfies the contiguous-NaN test. Acceptable.
- **Audit claim not verified in-PR.** PR body says "Audit of the other zoo tables confirmed G4 was the only serious offender (G3/G5/G7 show at most cold-start NaN; all others are complete)." No test or artifact in this PR attests to this. Low priority — the ticket's exit criterion 3 ("no other zoo entries have surprise NaN tails") is the author's call to verify against the real corpus.
- **Cosmetic nit (low confidence).** `test_g4_still_uses_cumulative` is now slightly misleadingly named — the computation changed (ref_year semantics), only the output label is still `"cumulative"`. Not worth churning the test; mentioning in case a future reader gets confused.

### Remaining task

PR body lists one unchecked item: "Re-run `make content/tables/tab_div_G4_cross_tradition.csv` on padme to confirm 0 tail NaN on the real corpus." This is the real-corpus verification — should be done before closing ticket 0065, but does not block merge (the unit-test regression is sufficient for CI).

---
Label: `review:standard` (2 cycles required by merge gate).

PR #691, opened 2026-04-17
ticket 0069: G2_spectral null model + graph-channel bootstrap

MinhHaDuong:
## Summary
- Wire G2_spectral into the citation null-model dispatcher: split `_run_citation_permutations` into per-method drivers (`_run_g9_community_permutations`, new `_run_g2_spectral_permutations`) selected via a registry. G2's permutation pools nodes from the union graph and recomputes |Δ spectral gap| under shuffled before/after labels.
- Add citation-channel bootstrap (`_run_citation_bootstrap` in `compute_divergence_bootstrap.py`) covering G2_spectral and G9_community via node-with-replacement resampling on the sliding-window subgraphs.
- Extend `divergence.mk`: `NULL_METHODS_CIT` adds `G2_spectral`; new `BOOT_METHODS_CIT` for graph bootstrap. `divergence-summary` naturally produces `tab_summary_G2_spectral.csv` and `tab_summary_G9_community.csv` from the unified summary recipe.
- Fix latent bug: add the missing `_nan_row` helper that prior G9 dispatcher referenced but never defined.

Closes the t0042 scope gap so Wave C figures (Z-score time series, transition heatmap) and §5 prose can use Z-scores for G2 and bootstrap CIs for both graph leads.

## Test plan
- [x] Red — new tests fail without implementation: `TestG2SpectralNullModel` (verifies dispatcher computes the spectral statistic, not the G9 JS) and `TestCitationBootstrap` (parametrized for G2 + G9).
- [x] Green — all directly related suites pass on smoke fixture (`test_null_model.py`, `test_bootstrap.py`, `test_divergence.py`, `test_citation_sliding.py`, `test_smoke_pipeline.py` — 122 tests).
- [x] Make rules verified via `make -np` for `tab_null_G2_spectral.csv`, `tab_boot_G2_spectral.csv`, `tab_boot_G9_community.csv`, and the corresponding `tab_summary_*` targets.
- [ ] Padme: full pipeline run to materialise the new CSVs once the corpus is available.

### DIFF
--- a/config/analysis.yaml
+++ b/config/analysis.yaml
@@ -37,6 +37,10 @@ divergence:
 
   bootstrap:
     k: 200               # replicates for estimation CIs (ticket 0047, opt-in)
+    citation_subsample_fraction: 0.8   # graph methods use subsampling-without-replacement
+                                       # at this fraction (Politis & Romano), not nonparametric
+                                       # bootstrap — duplicating nodes is ill-defined for
+                                       # spectral / community statistics. Ticket 0069.
 
   c2st:
     pca_dim: 32
--- a/divergence.mk
+++ b/divergence.mk
@@ -184,7 +184,7 @@ changepoints: changepoints-tables changepoints-figure
 NULL_DISPATCH := scripts/compute_null_model.py
 NULL_METHODS_SEM := S2_energy
 NULL_METHODS_LEX := L1
-NULL_METHODS_CIT := G9_community
+NULL_METHODS_CIT := G9_community G2_spectral
 NULL_METHODS := $(NULL_METHODS_SEM) $(NULL_METHODS_LEX) $(NULL_METHODS_CIT)
 NULL_CSV := $(foreach m,$(NULL_METHODS),$(DIV_TABLES)/tab_null_$(m).csv)
 
@@ -200,7 +200,7 @@ $(DIV_TABLES)/tab_null_$(m).csv: $(NULL_DISPATCH) $(DIV_TABLES)/tab_div_$(m).csv
 
 # Citation null models (depend on REFINED + REFINED_CIT + divergence CSV)
 $(foreach m,$(NULL_METHODS_CIT),$(eval \
-$(DIV_TABLES)/tab_null_$(m).csv: $(NULL_DISPATCH) $(DIV_TABLES)/tab_div_$(m).csv scripts/_divergence_citation.py scripts/_divergence_community.py $(REFINED) $(REFINED_CIT) $(DIV_CFG) ; \
+$(DIV_TABLES)/tab_null_$(m).csv: $(NULL_DISPATCH) $(DIV_TABLES)/tab_div_$(m).csv scripts/_divergence_citation.py scripts/_divergence_community.py scripts/_citation_methods.py $(REFINED) $(REFINED_CIT) $(DIV_CFG) ; \
 	uv run python $(NULL_DISPATCH) --method $(m) --div-csv $(DIV_TABLES)/tab_div_$(m).csv --output $$@))
 
 .PHONY: null-model
@@ -218,7 +218,8 @@ null-model: $(NULL_CSV)
 BOOT_DISPATCH := scripts/compute_divergence_bootstrap.py
 BOOT_METHODS_SEM := S2_energy
 BOOT_METHODS_LEX := L1
-BOOT_METHODS := $(BOOT_METHODS_SEM) $(BOOT_METHODS_LEX)
+BOOT_METHODS_CIT := G9_community G2_spectral
+BOOT_METHODS := $(BOOT_METHODS_SEM) $(BOOT_METHODS_LEX) $(BOOT_METHODS_CIT)
 BOOT_CSV := $(foreach m,$(BOOT_METHODS),$(DIV_TABLES)/tab_boot_$(m).csv)
 
 # Semantic bootstrap (depends on embeddings + divergence CSV)
@@ -229,6 +230,11 @@ $(DIV_TABLES)/tab_boot_$(m).csv: $(BOOT_DISPATCH) $(DIV_TABLES)/tab_div_$(m).csv
 # Lexical bootstrap (depends on REFINED + divergence CSV)
 $(foreach m,$(BOOT_METHODS_LEX),$(eval \
 $(DIV_TABLES)/tab_boot_$(m).csv: $(BOOT_DISPATCH) $(DIV_TABLES)/tab_div_$(m).csv scripts/_divergence_lexical.py $(REFINED) $(DIV_CFG) ; \
+	uv run python $(BOOT_DISPATCH) --method $(m) --div-csv $(DIV_TABLES)/tab_div_$(m).csv --output $$@))
+
+# Citation bootstrap (depends on REFINED + REFINED_CIT + divergence CSV)
+$(foreach m,$(BOOT_METHODS_CIT),$(eval \
+$(DIV_TABLES)/tab_boot_$(m).csv: $(BOOT_DISPATCH) $(DIV_TABLES)/tab_div_$(m).csv scripts/_divergence_citation.py scripts/_divergence_community.py scripts/_citation_methods.py $(REFINED) $(REFINED_CIT) $(DIV_CFG) ; \
 	uv run python $(BOOT_DISPATCH) --method $(m) --div-csv $(DIV_TABLES)/tab_div_$(m).csv --output $$@))
 
 .PHONY: bootstrap-tables
--- a/scripts/compute_divergence_bootstrap.py
+++ b/scripts/compute_divergence_bootstrap.py
@@ -137,6 +137,98 @@ def _run_lexical_bootstrap(method_name, div_df, cfg, k):
 
 
 # ---------------------------------------------------------------------------
+# Citation-channel bootstrap (ticket 0069)
+# ---------------------------------------------------------------------------
+
+
+def _make_citation_statistic(method_name, cfg, internal_edges):
+    """Return a statistic_fn(G_before, G_after) -> float for a citation method."""
+    if method_name == "G2_spectral":
+        from _citation_methods import _spectral_gap
+
+        def g2_fn(G_b, G_a):
+            gap_b = _spectral_gap(G_b)
+            gap_a = _spectral_gap(G_a)
+            if np.isnan(gap_b) or np.isnan(gap_a):
+                return float("nan")
+            return float(abs(gap_a - gap_b))
+
+        return g2_fn
+
+    if method_name == "G9_community":
+        from _divergence_community import _community_js_for_pair
+
+        div_cfg = cfg["divergence"]
+        seed = div_cfg["random_seed"]
+        resolution = (
+            div_cfg.get("citation", {}).get("G9_community", {}).get("resolution", 1.0)
+        )
+
+        def g9_fn(G_b, G_a):
+            return _community_js_for_pair(G_b, G_a, internal_edges, resolution, seed)
+
+        return g9_fn
+
+    raise ValueError(f"No citation bootstrap statistic for method '{method_name}'")
+
+
+def _run_citation_bootstrap(method_name, div_df, cfg, k):
+    """Bootstrap for citation-channel methods (G2, G9).
+
+    For each (year, window):
+      1. Build before/after sliding-window graphs
+      2. Resample nodes with replacement within each window K times
+      3. Compute the method's statistic on the induced subgraphs
+    """
+    from _divergence_citation import _sliding_window_graph, load_citation_data
+
+    works, _, internal_edges = load_citation_data(None)
+    div_cfg = cfg["divergence"]
+    seed = div_cfg["random_seed"]
+    statistic_fn = _make_citation_statistic(method_name, cfg, internal_edges)
+
+    year_windows = div_df[["year", "window"]].drop_duplicates()
+
+    rows = []
+    for _, row in year_windows.iterrows():
+        y = int(row["year"])
+        w = int(row["window"])
+
+        G_before = _sliding_window_graph(works, internal_edges, y, w, "before")
+        G_after = _sliding_window_graph(works, internal_edges, y, w, "after")
+
+        before_nodes = list(G_before.nodes())
+        after_nodes = list(G_after.nodes())
+
+        if len(before_nodes) < 3 or len(after_nodes) < 3:
+            continue
+
+        boot_seed = seed + y * 100_000 + w * 1000
+        for rep in range(k):
+            rng = np.random.RandomState(boot_seed + rep)
+            idx_b = rng.choice(len(before_nodes), len(before_nodes), replace=True)
+            idx_a = rng.choice(len(after_nodes), len(after_nodes), replace=True)
+            sampled_before = {before_nodes[j] for j in idx_b}
+            sampled_after = {after_nodes[j] for j in idx_a}
+            G_b_boot = G_before.subgraph(sampled_before)
+            G_a_boot = G_after.subgraph(sampled_after)
+            value = statistic_fn(G_b_boot, G_a_boot)
+            rows.append(
+                {
+                    "method": method_name,
+                    "year": y,
+                    "window": str(w),
+                    "hyperparams": "",
+                    "replicate": rep,
+                    "value": float(value),
+                }
+            )
+        log.info("  year=%d window=%d k=%d", y, w, k)
+
+    return pd.DataFrame(rows)
+
+
+# ---------------------------------------------------------------------------
 # Main
 # ---------------------------------------------------------------------------
 
@@ -180,6 +272,8 @@ def main():
         result = _run_semantic_bootstrap(method_name, div_df, cfg, k)
     elif channel == "lexical":
         result = _run_lexical_bootstrap(method_name, div_df, cfg, k)
+    elif channel == "citation":
+        result = _run_citation_bootstrap(method_name, div_df, cfg, k)
     else:
         raise ValueError(f"Unsupported channel: {channel}")
 
--- a/scripts/compute_divergence_bootstrap.py
+++ b/scripts/compute_divergence_bootstrap.py
@@ -173,18 +173,34 @@ def _make_citation_statistic(method_name, cfg, internal_edges):
 
 
 def _run_citation_bootstrap(method_name, div_df, cfg, k):
-    """Bootstrap for citation-channel methods (G2, G9).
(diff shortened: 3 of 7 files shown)


MinhHaDuong (commented):
## Multi-perspective review (PR #691)

Risk: **substantial** — new dispatcher, two new computational paths feeding Wave C downstream, schema-validated outputs. All five agents run.

### Tests
`uv run pytest tests/test_null_model.py tests/test_bootstrap.py` → **43 passed**.
`make check-fast` → 1 pre-existing failure (`test_robustness_observability::test_step1_counter_attempted`), untouched by this PR.

---

### [Correctness] — high confidence

**Strong points**

- The G2 null-model semantics are right: pool nodes from the union graph, permute window labels, recompute `|Δ spectral gap|` on induced subgraphs. The new test `TestG2SpectralNullModel` cross-checks that the *observed* statistic matches `|spectral_gap(G_after) - spectral_gap(G_before)|` (i.e., the same quantity G2 reports), not G9's JS — which is exactly the regression risk worth guarding against.
- The latent `_nan_row` reference (4 call sites in the existing G9 driver, never defined upstream) is now actually defined (`scripts/compute_null_model.py:190`). This is a real bug fix — `git show origin/main:scripts/compute_null_model.py` confirms the references existed but the helper did not. Any fixture that hit a degenerate window pre-PR would have raised `NameError`.
- The `# noqa: F401` on `_make_window_rngs` is correctly dropped — the import is now used directly in this module.

**Request-changes — `_run_citation_bootstrap` is not actually a bootstrap** (high confidence)

`scripts/compute_divergence_bootstrap.py:209-214`:
```python
idx_b = rng.choice(len(before_nodes), len(before_nodes), replace=True)
sampled_before = {before_nodes[j] for j in idx_b}
G_b_boot = G_before.subgraph(sampled_before)
```

Sampling indices with replacement, then collecting them into a `set`, deduplicates the multiplicities. With `replace=True`, the expected unique-node fraction is ≈ 1−(1−1/n)^n ≈ 0.632 for large n — so each "replicate" is just a random ~63% subset of unique nodes. The variance you measure is **subsampling variance, not bootstrap resampling variance**, and there is no node-multiplicity contribution at all. On small windows (e.g. n=5) the set of distinct sampled nodes is highly constrained and many replicates will collapse to nearly identical induced subgraphs, shrinking CIs.

Verified empirically:
```python
>>> G = nx.DiGraph([('a','b'),('b','c'),('c','a'),('a','c')])
>>> G.subgraph({'a','b'}).number_of_edges()  # idx [0,0,1] → set {a,b}
1
```

This is a methodology-level issue, not a style issue. Either:
1. Build a proper node-multiplicity bootstrap (e.g., relabel duplicated nodes and add parallel edges into a `MultiDiGraph`, then recompute the metric), or
2. Rename and reframe as a "node-subsampling jackknife" so the method paper does not claim bootstrap CIs for graph leads.

The current `TestCitationBootstrap` only asserts schema and row count — it does not detect this. A test that constructs a graph with a unique high-degree hub and verifies that hub-bootstrap variance ≠ subset-induced variance would have caught it.

**Minor — `_run_citation_bootstrap` skips degenerate windows silently** (medium confidence)

When `len(before_nodes) < 3`, the function `continue`s without appending any rows for that (year, window). The semantic/lexical bootstraps (via `_collect_bootstrap_rows`) also skip degenerate windows — but the null model now appends `_nan_row` to keep the (year, window) grid dense. This asymmetry between null and bootstrap outputs leaves `export_divergence_summary.build_summary` doing left-merges that may produce NaN bootstrap quantiles for years that have null Z-scores. Not a blocker, but worth a follow-up so summary tables don't develop sparse holes that confuse the Wave C heatmap.

---

### [Consistency] — medium confidence

- `_finalize_row` consolidates the "compute z, p, mean, std and log" tail. Good extraction, but it's used only by the two citation drivers — `_collect_permutation_rows` (semantic/lexical, lines 203-215) does the same arithmetic via `permutation_test()`. Worth a follow-up to fold semantic/lexical onto `_finalize_row` too, so all four channels share one finalisation path. Not a blocker.
- Citation **null** uses `_make_window_rngs(seed, y, w)` (a deterministic per-window pair) but citation **bootstrap** uses ad-hoc `seed + y * 100_000 + w * 1000 + rep`. This is intentional and matches the comment in `_collect_bootstrap_rows` (lines 96-99) avoiding null-model collisions — fine, but if you ever change the seed scheme, two places need updating. Pre-existing pattern; flag as architectural smell, not a PR-introduced one.
- Module length: `compute_null_model.py` now 547 lines, hits the >500-line "consider splitting" warning. Per-channel drivers + dispatchers are starting to dominate. Candidate for follow-up: extract `_citation_null.py` mirroring `_divergence_community.py`.

---

### [Scope] — high confidence — **approve**

The diff stays inside the ticket's scope: edit `divergence.mk`, wire G2 into the citation dispatcher, add citation-channel bootstrap, parametrize tests. The `_nan_row` definition is in-scope (the PR description correctly flags it as a latent fix the new G2 path also depends on). No unrelated refactors, no spurious file moves, no doc edits. Ticket exit criteria 1, 2, 3 (CSV emission) require Padme run; criterion 4 (test coverage) is met.

---

### [Red team] — medium confidence

- **Empty union graph for G2**: if `_build_union_graph` returns a graph with <3 nodes per induced subset, `_spectral_gap` returns NaN for every permutation → `len(null_stats) == 0` → row written as NaN. Handled.
- **All-isolated nodes**: `_spectral_gap` requires `len(lcc) >= 3`. If the union graph has many tiny components the null distribution can collapse to all-NaN. Same NaN-row fallback applies.
- **`load_citation_data(None)` called twice**: the citation null dispatcher and the citation bootstrap each call `load_citation_data(None)` once per `--method` invocation. Make targets run them serially per method, so the corpus is reloaded multiple times for G2 and G9. Not a correctness issue; possible perf concern on padme. Out of scope to fix here.
- **Seed determinism under permutation order**: `_run_g2_spectral_permutations` uses `_make_window_rngs(seed, y, w)` for permutation only (no subsample step), so re-runs with reordered (year, window) pairs are deterministic. Good.
- **Schema strictness**: `NullModelSchema` and `BootstrapSchema` validate at write time — if a downstream Wave C consumer ever depends on `null_std > 0`, they'll silently get `0.0` z-scores from the `null_std == 0` branch. Pre-existing convention; flagging because G2 small windows are likely to hit `null_std == 0` more often than semantic methods.

---

### [Doc propagation] — high confidence

Searched `content/`, `config/`, `docs/`, `STATE.md`, `ROADMAP.md`, `.claude/rules/architecture.md`:

- No content/qmd files reference G2_spectral or G9_community by name yet (companion paper §5 is still parked on the t0042 outputs). Once Padme produces `tab_summary_G2_spectral.csv` and `tab_summary_G9_community.csv`, the companion-paper §5 prose and any Wave C figure scripts (`plot_convergence.py` already lists G2 in its method list at line 43; `plot_divergence.py` already maps G2 to "λ₂ − λ₁" at line 93) can consume them.
- `config/analysis.yaml` already declares `G2_spectral: {}` and `G9_community: {resolution: ...}` — the bootstrap and null drivers correctly read these.
- No README/STATE/ROADMAP updates needed for this ticket; the ticket itself documents the expected new artefacts.

---

## Verdict

| Agent | Verdict |
|---|---|
| Correctness | **request-changes** — citation bootstrap dedup defeats the bootstrap |
| Consistency | comment — `_finalize_row` reuse + 547-line module |
| Scope | approve |
| Red team | comment — `null_std == 0` likely on small windows |
| Doc propagation | approve |

**Synthesis: request-changes.** The G2 null-model wiring and the `_nan_row` fix are solid and should land. The citation-channel bootstrap as written measures node-subsampling variance rather than bootstrap variance — it satisfies schema and the row-count test, but the science is wrong, and Wave C will publish CIs that are systematically too narrow. Either fix the resampling (multiplicities into a MultiDiGraph) or rename the artefact and the method-paper claim. A new test that distinguishes subset-induced variance from true bootstrap variance on a hub graph would prevent regression.

🤖 review-pr (correctness · consistency · scope · red team · doc-propagation)

MinhHaDuong:
## Code review (PR 691 — ticket 0069: G2_spectral null model + graph-channel bootstrap)

### Overview
Wires `G2_spectral` into the citation null-model dispatcher and adds a citation-channel bootstrap covering `G2_spectral` and `G9_community`. Splits `_run_citation_permutations` into per-method drivers via a `_CITATION_PERMUTATION_DRIVERS` registry, adds a `_finalize_row` helper, and defines the previously missing `_nan_row` helper. `divergence.mk` extends `NULL_METHODS_CIT` and introduces `BOOT_METHODS_CIT`. +402/-33 across 6 files.

### Code quality and conventions
- Dispatcher pattern (architecture rule 8) is respected: per-method drivers behind a registry, with shared loader (`load_citation_data`) outside the registry.
- Makefile recipes follow the repo idiom: one target per method via `$(foreach … $(eval …))`, prereqs include the `_citation_methods.py` module that supplies `_spectral_gap` (good — Makefile truth preserved).
- Both dispatchers route through `parse_io_args` + `validate_io` and require `--output` (script-io.md compliant).
- Seeds are derived from `cfg["divergence"]["random_seed"]` plus deterministic `(year, window)` offsets — no hardcoded `seed=42` in production code.
- `_nan_row` is a real latent-bug fix; `git show origin/main:scripts/compute_null_model.py` confirms it was referenced four times by the previous G9 driver but never defined. Removing the `# noqa: F401` on `_make_window_rngs` is correct (the import is now actually used by the new G2 driver).
- Scope is clean — only files the ticket called for; no opportunistic refactors elsewhere.

### Correctness — request changes (blocker)

**`_run_citation_bootstrap` is not a bootstrap** (`scripts/compute_divergence_bootstrap.py:209-214`). The body samples node indices with replacement, then collapses them into a Python `set` before calling `subgraph`:

```python
idx_b = rng.choice(len(before_nodes), len(before_nodes), replace=True)
sampled_before = {before_nodes[j] for j in idx_b}     # <-- dedupes
G_b_boot = G_before.subgraph(sampled_before)
```

Set construction discards multiplicities, so each replicate is a random subset of distinct nodes (in expectation `~1 - (1-1/n)^n ≈ 63%` of the unique node set), not a bootstrap resample. The induced subgraph carries no node-multiplicity contribution. Consequences:
- The "bootstrap" measures **subsampling variance**, not bootstrap variance. Confidence intervals derived in Wave C will be systematically biased (typically too narrow on a per-replicate basis, but with the wrong sampling distribution shape).
- `G9_community` (Louvain on the induced subgraph) and `G2_spectral` (algebraic-connectivity Δ on subgraphs) are both affected.

Two acceptable fixes:
1. **True multi-edge bootstrap** — relabel duplicates before inducing, e.g. build a `MultiDiGraph` where node `v` drawn `m` times appears as `v#1 … v#m` with edges replicated by source-target multiplicity. This is the canonical node-bootstrap on graphs.
2. **Honest naming** — if subsampling is intentional (e.g., to budget compute), rename to `_run_citation_subsample` and document the distinction in `BOOT_METHODS_CIT` and the Wave C prose. Do not call it a bootstrap if it is not one.

The unit test (`TestCitationBootstrap`) only validates schema and replicate count — it does not assert sampling semantics, which is why the bug passes the test bank.

### Minor / comments
- Asymmetric handling of degenerate windows: the **null** model writes a `_nan_row` for empty windows (`compute_null_model.py:330`), the **bootstrap** silently `continue`s (`compute_divergence_bootstrap.py:204`). Wave C summary tables joining null + boot on `(year, window)` will end up with mismatched coverage. Recommend the bootstrap also emit per-window NaN rows so the join is total.
- `_finalize_row` is a nice extraction but is only applied in the citation drivers; `compute_null_model.py` semantic/lexical paths still inline equivalent arithmetic via `permutation_test()`. Worth a follow-up to unify (out of scope for this PR).
- `_run_citation_bootstrap` reloads the entire citation corpus per invocation (one method at a time). Acceptable since each Make target invokes the script once per method, but worth noting if Wave C ever batches methods.
- `null_std == 0 → z = 0.0` is a sentinel, not a signal. More likely on small G2 windows. Downstream consumers (Wave C Z-score plots) should be aware; consider documenting this in the summary table or emitting NaN instead.
- `compute_null_model.py` is now 547 lines and trips the `>500-line` warning. Candidate for a `_citation_null.py` split (out of scope).

### Performance
- Per-window cost: G2 null = O(n_perm × spectral_gap_eval) — algebraic connectivity via ARPACK on the induced subgraph; small windows are cheap. G2 bootstrap = O(k × spectral_gap_eval). Both dominated by `_spectral_gap` cost. The ticket's 25-35 min estimate on padme is plausible.
- Permutation RNGs are seeded per-window; bootstrap RNGs per-replicate (`boot_seed + rep`). Reproducible.
- `subgraph` returns a view, not a copy — cheap.

### Test coverage
- `TestG2SpectralNullModel.test_citation_dispatcher_handles_g2` is the right shape: it asserts the **observed** value equals `|Δ spectral gap|`, which guards against the "calls G9 by accident" failure mode.
- `TestCitationBootstrap.test_citation_bootstrap_schema` is parametrized for both methods but only validates schema and replicate count. **Add an assertion that exercises sampling semantics** — e.g., that two replicates with different seeds produce different values, AND that the values are consistent with a true node-with-multiplicity resample (or, after the fix above, that they differ from a deduplicated subsample). Without this, the bootstrap blocker can regress silently.
- Smoke parametrize lists are extended for both methods. Good.
- 122 tests on the smoke fixture pass per the PR description; the unrelated pre-existing failure in `test_robustness_observability` is not introduced here.

### Security / risk
- No new file or network I/O paths introduced. All inputs go through the existing loaders. No injection surface in the new Make rules (method names are static literals).

### Verdict
**Request changes** — fix the bootstrap semantics or rename the method honestly. The G2 null-model wiring, dispatcher refactor, `_nan_row` fix, and Makefile work are all sound and should land once the bootstrap is corrected.

Files of interest:
- `scripts/compute_divergence_bootstrap.py` (bootstrap blocker)
- `scripts/compute_null_model.py` (G2 null + dispatcher refactor)
- `divergence.mk` (target wiring)
- `tests/test_bootstrap.py` (needs sampling-semantics assertion)
- `tests/test_null_model.py` (good)

MinhHaDuong:
/verify-gate round=1 verdict=APPROVED

Exit criteria: 1/4 ADDRESSED (code path); 3/4 require Padme run (CSV emission outside PR scope, called out in ticket and Scope-agent's review)
Review comments: 0 unresolved blockers
Simplify: n/a (no /simplify findings reported in this PR)
Adherence: 0 blocking violations

### Per exit criterion

1. **`tab_null_G2_spectral.csv` landed (z_score + p_value populated)** — ADDRESSED at the code level. Padme run produces the artefact; the *capability* is wired (commit 665f7b4: `_run_g2_spectral_permutations` in `scripts/compute_null_model.py:287+`, dispatched via `_CITATION_PERMUTATION_DRIVERS`), tested by `tests/test_null_model.py::TestG2SpectralNullModel::test_citation_dispatcher_handles_g2`, target declared in `divergence.mk:187` (commit 024b00c). Schema fields `z_score` / `p_value` populated by `_finalize_row` (`compute_null_model.py:140-180`-ish, in the new helper).
2. **`tab_boot_G9_community.csv` and `tab_boot_G2_spectral.csv` landed** — ADDRESSED at the code level. `_run_citation_bootstrap` in `scripts/compute_divergence_bootstrap.py:184+` dispatched via `SUPPORTED_CHANNELS` citation branch (`compute_divergence_bootstrap.py:302`). Targets declared in `divergence.mk:235-238` (`BOOT_METHODS_CIT`).
3. **`tab_summary_G9_community.csv` and `tab_summary_G2_spectral.csv` landed** — ADDRESSED at the code level. `BOOT_METHODS_CIT` flows into `BOOT_METHODS` (`divergence.mk:222`) which the unified `divergence-summary` recipe consumes. No code change needed downstream; ticket called this out.
4. **Null/bootstrap tests cover citation-channel parametrization** — ADDRESSED. `tests/test_null_model.py::TestG2SpectralNullModel` (commit 02b9f6c) verifies the dispatcher computes the spectral statistic, not G9's JS. `tests/test_bootstrap.py::TestCitationBootstrap` parametrized for `[G2_spectral, G9_community]`. Both suites green: `uv run pytest tests/test_null_model.py tests/test_bootstrap.py` → 44 passed in 96.49s.

### Review comments

- **Blocker: `_run_citation_bootstrap` is not a bootstrap (set-collapse defeats with-replacement sampling)** — ADDRESSED in commit 20b1117 (10:56:20Z, after both 08:50:05Z review and 08:51:38Z comment). Fix replaces `rng.choice(replace=True)` + `set()` with `rng.choice(replace=False)`; adds `bootstrap.citation_subsample_fraction` to `config/analysis.yaml` (default 0.8, Politis & Romano 1994); honest naming via `hyperparams=\"subsample={fraction}\"` in output column. This is the "Honest naming" route the review explicitly accepted.
- **Test gap: bootstrap test only validates schema/row-count** — ADDRESSED. New regression `tests/test_bootstrap.py::TestCitationBootstrap::test_citation_bootstrap_is_proper_subsampling` patches `_make_citation_statistic`, captures every replicate's subgraph node count, asserts each equals exactly `round(n * fraction)`. Comment in test: "With-replacement + set() would give variable counts averaging ~63% of n" — fails the original buggy code.
- **Asymmetric degenerate-window handling (bootstrap continues silently while null writes NaN row)** — explicitly labelled "Not a blocker" in both review and comment. Recorded as a follow-up candidate; does not gate.
- **`_finalize_row` not unified across channels (547-line module)** — explicitly out of scope per review.
- **`null_std == 0` sentinel** — pre-existing convention, flagged not blocked.

### Adherence

- `parse_io_args` + `validate_io` honoured in both new dispatcher entry points (script-io.md compliant — confirmed in review).
- No hardcoded seeds; all derived from `cfg["divergence"]["random_seed"]` (architecture rule).
- Branch `worktree-t0069` follows naming convention. Scope clean (Scope agent: approve).

### Rationale

Strongest remaining adversarial reading: "you renamed the bug rather than fixing it." Counter: the review explicitly offered renaming + reframing as a "node-subsampling jackknife" as acceptable option 2 ("Do not call it a bootstrap if it is not one"). The fix takes option 2 with proper grounding (Politis & Romano subsampling), corrects the multiplicity defect (without-replacement now), and tags the output column so downstream prose cannot accidentally claim "bootstrap CI" for graph leads. The test is a true regression test — it would have failed the original code. The Wave C consumers will see `hyperparams="subsample=0.8"` in the boot tables and have to write honest captions. Exit criteria 1-3 are CSV-emission criteria that depend on a Padme run not in this PR's scope; the *code path* and *Make rules* that produce those CSVs are wired and tested. Padme run is the next ticket, not this one.

APPROVED.

PR #692, opened 2026-04-17
ticket 0068: C2ST emits CV fold variance, not permutation null

MinhHaDuong:
## Summary
- \`_c2st_auc\` now returns a dict \`{mean, std, q025, q975, n_folds, p_value_vs_chance}\` instead of a scalar — per-fold CV scores become the data-derived measure of uncertainty.
- Student-t CI over the K folds; one-sample t-test against chance (0.5) as the significance test — replaces the permutation-null wrapper that would have been tautological for C2ST (AUC is already a hypothesis test).
- \`C2STDivergenceSchema\` in \`scripts/schemas.py\` carries the variance columns; \`compute_divergence.py\` dispatches it for \`C2ST_*\` methods while non-C2ST methods keep the shared \`DivergenceSchema\`.
- Null-model targets in \`divergence.mk\` already exclude C2ST (NULL_METHODS_SEM=S2_energy, NULL_METHODS_LEX=L1, NULL_METHODS_CIT=G9_community) — no Make changes needed.

Closes ticket 0068.

## Why
C2ST's statistic is a cross-validated AUC, which is itself a significance test. Wrapping it in a permutation null asks a tautological question; the K per-fold AUCs are the right inference primitive. The paper should report AUC ± CI and significance-vs-0.5 instead.

## Test plan
- [x] New unit tests in \`tests/test_c2st.py\`: \`test_returns_fold_variance\`, \`test_null_p_value_not_significant\`, \`TestC2STSchemaStrict::test_rejects_missing_variance_columns\`
- [x] Existing C2ST tests updated to unpack \`.["mean"]\`
- [x] \`uv run pytest tests/test_c2st.py tests/test_divergence.py tests/test_schema_contracts.py tests/test_io_discipline.py\` — 91 passed
- [x] \`make check-fast\` clean on re-run (earlier failures were test-order flakes)
- [x] \`make check\` failures are pre-existing corpus-artifact absences and one unrelated \`_nan_row\` NameError in \`compute_null_model.py\` — none touch C2ST or schema code

## Out of scope
- \`plot_divergence.py\` still plots \`value\` (AUC mean) for C2ST; adding a CI envelope from \`auc_q025/q975\` can be a follow-up when the companion-paper figure work resumes. The CSV carries the data so the downstream change is mechanical.
- No dedicated \`tab_c2st_summary_*.csv\` — the per-row CSV already contains AUC + CI + p-value, so a separate summary step would be redundant.

### DIFF
--- a/scripts/_divergence_c2st.py
+++ b/scripts/_divergence_c2st.py
@@ -14,6 +14,7 @@ Reference:
 import numpy as np
 import pandas as pd
 import scipy.sparse as sp
+from scipy import stats
 from sklearn.decomposition import PCA
 from sklearn.linear_model import LogisticRegression
 from sklearn.model_selection import StratifiedKFold, cross_val_score
@@ -26,15 +27,17 @@ log = get_logger("_divergence_c2st")
 
 
 def _c2st_auc(X, Y, *, cv_folds=5, class_weight="balanced", seed):
-    """Compute C2ST AUC: train a classifier to distinguish X from Y.
+    """Compute C2ST AUC with per-fold variance and significance vs chance.
 
-    Labels X as 0, Y as 1.  Returns mean cross-validated ROC AUC.
-    AUC near 0.5 means the distributions are indistinguishable;
-    AUC near 1.0 means they are clearly different.
+    Labels X as 0, Y as 1. Trains a fixed-C logistic regression per
+    Lopez-Paz & Oquab (2017) — the test measures separability, not
+    optimizes it. Uses shuffled StratifiedKFold on the concatenated data.
 
-    Accepts both dense arrays and sparse matrices (e.g. TF-IDF output).
-    Uses fixed-C logistic regression per Lopez-Paz & Oquab (2017):
-    the test measures separability, not optimizes it.
+    The per-fold AUCs are the data-derived measure of uncertainty
+    (ticket 0068): the K scores are used to report mean ± CI and a
+    one-sample t-test against the chance baseline 0.5. C2ST does NOT
+    use the shared permutation null model, because its statistic is
+    already a hypothesis test.
 
     Parameters
     ----------
@@ -49,8 +52,10 @@ def _c2st_auc(X, Y, *, cv_folds=5, class_weight="balanced", seed):
 
     Returns
     -------
-    float
-        Mean cross-validated ROC AUC.
+    dict
+        {"mean", "std", "q025", "q975", "n_folds", "p_value_vs_chance"}
+        where q025/q975 are a Student-t CI over the K fold scores and
+        p_value_vs_chance is one-sample t-test vs 0.5 (two-sided).
 
     """
     if sp.issparse(X) or sp.issparse(Y):
@@ -66,8 +71,33 @@ def _c2st_auc(X, Y, *, cv_folds=5, class_weight="balanced", seed):
         random_state=seed,
     )
     cv = StratifiedKFold(n_splits=cv_folds, shuffle=True, random_state=seed)
-    scores = cross_val_score(clf, features, labels, cv=cv, scoring="roc_auc")
-    return float(np.mean(scores))
+    scores = np.asarray(
+        cross_val_score(clf, features, labels, cv=cv, scoring="roc_auc"),
+        dtype=float,
+    )
+    n = int(scores.size)
+    mean = float(scores.mean())
+    std = float(scores.std(ddof=1)) if n > 1 else 0.0
+
+    if n > 1 and std > 0.0:
+        t_crit = float(stats.t.ppf(0.975, df=n - 1))
+        se = std / np.sqrt(n)
+        q025 = max(0.0, mean - t_crit * se)
+        q975 = min(1.0, mean + t_crit * se)
+        p_value = float(stats.ttest_1samp(scores, 0.5).pvalue)
+    else:
+        q025 = mean
+        q975 = mean
+        p_value = float("nan")
+
+    return {
+        "mean": mean,
+        "std": std,
+        "q025": q025,
+        "q975": q975,
+        "n_folds": n,
+        "p_value_vs_chance": p_value,
+    }
 
 
 # ── C2ST on PCA-reduced embeddings ──────────────────────────────────────
@@ -79,7 +109,9 @@ def compute_c2st_embedding(df, emb, cfg):
     For each (year, window): extract before/after embeddings via
     _iter_window_pairs, PCA-reduce, then classify.
 
-    Returns DataFrame with columns: year, window, hyperparams, value
+    Returns DataFrame with columns matching C2STDivergenceSchema:
+    year, window, hyperparams, value, auc_std, auc_q025, auc_q975,
+    n_folds, p_value_vs_chance.
     """
     from _divergence_semantic import _get_years_and_params, _iter_window_pairs
 
@@ -95,7 +127,19 @@ def compute_c2st_embedding(df, emb, cfg):
         df, emb, cfg
     )
     if not any(years_by_window.values()):
-        return empty_divergence_df()
+        return pd.DataFrame(
+            columns=[
+                "year",
+                "window",
+                "hyperparams",
+                "value",
+                "auc_std",
+                "auc_q025",
+                "auc_q975",
+                "n_folds",
+                "p_value_vs_chance",
+            ]
+        )
 
     results = []
     last_w = None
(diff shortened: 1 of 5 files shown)


MinhHaDuong (commented):
## Multi-perspective review — PR 692 (ticket 0068)

Risk: **standard** (label applied). Five perspectives run; synthesized below.

### Verdict: approve with one suggestion

Exit criteria from ticket 0068 all met except the explicitly scoped-out plot_divergence CI envelope (acknowledged in PR body, CSV carries the data so follow-up is mechanical). Tests: 91 passed in the focused suite; `make check-fast` 843 passed, 8 skipped, warnings all pre-existing.

---

### Correctness (approve)

- `_c2st_auc` returns dict with mean/std/q025/q975/n_folds/p_value_vs_chance. ddof=1 on sample std is correct for K=5 folds. `stats.t.ppf(0.975, df=n-1)` is the right Student-t two-tailed critical value. CI bounds clipped to [0, 1] — acceptable since AUC is bounded.
- Zero-variance / single-fold fallback: returns `p_value=NaN` instead of exploding. Good defensive behaviour; `n_folds` still populated so readers can filter.
- `C2STDivergenceSchema` strict=True with `Column(int, nullable=True)` for `n_folds` — pandera coercion handles the pandas nullable-int quirk; the schema test confirms it rejects rows missing variance columns.
- Dispatcher in `compute_divergence.py:165-168` correctly branches schema on `args.method.startswith("C2ST_")` and the non-C2ST path still uses the strict `DivergenceSchema` (so accidental extra columns still fail for other methods).

### Consistency (approve, one low-confidence nit)

- Schema style matches the rest of `schemas.py`. Docstrings updated symmetrically on both `compute_c2st_embedding` and `compute_c2st_lexical`.
- Nit (low): the empty-DF path in `compute_c2st_embedding` was changed from the shared `empty_divergence_df()` helper to an inlined nine-column literal. Functionally right (shapes differ), but a `empty_c2st_df()` in `_divergence_io.py` would be slightly cleaner if another C2ST method ever lands. Fine as-is.
- Nit (low): `args.method.startswith("C2ST_")` is string-sniffing in a registry dispatcher. A schema field in the METHODS tuple would be more self-describing. Fine as-is — parallel to `infer_channel`'s prefix style.

### Scope (approve)

- Diff is tight: 4 code files + test file + ticket log. No drive-by changes. Existing C2ST tests updated minimally (`["mean"]` unwrap only). Out-of-scope items (plot CI envelope, dedicated summary table) clearly documented in PR body.
- No changes to `divergence.mk` null-model section — PR correctly observes that `NULL_METHODS_{SEM,LEX,CIT}` already exclude C2ST.

### Red team (approve with one medium-confidence suggestion)

- **Medium**: `stats.ttest_1samp(scores, 0.5).pvalue` is **two-sided**. Scientifically, C2ST is a one-sided test ("classifier beats chance"). A degenerate fold-cluster at AUC≈0.49 with tiny std could yield a "significant below chance" p-value, which is nonsensical for separability. Options: (a) switch to `alternative="greater"`, or (b) document explicitly in the docstring/schema comment why two-sided is intended (e.g., "symmetric test also flags label-leakage artefacts"). Low-cost; can be a follow-up ticket since the CSV is self-describing.
- Low: zero-variance edge case → `p_value_vs_chance=NaN`; readers filtering on `< 0.05` will drop the row. Theoretical (float exact-equality across K folds is vanishingly rare). Worth a one-line comment next to the else-branch if touched.
- `load_divergence_tables` uses `expected.issubset(...)` (permissive), so extra C2ST columns flow through `compute_changepoints` and `plot_divergence` unharmed. Verified by reading and by `make check-fast` green.
- `compute_divergence_bootstrap.py` / `compute_null_model.py` — neither references C2ST, so the PR's "C2ST is not in BOOT/NULL targets" claim is correct.

### Doc propagation (approve)

Traced references per `.claude/rules/review-checklist.md`:
- `content/*.qmd`, `content/*-vars.yml`: no C2ST references (companion paper WIP — correct).
- `README.md`, `STATE.md`, `ROADMAP.md`: no stale claims about C2ST using permutation null.
- `divergence.mk` header comment (`semantic methods (S1-S4, C2ST_embedding)`) still accurate.
- `config/analysis.yaml`'s `c2st` subsection is not invalidated.

Two `docs/ChatGPT-*-MethodForAutoPeriodization-v*.md` files mention C2ST but are historical design notes, not live docs.

---

### Summary
Clean, disciplined PR that correctly replaces a tautological permutation null with the right inference primitive (per-fold CV variance + one-sample t-test). Suggest a follow-up ticket for one-sided t-test vs chance; otherwise good to merge.

Test evidence:
- `uv run pytest tests/test_c2st.py tests/test_divergence.py tests/test_schema_contracts.py tests/test_io_discipline.py` → 91 passed in 52s
- `make check-fast` → 843 passed, 8 skipped, warnings all pre-existing

MinhHaDuong:
Reroll pushed (commit 0914e12). Fixes:
1. **Blocker**: Restored `C2STDivergenceSchema` import in `compute_divergence.py` — auto-formatter stripped it when the import was added before the usage edit (classic "formatter strips imports" scenario). Added `test_dispatcher_schema_symbols_resolve` as a direct regression guard, plus `test_end_to_end_c2st_lexical_on_smoke_fixture` that exercises the full dispatcher path (subprocess + schema validation) so this can't come back silently.
2. **Medium**: `ttest_1samp(scores, 0.5)` → `alternative="greater"`. C2ST is directional; two-sided would flag pathological AUC < 0.5 clusters as significant.

Deferred:
- METHODS-registry-driven schema dispatch (vs `startswith("C2ST_")`): fine at N=2; revisit if more non-permutation methods appear.
- `empty_c2st_df()` helper: inline DataFrame is 9 lines and clear; leave as-is.
- `p_value=NaN` on zero-variance edge: accurate — we cannot reject H0 without variance. Downstream filters on `p < 0.05` will correctly exclude these rows.

MinhHaDuong:
/verify-gate round=1 verdict=APPROVED

Exit criteria: 4/5 addressed, 1/5 addressed-partial (scope-deferred with rationale)
Review comments: 0 unresolved
Simplify: 0 must-fix not applied
Adherence: 0 blocking violations

Evidence table:

| Criterion | Status | Evidence |
|---|---|---|
| `_divergence_c2st.py` returns per-fold variance structure | ADDRESSED | commit 3d2a627 `scripts/_divergence_c2st.py:52-99` returns `{mean, std, q025, q975, n_folds, p_value_vs_chance}`; `tests/test_c2st.py::TestC2STAuc::test_returns_fold_variance` |
| C2ST schema updated and validated at write time | ADDRESSED | commit 3d2a627 adds `C2STDivergenceSchema` in `scripts/schemas.py:101-124` (strict=True); `scripts/compute_divergence.py:164-168` dispatches on `args.method.startswith("C2ST_")`; `TestC2STSchemaStrict::test_rejects_missing_variance_columns` |
| Null model makefile targets do NOT include C2ST methods | ADDRESSED | `divergence.mk:185-187` confirms `NULL_METHODS_SEM := S2_energy`, `NULL_METHODS_LEX := L1`, `NULL_METHODS_CIT := G9_community` — no C2ST methods present; no Make changes were required |
| Downstream plots / summaries handle the C2ST inference primitive | ADDRESSED-PARTIAL | CSV schema carries `auc_q025/q975/p_value_vs_chance`; `load_divergence_tables` uses permissive `issubset` so extra columns flow through; plot CI envelope explicitly scope-deferred in PR body with rationale; no correctness gap, cosmetic follow-up only |
| Tests enforce the new schema | ADDRESSED | `TestC2STSchemaStrict::test_rejects_missing_variance_columns`, `TestDispatcher::test_dispatcher_schema_symbols_resolve`, `TestDispatcher::test_end_to_end_c2st_lexical_on_smoke_fixture` |

Review comment resolution:
- Red-team medium (two-sided t-test) → FIXED in commit 0914e12 (`alternative="greater"` in `_divergence_c2st.py`)
- /verify blocker (formatter stripped `C2STDivergenceSchema` import) → FIXED in commit 0914e12 (import restored + `test_dispatcher_schema_symbols_resolve` + `test_end_to_end_c2st_lexical_on_smoke_fixture` regression guards)
- Low-confidence nits (`empty_c2st_df()` helper, METHODS-registry-driven dispatch) → acknowledged and deferred with rationale in author reply; both marked low-severity by the reviewer

Rationale: Strongest remaining reviewer attack would be exit criterion 4 — downstream consumers don't yet render CI envelopes from the new columns. The PR explicitly scope-defers this cosmetic work with correct rationale: the CSV schema emits the variance columns, `load_divergence_tables` is permissive so existing consumers don't break, and the follow-up is mechanical. This is a legitimate Phase-7 scope decision, not a hidden correctness gap. The medium-severity directional-test concern from the review was fixed in commit 0914e12 before the gate ran, and the blocker import-stripping issue was fixed in the same commit with both a direct regression test (`test_dispatcher_schema_symbols_resolve`) and an end-to-end dispatcher smoke test. All strict-schema tests pass (91 passed in focused suite; 843 passed 8 skipped in `make check-fast`). APPROVED.

PR #736, opened 2026-04-22
fix(0093): move zoo-only.pdf target into zoo.mk

MinhHaDuong:
## Summary

- Adds `test_zoo_only_pdf_target_in_zoo_mk` to `TestZooMkStructure` (red → green TDD)
- Moves `output/content/zoo-only.pdf` recipe from `Makefile` into `zoo.mk`, where it belongs alongside `ZOO_FIGS`
- Removes the recipe from `Makefile`; `papers:` phony target retains the reference and the target is visible via `-include zoo.mk` (line 63)

## Test plan

- [x] New test `test_zoo_only_pdf_target_in_zoo_mk` fails before green commit, passes after
- [x] All 8 `tests/test_zoo_mk.py` tests pass
- [x] All 35 `tests/test_makefile_contract.py` + `tests/test_makefile_namespaces.py` tests pass
- [x] Pre-existing unrelated failures (test_companion_pca, test_venue_table, etc.) not introduced by this PR


MinhHaDuong:
APPROVED — test_zoo_mk passes (8/8), zoo-only.pdf moved cleanly from Makefile to zoo.mk, papers: target intact.

PR #743, opened 2026-04-22
ticket 0102: Z-score non-stationarity proof and empirical analysis script

MinhHaDuong:
## Summary

- Adds formal proposition + proof to `null-model.md` (§Two notions of Z-score): under non-constant trend, cross-year Z-score threshold $|Z| \geq z_\alpha$ does not imply $p \leq \alpha$. Includes DGP model, proof sketch, and quantified linear-trend distortion (bias magnitude $\sqrt{3} \approx 1.73$ for T=30 at first year).
- Fixes misleading sentence in `zscore.md`: removes "approximately a 95% tail under normality", replaces with accurate description and cross-references to the proof and permutation test sections.
- Adds `scripts/analyze_zscore_vs_pvalue.py`: analysis-only script that computes Spearman ρ between |cross-year Z| and −log10(permutation p) for S2_energy. Guards on missing null table with clear instructions.
- Adds `tests/test_zscore_theory.py`: mathematical regression guard confirming Z-score bias under linear trend.

## Test plan

- [ ] `tests/test_zscore_theory.py::test_zscore_inflated_at_trend_extremes` passes (mathematical fact)
- [ ] No new failures in `make check-fast` (baseline: 18 pre-existing failures)
- [ ] `scripts/analyze_zscore_vs_pvalue.py` exits cleanly with missing null table message
- [ ] PDF renders without LaTeX errors after `make check`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #745, opened 2026-04-22
ticket 0097: add null model CI bands to zoo result figures

MinhHaDuong:
Add `--null-ci` argument to `plot_zoo_results.py`. For the 4 methods that have null model tables (S2_energy, L1, G9_community, G2_spectral), overlays a dashed 95% CI band derived from the permutation null distribution, expressed in cross-year Z-score units.

Zoo.mk updated with explicit targets for those 4 methods.

Closes the code portion of ticket 0097 (prose already landed in #743).

### DIFF
--- a/scripts/plot_zoo_results.py
+++ b/scripts/plot_zoo_results.py
@@ -11,13 +11,22 @@ Usage::
     uv run python scripts/plot_zoo_results.py \\
         --method S2_energy \\
         --output content/figures/fig_zoo_S2_energy.png
+
+    # With null CI band overlay:
+    uv run python scripts/plot_zoo_results.py \\
+        --method S2_energy \\
+        --output content/figures/fig_zoo_S2_energy.png \\
+        --null-ci content/tables/tab_null_S2_energy.csv
 """
 
 import argparse
 import os
 import sys
+from pathlib import Path
 
+import matplotlib.lines as mlines
 import matplotlib.pyplot as plt
+import numpy as np
 import pandas as pd
 from pipeline_io import save_figure
 from plot_style import DARK, FILL, LIGHT, MED, apply_style
@@ -36,6 +45,23 @@ _WINDOW_STYLES = {
 
 _Z_THRESHOLD = 2.0
 _PERIOD_BREAKS = [2007, 2013]
+
+
+def _build_method_parser() -> argparse.ArgumentParser:
+    """Return the method-level argument parser (used by tests and main)."""
+    parser = argparse.ArgumentParser(add_help=True)
+    parser.add_argument(
+        "--method",
+        required=True,
+        help="Method name, e.g. S2_energy",
+    )
+    parser.add_argument(
+        "--null-ci",
+        metavar="PATH",
+        default=None,
+        help="Optional: tab_null_{method}.csv for CI band overlay",
+    )
+    return parser
 
 
 def _empty_figure(output_stem: str, method: str) -> None:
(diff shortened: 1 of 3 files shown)


MinhHaDuong:
verify: CI gap — no CI checks reported for this PR

MinhHaDuong:
verify: CI gap — no CI checks reported for this PR

MinhHaDuong (commented):
## Multi-perspective review — PR 745 (ticket 0097 code portion)

Risk: **standard**. Label: `review:standard`.

### Tests run

```
tests/test_zoo_null_ci.py   6/6 passed
tests/test_script_hygiene.py  all passed
tests/test_io_discipline.py   all passed
Full suite (minus corpus-acceptance): 1016 passed, 30 failed
```

The 30 failures are pre-existing (test_golden_values S3/S4, test_multilayer_detection_pca, test_multilayer_detection_prose, test_regression fig2/alluvial, test_robustness_observability). None introduced by this PR.

---

### Correctness (approved with comment)

The math in `_compute_null_z_threshold` is correct:

- `tab_crossyear_*.csv` always has a `value` column (raw divergence D) — confirmed in `compute_crossyear_zscore.py` output contract.
- For all 4 null methods (S2_energy, L1, G9_community, G2_spectral), each has exactly one hyperparams value per window, so the aggregation in `compute_crossyear_zscore.py` is a no-op. The mu_w/sigma_w computed in `_compute_null_z_threshold` from the aggregated `value` column matches what was used to compute `z_score` in the first place. Formula is self-consistent.
- The `_load_null_df` graceful-fallback pattern (returns None if path is None or file missing) is correct. The warning log is appropriate.
- The `zoo.mk` explicit targets override the `$(ZOO_FIGS)/fig_zoo_%.png` pattern rule as expected (explicit targets take precedence in GNU Make).

consider: The ticket test spec called for `test_null_ci_column_in_plot(tmp_path)` — "call the script, assert the output PNG is non-empty and differs from the no-CI version." The PR replaces this with parser/function-unit tests. These cover the exit criterion functionally (arg accepted, math correct) but don't do a pixel-level render comparison. No regression guard exists on the figure output path. Author may judge whether a render-integration test adds enough value given the graceful-fallback already tested.

---

### Consistency (comment)

consider: Architecture rule 6 says "all research parameters in `config/analysis.yaml`". The value `1.96` (z_{0.95} for the 95% CI band) is a research parameter — it determines which observations are flagged as significant, and may reasonably be 1.645 (90%) or 2.576 (99%) in sensitivity analyses. Currently hardcoded at `scripts/plot_zoo_results.py:100,113`. The pre-existing `_Z_THRESHOLD = 2.0` on line 45 is the same kind of constant and is also not in config. Both are in scope of the rule but neither was introduced by this PR — the `1.96` constant is new here. Author may defer config migration to a follow-up ticket.

nofollow: The legend adds a grey dashed "null 95% CI" entry with `color="0.5"`, while the per-window CI lines use `color=style["color"]`. This is a deliberate design choice (one legend entry for all windows) noted inline in the code. Acceptable.

---

### Scope (approved)

The PR changes exactly: `scripts/plot_zoo_results.py` (new `--null-ci` arg + helper functions), `tests/test_zoo_null_ci.py` (new tests), `zoo.mk` (4 explicit targets). PR description correctly notes "prose already landed in #743." No scope creep.

---

### Red team (approved)

- `null_df` passed to `_plot` without checking the required `z_threshold` column exists. However, `_compute_null_z_threshold` always adds `z_threshold` before `null_df` enters `_plot`, and `_load_null_df` returns None otherwise. The only path to `_plot(null_df=non-None)` goes through `_compute_null_z_threshold`. Safe.
- If `null_df` has rows for a window that isn't plotted (no data in `df`), those rows are silently ignored via `ci_sub.empty` check. Correct graceful handling.
- `null_df["window"].map(mu_w)` will produce NaN for any window present in null_df but absent in df. This would silently produce NaN `z_threshold` values, which matplotlib would skip in the plot without error. Acceptable — the `test_compute_null_z_threshold_adds_column` test asserts `not result["z_threshold"].isna().any()` but only for matching windows.

---

### Doc propagation (approved)

- `content/_includes/techrep/zscore.md` correctly states `|Z| >= 2` is not a significance threshold when non-stationary, cross-references the permutation test. ✓ (PR #743)
- `content/_includes/techrep/null-model.md` contains "exchangeability" and "comparability". The word "stationarity" is absent — the concept appears as "non-constant trend" in the proposition — so grep for "stationarity" (exit criterion 1) would fail. But this was in PR #743's scope, not #745's.

nofollow: Prose exit criteria 1 and 2 (null-model.md stationarity grep, zscore.md significance clarification) were supposed to land in #743. This PR (#745) explicitly covers only the code portion. The exit criteria mismatch is a prior-PR concern.

---

### Summary

**Verdict: approve.** The code is correct, tests pass, scope is clean, and math is consistent. Two `consider:` items for author discretion — both are pre-existing policy gaps touched (not introduced) by this PR.

PR #755, opened 2026-04-22
Ticket 0084: Z-score ribbon from equal-n subsampling variance

MinhHaDuong:
## Summary

- **New script** `compute_divergence_subsampled.py`: draws R=20 independent equal-n subsamples per (year, window), recomputes the statistic, produces `tab_subsample_{method}.csv`. All four lead methods: S2 (semantic), L1 (lexical), G9/G2 (citation — equal-n node subsampling).
- **New config keys**: `divergence.equal_n_r: 20`, `divergence.subsample_trim: 2`
- **New schema**: `DivergenceSubsampleSchema`; `DivergenceSummarySchema` gains four nullable ribbon columns: `z_trim_lo`, `z_trim_hi`, `z_median_subsample`, `n_subsamples`
- **`export_divergence_summary.py`**: optional `--subsample-csv` arg; derives ribbon from `null_mean`/`null_std`
- **`plot_companion_zseries.py`**: shaded ribbon (`alpha=0.2`); plots `z_median_subsample` so the line sits inside its ribbon; degrades gracefully to line-only
- **`divergence.mk`**: `tab_subsample_%.csv` targets for all four lead methods; `tab_summary_%.csv` recipes wire `--subsample-csv` for all four
- **Prose** `multilayer-detection.qmd` §4.5/§4.8 updated

## RNG independence

`_make_subsample_rng(seed, y, w, r)` seeds in `[+100000, +104511]` — disjoint from null subsample (`+199044–+202546`) and permutation (`+249044–+252546`). Proven by `test_subsample_rng_no_overlap_with_null_model_seeds`.

## Test plan

- [x] `test_subsampling_ribbon.py` — 12 tests: schema, smoke, bracket inequality, RNG independence
- [x] `test_bootstrap.py` — updated for expanded `DivergenceSummarySchema`
- [x] No new regressions (17 pre-existing failures unchanged — §5 prose stubs in STATE.md)
- [x] G9_community citation path smoke-tested: 63 rows (21 cells × R=3) produced cleanly
- [ ] Regenerate `tab_summary_*.csv` via `make divergence-summary` after corpus reruns (~2h GPU per lead method)

### DIFF
--- a/config/analysis.yaml
+++ b/config/analysis.yaml
@@ -32,6 +32,9 @@ divergence:
   pelt_model: rbf
   pelt_min_size: 2
 
+  equal_n_r: 20          # subsampling replicates for Z-score ribbon (ticket 0084)
+  subsample_trim: 2      # replicates to drop from each tail (10th/90th pctile for R=20)
+
   permutation:
     n_perm: 500
     z_threshold: 2.0    # consumed by compute_transition_zones.py (ticket 0055b)
--- a/content/multilayer-detection.qmd
+++ b/content/multilayer-detection.qmd
@@ -110,7 +110,7 @@ Each document is represented in three spaces that expose different aspects of th
 
 For each candidate year $t$ and each half-width $w$, we form a before-window $\mathcal{D}_t^- = \{d : t - w \le \text{year}(d) \le t\}$ and an after-window $\mathcal{D}_t^+ = \{d : t < \text{year}(d) \le t + w\}$. Windows that fail a minimum-size filter are skipped. In practice, $w = 3$ is the default and $w \in \{2, 4\}$ are reported as sensitivity checks. Year bounds are stored per window rather than recomputed from the year index, following the ticket 0067 convention that separates segmentation logic from timestamp arithmetic.
 
-Because the corpus grows over time, the after-window is typically larger than the before-window. All statistics are either sample-size invariant by construction or paired with an equal-$n$ subsampling robustness check (§4.8).
+Because the corpus grows over time, the after-window is typically larger than the before-window. All statistics are computed on equal-$n$ subsamples (§4.5), and the subsampling variance is shown explicitly as a ribbon in @fig-companion-zseries.
 
 ### 4.4 Distributional comparison
 
@@ -130,7 +130,7 @@ For each $(t, w)$ we compute four statistics across the two windows.
 
 The four statistics have different sources of variation, and we handle them with two inference primitives.
 
-*Permutation Z-scores for S2, L1, G9.* For each $(t, w)$, we pool the two windows, relabel them at random $B = 500$ times while preserving sample sizes, and recompute the statistic. The standardised effect size $Z(t) = (S_{\text{obs}} - \bar{S}_{\text{perm}}) / \mathrm{sd}(S_{\text{perm}})$ is directly comparable across years despite changing corpus size. A permutation scheme is appropriate here because the observed statistic has no closed-form null under general distributional alternatives.
+*Permutation Z-scores for S2, L1, G9.* For each $(t, w)$, we pool the two windows, relabel them at random $B = 500$ times while preserving sample sizes, and recompute the statistic. The standardised effect size $Z(t) = (S_{\text{obs}} - \bar{S}_{\text{perm}}) / \mathrm{sd}(S_{\text{perm}})$ is directly comparable across years despite changing corpus size. A permutation scheme is appropriate here because the observed statistic has no closed-form null under general distributional alternatives. To expose subsampling variance in the numerator, we draw $R = 20$ independent equal-$n$ subsamples of the observed side and compute a trimmed Z-score ribbon: the central sixteen values (discarding the two extremes on each end, equivalent to the 10th and 90th percentiles) bound the shaded region in @fig-companion-zseries. The denominator --- the permutation null mean and standard deviation --- is identical across all twenty draws; only the observed subsample changes.
 
 *Cross-validation fold variance for C2ST.* The C2ST AUC already has a built-in variance estimator: the standard deviation of AUC across the five CV folds. Permuting labels for C2ST is expensive (each permutation retrains the classifier $k$ times) and yields less information than the fold variance, which directly reflects sampling variability in the classifier's decision boundary. We therefore report AUC with its CV standard deviation rather than a permutation Z.
 
@@ -148,15 +148,13 @@ A validated zone could still reflect gradual redistribution rather than a genuin
 
 ### 4.8 Robustness
 
-Three robustness checks accompany every reported result.
-
-*Equal-$n$ subsampling.* The larger window is downsampled to the size of the smaller window, the statistic is recomputed over $R = 100$ subsampling repetitions, and the mean and empirical 95% interval are reported alongside the full-sample Z. This guards against the sample-size sensitivity that @szekely2013 warn about for energy distance on very unequal splits.
+Two robustness checks accompany every reported result.
 
 *Parameter sensitivity.* Window half-width $w$ is varied over $\{2, 3, 4\}$. We report the peak year and peak Z for each $w$; a zone's credibility rises when peaks agree across half-widths. $w = 3$ is the lead window: it is the smallest for which all three half-widths can be compared simultaneously, and year-to-year sampling noise is attenuated without smearing breakpoints beyond the annual COP cadence.
 
 *Bootstrap confidence intervals.* For each validated zone we draw $K$ bootstrap resamples of the underlying corpus, rerun the detection pipeline, and report the fraction of resamples that retain the zone. Bootstrap $K$ is read from the analysis configuration (ticket 0047).
 
-One caveat deserves explicit mention. The bias audit tickets B2 and B9 documented a residual growth bias in the permutation null at small window widths: when the corpus grows sharply across the window, the null distribution is inflated and Z-scores are biased upward in years of rapid publication growth. We flag the bias rather than correct it. A stratified permutation that conditions on year-specific sample sizes is the principled fix, and we defer it to future work. The equal-$n$ subsampling check partly compensates.
+One caveat deserves explicit mention. The bias audit tickets B2 and B9 documented a residual growth bias in the permutation null at small window widths: when the corpus grows sharply across the window, the null distribution is inflated and Z-scores are biased upward in years of rapid publication growth. We flag the bias rather than correct it. A stratified permutation that conditions on year-specific sample sizes is the principled fix, and we defer it to future work. The subsampling ribbon in @fig-companion-zseries exposes whether this bias is large relative to subsampling noise.
 
 ### 4.9 Interpretation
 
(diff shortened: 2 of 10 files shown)


MinhHaDuong:
verify: CI gap — no CI checks reported for this PR

MinhHaDuong (commented):
## Multi-perspective review (5 agents)

Risk: **standard + scripts + prose** → all five perspectives run. Label: `review:standard`.

All 31 tests in `test_subsampling_ribbon.py` and `test_bootstrap.py` pass.

---

### Correctness

**BLOCKER — `subsample_trim` config key is declared but never consumed (arch rule #6).**

`config/analysis.yaml` adds `subsample_trim: 2`, which signals intent to make the trim count config-driven. But `export_divergence_summary.py` hardcodes `trim = 2` (line 155) and never reads `load_analysis_config()`. The config key is a dead stub. Either wire it (read from config in `_aggregate_subsample_ribbon`) or remove the key. A dead config key that diverges from the hardcoded constant is a latent future bug.

**Minor — RNG independence test threshold is 5× weaker than the ticket exit criterion.**

The ticket specifies `|corr| < 0.1` on `R=20 × B=500`. The test (`test_subsample_and_permutation_rng_uncorrelated`) uses `|corr| < 0.5` on `R=20 × B=20`. The no-overlap seed proof (`test_subsample_rng_no_overlap_with_null_model_seeds`) provides structural guarantees, so the weaker threshold may be an intentional speed tradeoff. But the exit criterion in the ticket is not met as written.

consider: tighten to `|corr| < 0.3` (still fast with B=20) and update the ticket exit criterion to match, or add a comment explaining why < 0.5 is sufficient given the structural proof.

**Minor — `z_median_subsample` is median of the *trimmed* distribution, not the full 20 draws.**

`_aggregate_subsample_ribbon` computes `np.median(z_trimmed)` where `z_trimmed = z_sorted[2:-2]` (16 values). For symmetric distributions the two medians agree, but for skewed draws the trimmed median can differ from the full median. The ticket says "z_score stays as the median or mean of the 20 draws" — the current implementation delivers neither exactly. Not a bug, but worth documenting in the docstring.

consider: rename to `z_median_trimmed` to be precise, or note in the docstring that it is the median of the central 16 values.

**Minor — y-axis scaling in `plot_companion_zseries.py` ignores ribbon bounds.**

`max_abs_z` is computed from `sub["z_score"].abs().max()` (line 124). The plotted ribbon (`z_trim_hi`) could extend above this value, causing clipping at the axis boundary. In practice the ribbon is close to the median Z, so this is unlikely to matter visually, but it is a latent display bug.

consider: extend `max_abs_z` to also track `z_trim_hi.abs().max()` when ribbon columns are present.

---

### Consistency

**Minor — docstring in `_make_subsample_rng` states wrong seed bounds.**

The docstring says the range is `[seed+299042, seed+304567]`. Verified computation: actual range is `[seed+299002, seed+303511]` for `y ∈ [1990, 2025]`, `w ∈ {2,3,4}`, `r ∈ [0,19]`. The no-overlap property still holds (gap of ~46,000 above the permutation range), so this is documentation-only. The PR description also claims `[+100000, +104511]` which conflates the *offset from the window base* (`[+100000, +101007]`) with absolute seed values.

consider: correct the docstring to say `[seed+299002, seed+303511]`.

**Minor — test file docstring says "Three red tests" but there are four test classes.**

`tests/test_subsampling_ribbon.py` line 3: "Three red tests that define the expected behaviour … 1. Smoke 2. Bracket 3. RNG independence". The Schema class (`TestDivergenceSubsampleSchema`) is a fourth class not enumerated in the docstring.

consider: update the count to "Four test classes" or enumerate the schema class.

---

### Scope

No unrelated changes detected. The 10 files changed are all directly required by ticket 0084. The change that replaces the old summary recipe (covering all `BOOT_METHODS` individually) with a recipe covering only `SUBSAMP_METHODS` is correct today because the two lists are identical. However, the structure now silently breaks if a new BOOT_METHOD is added without a matching SUBSAMP entry: `SUMM_CSV` would reference a target with no recipe. 

consider: add a comment in `divergence.mk` noting that `SUMM_CSV` uses `BOOT_METHODS` for the phony target list but the recipe loop uses `SUBSAMP_METHODS`, and that these must stay in sync.

---

### Red team

**Minor — prose in §4.3 over-generalises: C2ST does not use equal-n subsampling.**

Old §4.3: "All statistics are either sample-size invariant by construction or paired with an equal-n subsampling robustness check (§4.8)."
New §4.3: "All statistics are computed on equal-n subsamples (§4.5)."

The new sentence is inaccurate: C2ST AUC uses cross-validation fold variance, not equal-n subsampling (as correctly explained in §4.5 itself). Replacing the careful hedge ("either … or") with a blanket claim introduces a factual error.

request-changes: restore the qualifier — e.g., "The three distance statistics (S2, L1, G9) are computed on equal-n subsamples (§4.5); C2ST AUC uses cross-validation fold variance." or similar.

**Minor — same RNG used sequentially for X and Y subsampling within one replicate.**

In `subsample_one_window`, `rng_r.choice(len(X_before), n)` is called first, then `rng_r.choice(len(Y_after), n)` on the same RNG state. The draws are statistically independent (sequential draws from a single stream are uncorrelated), but the X and Y selection within one replicate are generated from the same seeded stream. This is a deliberate engineering choice (one RNG per replicate) and is not a bug — merely worth noting in the docstring for future auditors.

nofollow: design choice, no action needed.

---

### Doc propagation

**Minor — "10th and 90th percentiles" claim in §4.5 is convention-dependent.**

§4.5 prose: "the central sixteen values (discarding the two extremes on each end, equivalent to the 10th and 90th percentiles)". With `trim=2` and `R=20`, `z_sorted[2]` corresponds to 2/20 = 10th percentile (fraction convention) and `z_sorted[17]` corresponds to 17/20 = 85th percentile (not 90th). The 90th percentile under the fraction convention would be `z_sorted[18]`. The claim is either approximate or dependent on a specific convention.

consider: say "approximately the 10th and 90th percentiles" or cite the specific convention (e.g., "the 3rd-lowest and 3rd-highest values").

---

### Summary

| Finding | Severity | Tag |
|---|---|---|
| `subsample_trim` config key unused (arch rule #6) | **blocker** | — |
| §4.3 prose: "all statistics" over-includes C2ST | **request-changes** | — |
| RNG independence test threshold weaker than ticket exit criterion | minor | `consider:` |
| `z_median_subsample` is median of trimmed, not full, distribution | minor | `consider:` |
| y-axis scaling ignores ribbon bounds | minor | `consider:` |
| Docstring `_make_subsample_rng` seed range wrong ([+299042] vs actual [+299002]) | minor | `consider:` |
| Test docstring says "Three red tests" but there are four classes | minor | `consider:` |
| `divergence.mk` BOOT_METHODS / SUBSAMP_METHODS sync risk | minor | `consider:` |
| §4.5 "10th and 90th percentiles" is convention-dependent | minor | `consider:` |
| Sequential RNG for X and Y in `subsample_one_window` | design note | `nofollow:` |

MinhHaDuong:
## Prose review — PR 755 (simulated peer panel)

**Panel:** domain expert (statistics/bibliometrics) · adversarial referee · copy editor · AI-tells auditor.
**Scope:** `content/multilayer-detection.qmd` — section rewrites in §4.3, §4.5, §4.8, §6.4.

---

### MAJORS (block acceptance)

**1. §4.3 overclaim — factually wrong**
Line 113: *"All statistics are computed on equal-$n$ subsamples (§4.5)"* is false. The point estimates S_obs in the Z-score formula are computed on the full unequal windows; the R=20 subsamples produce only the ribbon visualization. The old text was accurate: energy distance is sample-size invariant by construction; the permutation null handles the rest. A reviewer who reads §4.5 will immediately catch the contradiction. Fix: replace "are computed on" with "are accompanied by a subsampling variance ribbon (§4.5);".

**2. §4.5 percentile arithmetic error**
Line 133: *"equivalent to the 10th and 90th percentiles"* is incorrect. For R=20, dropping the 2 lowest and 2 highest values retains ranks 3–18. Using the standard interpolation formula (k−0.5)/n, the bounds are the 12.5th and 87.5th percentiles, not the 10th/90th. The config file sets `subsample_trim: 2`, confirming that 2 are dropped from each tail. Fix: change "10th and 90th" to "12th and 88th" (rounded), or "roughly the 12th–88th percentile range", or drop the percentile gloss entirely and simply say "the inner 16 of 20 draws".

**3. Figure cross-reference broken**
The new text references `@fig-companion-zseries` in §4.3 (line 113), §4.5 (line 133), and §4.8 (line 157). The figure is defined at line 200 with `{#fig-zseries}`, not `{#fig-zseries-companion}`. The reference will fail to resolve. The body text of §5.1 correctly uses `@fig-zseries`. Fix: replace all three occurrences of `@fig-companion-zseries` with `@fig-zseries`.

---

### MINORS

**4.** verifiable: Line 133 — "only the observed subsample changes" is grammatically odd; a new subsample is drawn each time, not the same subsample that "changes." Fix: "only the observed sample varies across draws."

**5.** verifiable: Line 133 — "the two extremes on each end" is redundant ("extremes" already implies the ends). Fix: "the two values at each end."

**6.** verifiable: Line 157 — "exposes whether this bias is large" implies a binary test; the ribbon is a quantitative display. Fix: "quantifies this bias relative to subsampling noise."

**7.** verifiable: Line 243 — "subsample-to-subsample noise" is informal and inconsistent with "subsampling noise" used in §4.8. Standardise on one term throughout.

---

### SUGGESTIONS

**8.** consider: The deletion of the §4.8 equal-$n$ subsampling paragraph removed the @szekely2013 citation warning about energy distance on very unequal splits. That citation motivated the entire ribbon procedure. Consider adding it as a parenthetical in §4.5: "...to expose sample-size sensitivity of the kind @szekely2013 warn about."

**9.** consider: The §4.5 sentence embedding the ribbon procedure inside the "Statistical inference" subsection blurs estimation and robustness visualization. A parenthetical label — "*(robustness display)*" — or a one-sentence preamble distinguishing the ribbon from the inferential Z would help readers track the logical layers.

**10.** consider: R=20 is a small ribbon sample. Some readers will notice that at years with fast corpus growth, 20 subsamples may not smooth the ribbon. An acknowledgement ("R=20 is chosen for computational speed; ribbons may appear coarser near peak-growth years") pre-empts that question.

---

### AI-TELLS AUDIT — CLEAN

No blacklisted words found. No blacklisted phrases found. Conditional word "robust" appears four times (lines 15, 125, 151, 254) — all in legitimate statistical context ("robustness check", "robustness reference"), meeting the `ok_when` criterion. "Landscape" does not appear. Sentence-initial connectors (Moreover, Furthermore, Additionally) absent. Contrast-farming ("not X but Y") count: 14 occurrences across the full paper — above the 3-per-document limit in `config/ai-tells.yml`. However, inspection shows these are genuine contrastive constructions in a methodology paper, not filler. Authors should spot-check and consolidate the least essential ones if the limit is meant as a hard rule.

---

### Panel verdict: **MAJOR REVISION**

Items 1–3 must be corrected before this PR merges. Items 4–7 are minor but straightforward to fix in the same pass. The AI-tells count on contrast constructions may need a policy decision (is the `max_per_document: 3` limit for the diff only, or the full document?).

PR #764, opened 2026-04-25
fix(0118): S4_frechet empty-results guard + smoke-mode min_papers precedence

MinhHaDuong:
## Summary

- Reorder `get_min_papers()` in `_divergence_io.py` so smoke mode (`n_works < 200`) fires **before** per-method overrides, preventing `S4_frechet` from using `min_papers=300` on the 90-row smoke fixture.
- Add `if not results: return empty_divergence_df()` guard before each `return pd.DataFrame(results)` in all four semantic compute functions (S1–S4), avoiding a zero-column DataFrame that fails `DivergenceSchema` validation.
- Add `empty_divergence_df` to the `from _divergence_io import (...)` block in `_divergence_semantic.py`.

## Test plan

- [x] RED: `pytest tests/test_divergence.py::TestSmokeSemantic::test_compute_method[S4_frechet]` failed before changes
- [x] GREEN: all 4 `TestSmokeSemantic::test_compute_method` tests pass after changes
- [x] `make check-fast` clean (967 passed, 13 skipped)

Closes ticket `tickets/0118-fix-s4-frechet-empty-results.erg`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #767, opened 2026-04-25
chore(0119): regenerate golden values (102-work smoke fixture)

MinhHaDuong:
## Summary

- Regenerates 10 stale golden CSV files in `tests/fixtures/smoke/golden/` that had 108 rows (old fixture) vs the current 96-row output
- Runs after ticket 0118 landed the S4_frechet smoke-mode guard, so all 15 methods produce valid output
- G9/C2ST golden files deliberately not created — tests auto-skip those methods

## Test plan

- [ ] `pytest tests/test_golden_values.py` → 15 passed, 3 skipped (G9_community, C2ST_embedding, C2ST_lexical)
- [ ] All 10 methods with golden files pass shape + value match (atol=1e-6)


MinhHaDuong:
## /verify-gate verdict: REROLL

**Primary criterion FAILS** — \`pytest tests/test_golden_values.py\` produces 13 pass / 2 fail / 3 skip, not the required 15 pass / 3 skip.

### Failing tests

**S3_sliced_wasserstein** — 96/96 values wrong, max 11.7% relative error. The golden CSV commits values from a different execution than what the current code produces. Example mismatch (row 0): committed 0.019575568800844 vs fresh run 0.019827542826533.

**S4_frechet** — 2/32 values differ by 6.4e-6 and 1.5e-6 (both above the atol=1e-6 threshold). Small FP difference, but still a failure.

### Other checks (pass)

- No G9/C2ST golden files accidentally created — OK
- \`data/catalogs/wp7_unified_wud1b8y1.csv\` not committed — OK
- The golden-regen commit (447103a) only touches `tests/fixtures/smoke/golden/tab_div_*.csv` — OK
- Note: branch also carries the 0118 code fix commits (4676b30, 5217876) which were merged as PR #764 — those are on main now, but they appear in the diff because the branch was cut before that merge. This is a pre-existing branch hygiene issue, not a blocker.

### Required fix

Regenerate the 10 golden CSVs again — this time confirming that `pytest tests/test_golden_values.py` passes *before* committing, i.e., run the tests from the same machine/env immediately after generating. S3 has stochastic subsampling (equal_n=true), so the golden must be regenerated and validated in one session.

MinhHaDuong:
**Verify-gate result: APPROVED (round 2)**

Re-ran full suite in the PR worktree: 15 passed, 3 skipped (G9/C2ST — expected). S3 and S4 golden files are byte-for-byte correct in the PR branch. Round 1 REROLL failures were a test-environment artifact (wrong working directory in the verify agent).

Both exit criteria confirmed:
- ✅ `pytest tests/test_golden_values.py` — 15 pass, 3 skip
- ✅ Golden CSVs committed (data-only change, no G9/C2ST files created)

PR #797, opened 2026-06-18
fix(0144): co-locate uv env with cache on /data (stop 1.8 GB per-worktree copy)

MinhHaDuong:
## What

Worktree creation copied the full torch closure (~1.8 GB, ~39 000 files) into every worktree's `.venv`, and recently started **failing outright** (EnterWorktree times out, leaving an orphan branch).

**Root cause:** the uv cache lives on `/data` (`/data/cache/uv`, a separate filesystem) while repos live on `/home/haduong`. uv hardlinks wheels from cache into each env *only* when both share a filesystem; across filesystems it copies the whole closure. The cross-fs copy is the slowness, and it now exceeds the worktree tool's timeout.

**Fix:** apply the cache's own hygiene principle (big regenerable things on `/data`) to the env too. `hooks/post-checkout` now pre-creates a shared env at `/data/envs/venv/oeconomia` (beside the cache → uv hardlinks) and symlinks the worktree's `.venv` at it. A dangling symlink makes `uv run` error, hence the pre-create. Guarded on `/data/envs` so machines without the data disk keep a local `.venv`. Filesystem-determined — no `.env` dependency, consistent with unifying secrets under `~/.config/keys`.

## Evidence (doudou)

- torch `.so` in the shared env is hardlinked to `/data/cache` (`nlink=2`, same device).
- A second worktree creates only a **25-byte symlink** instead of a 1.8 GB copy.
- 7 hook tests green, ruff green on the diff.

## Heads-up: pre-existing red CI on main

`make check-fast` currently fails on `origin/main` (unrelated to this PR): `scripts/compute_crossyear_zscore.py:67` uses `Optional[pd.DataFrame]` (legacy typing, caught by `test_no_legacy_typing`). This one-liner blocks **all** PRs' CI until fixed; it is not in this diff. Suggest a separate tiny fix.

**Ticket:** tickets/0144-uv-env-on-data-filesystem.erg

### DIFF
--- a/hooks/post-checkout
+++ b/hooks/post-checkout
@@ -3,9 +3,26 @@
 #
 # .env and .dvc/config.local are handled by .worktreeinclude (auto-copied
 # into worktrees created by EnterWorktree / `claude --worktree`).
-# This hook handles DVC data population only.
+#
+# This hook (a) points the worktree's uv environment at a shared env that lives
+# on the same filesystem as the uv cache, then (b) populates DVC-managed data.
 
-# Populate DVC-managed data from local cache.
+# (a) Co-locate the uv project environment with the uv cache on /data.
+# uv hardlinks wheels from its cache into the environment, but only when both
+# share a filesystem; across filesystems it falls back to copying the whole
+# torch closure (~1.8 GB) into every worktree, which is slow enough to exceed
+# the worktree tool's timeout and make worktree creation fail outright. The uv
+# cache already lives on /data, so we put the environment there too and point
+# this worktree's default `.venv` location at it via a symlink. A dangling
+# symlink makes `uv run` error, so the shared env is created first. Skipped
+# where /data is absent (the default local .venv is then used).
+SHARED_ENV=/data/envs/venv/oeconomia
+if [ -d /data/envs ] && [ ! -e .venv ]; then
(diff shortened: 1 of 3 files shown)


MinhHaDuong (commented):
**Review cycle 1 — independent (Sonnet, different model from coder).** Found two real edge cases in a hook that runs on every (incl. concurrent) worktree checkout:
- *Blocker:* dangling `.venv` symlink → old `[ ! -e .venv ]` guard proceeded and `ln -s` failed 'File exists'.
- *Blocker:* concurrent worktrees raced to `uv venv` the same shared path; `ln` unprotected.
Both fixed in c9d8f69: `ln -sfn` (idempotent, repairs stale link, leaves a real local .venv alone) + `flock` serialization on first-ever creation. Minor stderr/real-dir notes addressed by comments.

MinhHaDuong (commented):
**Review cycle 2 — verification.** Exit criteria met: fresh worktree `.venv` is a 25-byte symlink to /data/envs/venv/oeconomia (not a 1.8 GB copy); second worktree does no copy; torch .so hardlinked to /data/cache (nlink=2). Functionally tested all three .venv states (absent→linked, dangling→repaired, real-dir→untouched). 9 hook tests green; full `make check-fast` 998 passed on the rebased base; ruff + shellcheck clean on the new block. Ticket 0144 closed in d85b6b1. APPROVED.

PR #799, opened 2026-06-18
tickets: venv rotation (0145) + dvc-checkout worktree cost (0146)

MinhHaDuong:
Two follow-up tickets from PR #797 (uv env on /data).

- **0145** — rotate pre-existing real `.venv` directories (the ~2 GB main-checkout `.venv` on `/home`, stale worktrees) into symlinks to the shared `/data/envs/venv/oeconomia`. Same end state #797 gives fresh worktrees; reclaims `/home` space.
- **0146** — the venv fix unmasked the real remaining bottleneck: `hooks/post-checkout` runs `dvc checkout`, copying ~1.7 GB of DVC data into **every** worktree (DVC cache is same-fs but default link type copies on ext4). This is why EnterWorktree still times out. Make it opt-in and link-not-copy.

Filing only; no code change. `erg check` PASS.

Ticket: none

### DIFF
--- a/tickets/0145-rotate-existing-venv-to-shared-symlink.erg
+++ b/tickets/0145-rotate-existing-venv-to-shared-symlink.erg
@@ -1,0 +1,49 @@
+%erg 0.1
+Title: Rotate existing real .venv directories into symlinks to the shared /data env
+Created: 2026-06-18
+Author: HDMX-coding-agent
+
+--- log ---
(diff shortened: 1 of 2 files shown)


MinhHaDuong (commented):
Verified: PR adds only two well-formed ticket files (0145 venv rotation, 0146 dvc-checkout cost). `erg check` PASS. No code. 0146 is implemented by the follow-up PR. APPROVED.

PR #801, opened 2026-06-18
fix(0146): stop eager dvc checkout in worktree hook; add 'make data'

MinhHaDuong:
## What

Completes the worktree-creation fix. #797 made `.venv` a symlink; this removes the *other* per-worktree cost: `hooks/post-checkout` ran `dvc checkout` on every creation, copying ~1.7 GB of DVC data in. That is what still timed out EnterWorktree after #797.

- **Hook**: drop the `dvc checkout` block. Worktree creation now only symlinks the shared uv env — fast, data-free.
- **Makefile**: `make data` → `$(UV_RUN) dvc checkout`, to populate this worktree's data from the local cache on demand (no network). Added to `.PHONY` (a `data/` directory exists, which would otherwise shadow the target). `corpus-sync` still does the full remote pull.
- **Tests**: the hook must NOT run dvc checkout; `make data` must.

## Per-machine companion (not in this diff)

`.dvc/config.local` is gitignored. On doudou I set `cache.type = symlink` + `cache.protected = true` so `make data` links instead of copies. Deliberately **not** tracked: it is safe on doudou (read-only consumer) but unsafe on padme (the Phase-1 data producer, which rewrites DVC outs).

## Verification

- TDD red→green; `make check-fast`: **1001 passed**. ruff + shellcheck clean. `make -n data` parses. DVC `cache.type=symlink` confirmed on a scratch repo (checkout produces a symlink into the cache, file readable).

**Ticket:** tickets/0146-post-checkout-dvc-checkout-blocks-worktree-creation.erg

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,7 +207,7 @@ ZOO_FIGS := $(ZOO_SCHEMATICS) $(ZOO_RESULT_FIGS)
 ALL_FIGS := $(MANUSCRIPT_FIGS) $(DATAPAPER_FIGS) $(MULTILAYER_FIGS) $(TECHREP_FIGS) $(NCC_FIGS)
 
 # ── Default target ────────────────────────────────────────
-.PHONY: all setup manuscript papers figures figures-manuscript figures-datapaper figures-companion figures-techrep figures-ncc stats check check-fast smoke benchmark determinism-check regression regression-update check-corpus check-manuscript-data corpus corpus-sync corpus-discover corpus-enrich corpus-extend corpus-filter corpus-align corpus-filter-all corpus-tables corpus-validate deploy-corpus clean rebuild archive-analysis archive-manuscript archive-datapaper analysis-figures analysis-tables analysis-stats manuscript-render manuscript-figures datapaper-render datapaper-figures corpus-handoff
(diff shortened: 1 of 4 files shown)


MinhHaDuong (commented):
**Review cycle 1 — independent (Sonnet, different model).** APPROVE, no blockers. venv-symlink block intact; `data` correctly in `.PHONY` (a `data/` dir would otherwise shadow it); tests meaningful (hook must not contain dvc checkout; Makefile data: recipe must). Minor raised: `make data` uses bare `$(UV_RUN)` without `--group corpus`. Resolved — verified `uv run dvc --version` → 3.67.0 with `.venv` staying a symlink to the shared env; and this mirrors the existing `corpus-sync` target's pattern exactly, so it is consistent project behaviour, not a new risk.

MinhHaDuong (commented):
**Review cycle 2 — verification.** TDD red→green; `make check-fast` 1001 passed; ruff + shellcheck clean; `make -n data` parses; DVC cache.type=symlink confirmed on a scratch repo (checkout yields a readable symlink into the cache). Functionally: with the fixed hook, a worktree's post-checkout only symlinks the env — no 1.7 GB dvc copy. Ticket 0146 closed in b3fbf39. APPROVED.

PR #803, opened 2026-06-18
ticket(0157): standing regression test for worktree-creation speed

MinhHaDuong:
Roar follow-up from the #797/#801 worktree-speed fixes.

The post-checkout hook twice grew an eager heavy per-worktree step — venv copy (#797), dvc checkout (#801) — same defect class. The current `tests/test_post_checkout_hook.py` guards those two instances by source inspection; ticket 0157 asks for a behavioural integration test that times a throwaway worktree creation and asserts no GB-scale copy, catching any *new* heavy step in the class.

Filing only; no code. `erg validate` PASS.

Ticket: none

### DIFF
--- a/tickets/0157-worktree-creation-speed-regression-test.erg
+++ b/tickets/0157-worktree-creation-speed-regression-test.erg
@@ -1,0 +1,51 @@
+%erg 0.1
+Title: Standing regression test: worktree creation must not block on heavy eager work
+Created: 2026-06-18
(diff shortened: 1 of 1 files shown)


MinhHaDuong (commented):
Verified: adds only tickets/0157-*.erg (standing regression-test ticket from the #797/#801 roar sweep). No code. erg validate PASS. Mergeable/clean. APPROVED.

PR #804, opened 2026-06-18
feat(0145): rotate real .venv dirs into shared-env symlinks

MinhHaDuong:
## What

`scripts/rotate-venv-to-shared.sh` — replaces a real `.venv` directory with a symlink to the shared env on `/data`, reclaiming `/home` space. New worktrees already get a symlinked `.venv` from the post-checkout hook (#797); this rotates the pre-existing real ones (the main checkout's is ~2 GB).

- **Idempotent**: an existing symlink or absent `.venv` is left alone.
- **Safe**: never removes a real `.venv` unless the shared env exists and `lsof` shows no process holding the venv open.
- `SHARED_ENV` overridable (default `/data/envs/venv/oeconomia`), so the 5 integration tests are hermetic — no `/data` dependency.

`make check-fast` 1000 passed; the new tests (integration-marked) pass; ruff + shellcheck clean.

## Residual step (yours to run)

The one-time rotation of the **main checkout's** live 2 GB `.venv` is left for you to run:

```bash
bash scripts/rotate-venv-to-shared.sh "/home/haduong/CNRS/papiers/actif/Oeconomia - Climate finance"
```

I am sandboxed from running it (it deletes a real directory on a live checkout — correctly blocked). Run it when no session is using that venv; the `lsof` guard will skip it otherwise. It is idempotent, so re-running is harmless.

**Ticket:** tickets/0145-rotate-existing-venv-to-shared-symlink.erg

### DIFF
--- a/scripts/rotate-venv-to-shared.sh
+++ b/scripts/rotate-venv-to-shared.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+# Rotate a real .venv directory into a symlink to the shared env on /data.
+#
+# New worktrees already get a symlinked .venv from the post-checkout hook, but
+# checkouts that predate that fix still carry a full real .venv (the main
+# checkout's is ~2 GB on /home). This reclaims that space by replacing each real
+# .venv with a symlink to the shared env, the same end state the hook produces.
+#
+# Idempotent: an existing symlink or absent .venv is left alone. Safe: a real
+# .venv is never removed unless the shared env exists and no process holds the
+# venv open.
+#
+# Usage: rotate-venv-to-shared.sh [CHECKOUT_PATH...]   (default: current dir)
+# Override the shared env with SHARED_ENV=/path (used by tests).
+set -euo pipefail
+
+SHARED_ENV="${SHARED_ENV:-/data/envs/venv/oeconomia}"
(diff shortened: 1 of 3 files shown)


MinhHaDuong (commented):
**Review cycle 1 — independent (Sonnet, different model).** APPROVE, no blockers; flagged the in-use guard as untested. Acting on it exposed a REAL safety bug: `lsof +D` exits non-zero even when it lists open files, so the exit-code-based guard never detected a busy venv (would always rotate). Fixed in 4200bb1 to test lsof's output; added in-use + multi-target tests and an lsof-absent warning. Guard ordering (symlink→absent→not-dir→shared-env→lsof→rm) confirmed correct; rm cannot fire when shared env absent or path is a symlink.

MinhHaDuong (commented):
**Review cycle 2 — verification.** 7 integration tests green (incl. the now-passing in-use guard that caught the bug); shellcheck + ruff clean; make check-fast 1000 passed (integration tests excluded as designed). Idempotent and safe per the test matrix. Ticket 0145 closed. The one-time main-checkout rotation is documented as a user-run command (agent sandboxed from destructive ops on a live dir). APPROVED.

PR #807, opened 2026-06-18
fix(make): run pytest via `python -m`, add $(PYTHON) macro

MinhHaDuong:
Two commits.

## 1. Fix `make check` (`pytest: error: unrecognized arguments: -n`)

`pytest-xdist` is **not** missing — it is installed in the shared `/data` venv. The generated `bin/pytest` console script carried a **dangling shebang**:

```
#!/home/haduong/.claude/jobs/fddacf5c/tmp/wt-nohook/.venv/bin/python3   ← deleted worktree
```

`.venv` symlinks to the shared `/data/envs/venv/oeconomia` (one env across all worktrees, per the torch-hardlink constraint). A prior background-job worktree ran `uv sync` and uv stamped **all 52** console scripts with *that worktree's* local `.venv` path. When the worktree was removed the interpreter vanished, so `uv run pytest` fell through to the **system** pytest — which cannot import the venv's xdist.

`$(UV_RUN) python -m pytest` bypasses the generated console script, immune to stale-shebang drift no matter which worktree last synced. Matches what `check-fast` already did; the 5 remaining `$(UV_RUN) pytest` sites are aligned to it.

## 2. Add `$(PYTHON)` macro (DRY)

`$(UV_RUN) python` appeared 116 times across `Makefile` + the four included `.mk` files. Introduce `PYTHON ?= $(UV_RUN) python` and use `$(PYTHON)` everywhere. Semantically identical — expands to `uv run --env-file .env python` — but the "run Python through uv, never the console script" intent now lives in one named, overridable place. `UV_RUN` stays for the 11 `dvc` calls (no `python -m` entry point).

## Verification

- `make -n check` → `uv run --env-file .env python -m pytest tests/ -v --tb=short -n 4`.
- `$(PYTHON)` resolves to `uv run --env-file .env python`, including across the `.mk` includes (shared namespace).
- Test slices pass under `-n 4` (`test_script_hygiene` 60 passed, `test_io_discipline` 17 passed); full collection: 1227 tests.

## Out of scope (follow-up)

The 52 stale console scripts (`pytest`, `ruff`, `mypy`, `dvc`, …) still want regeneration via `uv sync` for interactive and `dvc` use — `dvc` has no `python -m` entry. Worth a ticket to pin shared-venv shebangs to the canonical `/data` interpreter so this can't recur.

Ticket: none

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,7 @@ content/*-vars.yml
 !content/manuscript-vars.yml
 
 # Python / uv
-.venv/
+.venv
 __pycache__/
 *.pyc
 .mypy_cache/
--- a/.venv
+++ b/.venv
@@ -1,0 +1,1 @@
+/data/envs/venv/oeconomia
--- a/.venv
+++ b/.venv
@@ -1,1 +1,0 @@
-/data/envs/venv/oeconomia
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,9 @@ export SOURCE_DATE_EPOCH := 0
 # ~/.bashrc isn't sourced — uv lives in ~/.local/bin by default.
 UV      ?= uv
 UV_RUN  ?= $(UV) run $(if $(wildcard .env),--env-file .env,)
+# Run Python through uv (via `python -m`, never the generated console scripts —
+# their shebangs point at the building worktree and break when it is removed).
+PYTHON  ?= $(UV_RUN) python
 export PATH := $(HOME)/.local/bin:$(PATH)
 
 # ── Modular Makefile includes ────────────────────────────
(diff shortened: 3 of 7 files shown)


MinhHaDuong (commented):
**Review 1 — Makefile fix.** Correct and matches the diagnosed root cause: console scripts in the shared env carried a dangling worktree shebang, so the durable fix is to invoke tools via the interpreter. PR adds `PYTHON ?= $(UV_RUN) python` and replaces `$(UV_RUN) python`→`$(PYTHON)` and `$(UV_RUN) pytest`→`$(PYTHON) -m pytest` across Makefile + divergence/multilayer-detection/venues/zoo .mk. Bypasses the generated scripts entirely. Sound.

MinhHaDuong (commented):
**Review 2 — .venv hygiene (added in 0f34b5f).** The PR had force-added `.venv` as a tracked symlink to /data/envs/venv/oeconomia (machine-local, would dangle elsewhere). Untracked it and fixed the root cause: ignore rule was `.venv/` (dir-only), which stopped matching once .venv became a symlink — changed to `.venv`. Verified: .venv now absent from the PR diff and `check-ignore .venv` passes. Fixes main's same latent gap on merge. APPROVED.

PR #811, opened 2026-06-18
fix(0158): pin shared-venv console-script shebangs to canonical interpreter

MinhHaDuong:
## What

Pin the shared uv env's generated console-script shebangs to the canonical,
symlink-resolved interpreter so they survive worktree removal.

**Ticket:** tickets/0158-pin-shared-venv-console-script-shebangs.erg

## Why

The project uses one shared uv env at `/data/envs/venv/oeconomia` (torch
hardlink constraint, tickets 0145/0157); every worktree's `.venv` is a symlink
to it. On `uv sync`, uv stamps each `bin/` console script (`pytest`, `mypy`,
`dvc`, …) with the **unresolved** path `<worktree>/.venv/bin/python3`. When that
worktree is removed the interpreter dangles and the tool silently falls through
to the system one (PR #807 root cause: `pytest: error: unrecognized arguments: -n`).

PR #807 routed the test path around this via `python -m pytest`, but `dvc` has
no `python -m` entry point, and interactive `uv run ruff/mypy/dvc` still hits
the stale shebang.

## How — mechanism (option c + a, both machine-agnostic via `realpath`)

- **`export UV_PROJECT_ENVIRONMENT := $(realpath .venv)`** — a Make-driven
  `uv sync` writes the canonical resolved interpreter, not the worktree symlink
  path. Empirically confirmed: plain sync writes `<wt>/.venv/bin/python3`; with
  this set, it writes `/data/envs/venv/oeconomia/bin/python3`.
- **`venv-canonicalize`** target — idempotent `sed` sweep repairing `bin/`
  python shebangs uv left stale (uv only rewrites on an *actual* sync, so
  scripts a removed worktree left behind keep its dangling interpreter). Wired
  as an order-only prereq of `check`/`check-fast`.

`realpath` resolves the symlink to `/data/...` on doudou and is a no-op where
`.venv` is a real local dir (padme, CI). Empty `.venv` → unexported. No
machine-local path enters tracked code (computed at build time).

## Test

`tests/test_venv_shebang_canonical.py` (integration) reads each `bin/` shebang
and asserts the interpreter lives under the canonical env, never a removable
worktree path. Asserts on the **literal** shebang — a worktree-local shebang
resolves *through* the symlink back to canonical, so only the literal exposes
drift. Marked integration because the shared env is mutable state across
parallel sessions (belongs in the pre-PR gate, not the rapid check-fast loop).

## Verification

- Red against the live env (`bin/pytest` pointed at a removed job worktree) → green after fix.
- `make venv-canonicalize` repairs an injected worktree-local shebang; idempotent (mtimes stable when already canonical).
- `make check-fast`: 1001 passed, 14 skipped.
- `/verify-adherence`: **PASS** — no machine-local paths in tracked code, ruff clean, adherence suite 3 passed.


MinhHaDuong (commented):
## /review-pr — five-perspective panel (correctness, consistency, scope, red team, doc propagation)

**Verdict: approve with two minor cleanups.** No blockers. `make check-fast` → 1001 passed, 14 skipped; the new `test_console_script_shebangs_are_canonical` passes against the live shared env. Both mechanisms verified working: `UV_PROJECT_ENVIRONMENT` confirmed honored by uv 0.7.8, and the `venv-canonicalize` guards/grep/sed compose correctly (dangling `.venv`, env-style and arg-bearing shebangs all handled).

### Minor findings (non-blocking)

- `verifiable:` **Duplicate `.PHONY: venv-canonicalize`.** The target is already in the consolidated `.PHONY` list (Makefile:224) and is declared again locally at Makefile:721. Harmless (Make dedupes) but inconsistent with the file's single-consolidated-list pattern — drop the local one. — *consistency*
- `verifiable:` **Typo, Makefile:717:** "so scripts a removed worktree left behind" → "so scripts that a removed worktree left behind". — *consistency*

### Considerations (author may dismiss)

- `consider:` **`venv-canonicalize` runs on every `make check`/`check-fast`**, mutating the shared `/data` env's `bin/` as an order-only prereq of read-feeling commands. Safe in practice: the `grep -qxF` idempotency guard means `sed -i` only fires on actual drift (zero rewrites observed here, all scripts already canonical), and `sed -i` replaces the inode atomically via `rename(2)` — already-`execve`'d scripts and concurrent identical rewrites cannot corrupt. Worth knowing the side effect exists. — *red team*
- `consider:` **`UV_PROJECT_ENVIRONMENT := $(realpath .venv)` expands at parse time.** On first-time setup (no `.venv` yet) it stays unexported for that invocation even if a prereq creates `.venv` mid-run; a re-run picks it up. The comment notes the empty case but not the created-mid-run edge. — *consistency* — As documented; no change required.
- `consider:` Out-of-repo memory `project_worktree_env_data.md` still frames ticket 0158 as open/pending and doesn't mention the implemented `UV_PROJECT_ENVIRONMENT` / `venv-canonicalize` fix. Flag for a `/memory` sweep at session close — not part of this PR. — *doc propagation*

### Scope / lifecycle
Doing both ticket options (c) preventive + (a) curative is the minimal correct set, not gold-plating — (c) fixes future syncs, (a) heals shebangs already stamped by a removed worktree. Three files map 1:1 to the deliverables; no unrelated changes. PR body carries the `**Ticket:**` line, so `erg-pr-merge` will auto-close 0158 on merge.

🤖 review-pr panel · sonnet reviewers

PR #946, opened 2026-07-09
ticket(0168): guard enrich_embeddings model.half() with CUDA check

MinhHaDuong:
## Summary
`scripts/enrich_embeddings.py` called `model.half()` unconditionally. fp16 has no fast CPU kernels, so on a CPU-only host this forced a *slower* fp32→fp16 path, not a faster one. This wraps the call in `if torch.cuda.is_available():`, matching the guard the companion `scripts/het_embed.py` already uses.

## Behavior
- **GPU hosts (padme):** unchanged — CUDA present → same fp16 path, identical embedding output.
- **CPU-only hosts:** stay in fp32, no forced slow fp16.

## Verification
- `make check-fast` passes (ruff adherence + unit suite, exit 0).
- `/verify-adherence` → PASS (no mechanical failures, no semantic findings). Per-module tests `test_enrich_embeddings.py` + `test_embeddings_cache.py`: 23 passed.

**Ticket:** tickets/0168-guard-enrich-embeddings-py-model-half-wi.erg

### DIFF
--- a/scripts/enrich_embeddings.py
+++ b/scripts/enrich_embeddings.py
@@ -204,8 +204,10 @@ def main():
         torch.set_num_threads(n_cpu)
         log.info("Loading %s (%d threads)...", MODEL_NAME, n_cpu)
         model = SentenceTransformer(MODEL_NAME)
-        # FP16 halves VRAM (~6.4 GB instead of ~12.8 GB), fits 16 GB GPUs
-        model.half()
+        # FP16 halves VRAM (~6.4 GB instead of ~12.8 GB), fits 16 GB GPUs.
+        # CPU has no fast fp16 kernels, so fp32 there (guard matches het_embed.py).
+        if torch.cuda.is_available():
+            model.half()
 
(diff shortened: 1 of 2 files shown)


MinhHaDuong (commented):
## /review-pr — Correctness (trivial-tier, single perspective)

A 4-line CUDA guard; proportional depth → Correctness only.

**Verified correct and in scope:**
- `torch` is imported (enrich_embeddings.py:159) and already used one line above the guard (`torch.set_num_threads` at :204), so `torch.cuda.is_available()` at :209 cannot NameError.
- Guard matches the companion reference exactly: het_embed.py:53–54 uses the same `if torch.cuda.is_available(): model.half()` pattern. Comment cross-references it.
- GPU path unchanged (CUDA present → fp16); CPU-only hosts stay fp32. Matches ticket 0168 invariant and exit criteria.
- `make check-fast`: 1106 passed, 21 skipped. Remaining warnings (PLR0912/C901/module-length) are pre-existing smells in untouched scripts, not introduced here.

No findings. Approve.

🤖 review-pr

PR #978, opened 2026-07-10
ticket(0216): fast-path tier guards — heavy-import auto-mark + duration ratchet

MinhHaDuong:
## What

Two enforcement layers keep heavy tests off the fast inner loop (`make check-fast` = `-m "not slow and not integration and not adherence"`), closing the "whack-a-mole forever" gap left open by ticket 0214:

1. **Auto-mark (the real fix).** A `pytest_collection_modifyitems` hook in `tests/conftest.py` marks every test whose module statically imports a heavy dependency (dcor / torch / ot / sentence_transformers / matplotlib) `slow` — unless it already carries slow/integration/adherence. Detection is a static source scan (`tests/_tier_autoscan.py`), mirroring the existing `test_no_slow_in_subprocess_files` convention; it catches lazy imports inside function bodies and never imports the heavy module to classify it.

2. **Ratchet (backstop for heavy compute without heavy imports).** An `adherence` meta-test (`tests/test_fast_path_budget.py`) reads recorded per-test durations and fails if any fast-path test exceeds the budget in `[tool.fast_path_ratchet]` (pyproject, **5.0s**), naming the offender and the tier it should join. Cold-start **skips** (never errors) when no durations file exists.

## Why the auto-mark, not per-test marking

`import dcor` costs ~7s (numba JIT at import, no disk cache) and is paid **once per xdist worker**, attaching to whichever dcor test runs first on that worker. So marking individual dcor tests `slow` is whack-a-mole: the 7s tax jumps to the next dcor test, nondeterministically per run — a pure per-test ratchet would flag a different test each run and never converge. A collection-time auto-mark on heavy-import **modules** removes all such cost in one deterministic stroke (all 7 dcor/torch/matplotlib-importing test files leave the fast path at once).

## Durations recorded serially, on purpose

`report.duration` is wall time; under `-n N` xdist oversubscription it inflates 3-5x from CPU contention (measured: a 1.9s test showed 8.9s under `-n4`), which would make a per-test budget fire on scheduling noise. So recording is **serial and opt-in** via `make test-durations` (`RECORD_TEST_DURATIONS=1`, `-p no:xdist`), giving budget >> noise. The `.test_durations.json` file is gitignored (machine/run-specific); a fresh checkout cold-starts the ratchet into skip.

## Re-tiering surfaced by the ratchet

The ratchet flagged 6 genuinely-heavy fast-path tests whose cost is heavy compute / **transitive** import the static scanner can't see (classifier + embedding + matplotlib pulled via the script modules they exercise). Re-tiered by cost:

- `test_multilayer_detection_figures.py` → module-level `integration` (all 8 tests drive a figure script via subprocess — a prior marker-discipline miss).
- `test_c2st.py` → `slow` on the two heavy C2ST schema classes (class-level, since the first to run pays the import — same shape as dcor); light `TestC2STCore` stays fast.
- `test_null_model_c2st.py` → `slow` on the two C2ST NullModel classes; file joins `TestMarkerDiscipline.EXCEPTIONS` (it also has an `integration` subprocess smoke class — slow + subprocess coexist by design).

## Before / after fast-path timing (`-n 4`, doudou)

| | Wall time | Max single fast-path test |
|---|---|---|
| before | ~84s | 5.5s (+ nondeterministic 7s dcor tax) |
| after | **~16s** | **1.6s** |

Serial fast-path max dropped 7.86s → **1.30s** (budget 5.0s → ~3.8x headroom), so the ratchet ships green and only trips on a genuine new regression. No coverage lost — `make check` still runs every tier.

## Notes

- Design diverges from the ticket's original ratchet-only framing: the **auto-mark is the refinement** (the ticket's own body predicted per-test marking would be whack-a-mole and pointed here). Both layers are implemented.
- Residual heavy tests left on the fast path: **none** over budget. The pre-existing `test_bib_doi_title.py` collection error (`bibtexparser`, an optional corpus-env dep) is unrelated environment drift; `make test-durations` uses `--continue-on-collection-errors` to stay best-effort under it.

TDD: pure-logic unit tests (heavy-import detection; ratchet selection incl. synthetic 10s red/green fixtures) were red on the missing module, green after implementation.

**Ticket:** tickets/0216-duration-ratchet-guard-fast-path-budget.erg

### DIFF
--- a/.claude/rules/coding.md
+++ b/.claude/rules/coding.md
@@ -40,6 +40,11 @@ Markers gate which `make` target runs a test. Pick the tier by **cost**, not by 
 
 `make check-fast` = `-m "not slow and not integration and not adherence"` (the inner loop — must stay pure logic). `make lint` = `-m adherence`. `make check` runs everything. No coverage is lost by moving a test to a slower tier — `make check` still runs it before every PR.
 
+Two guards keep the fast tier honest (ticket 0216, owned by `tests/test_fast_path_budget.py` + `tests/conftest.py`):
+
+- **Auto-mark** — a collection hook marks any test whose module imports a heavy dependency (dcor/torch/ot/sentence_transformers/matplotlib) `slow` automatically, so you rarely mark those by hand. `import dcor` alone costs ~7s (numba JIT), so this is the deterministic fix for the per-worker import tax.
+- **Ratchet** — an `adherence` test fails if any fast-path test on the last recorded run exceeds the budget in `[tool.fast_path_ratchet]` (pyproject). Heavy *compute* that no heavy *import* reveals lands here: give it `slow` (heavy compute / real data) or `integration` (subprocess). Refresh timings with `make test-durations` (serial, so the budget measures true cost, not xdist contention).
+
 ## Testing
 
 - Acceptance tests (e.g., `make corpus-validate`) are the top-level contract — never weaken without discussion.
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,9 @@ __pycache__/
 .mypy_cache/
 .ruff_cache/
 .pytest_cache/
+# Per-test durations recorded by the fast-path ratchet (ticket 0216) — a
+# machine/run-specific artifact regenerated on every `make check`.
+.test_durations.json
 .cache/
 
 # Replication archive
--- a/Makefile
+++ b/Makefile
@@ -228,7 +228,7 @@ ZOO_FIGS := $(ZOO_SCHEMATICS) $(ZOO_RESULT_FIGS)
 ALL_FIGS := $(MANUSCRIPT_FIGS) $(DATAPAPER_FIGS) $(MULTILAYER_FIGS) $(TECHREP_FIGS) $(NCC_FIGS)
 
 # ── Default target ────────────────────────────────────────
-.PHONY: all setup manuscript papers figures figures-manuscript figures-datapaper figures-companion figures-techrep figures-ncc stats check check-fast lint venv-canonicalize smoke benchmark determinism-check regression regression-update audit-pdf-content check-corpus check-manuscript-data data corpus corpus-sync corpus-discover corpus-enrich corpus-extend corpus-filter corpus-align corpus-filter-all corpus-tables corpus-validate deploy-corpus clean rebuild archive-analysis archive-manuscript archive-datapaper analysis-figures analysis-tables analysis-stats manuscript-render manuscript-figures datapaper-render datapaper-figures corpus-handoff
+.PHONY: all setup manuscript papers figures figures-manuscript figures-datapaper figures-companion figures-techrep figures-ncc stats check check-fast lint test-durations venv-canonicalize smoke benchmark determinism-check regression regression-update audit-pdf-content check-corpus check-manuscript-data data corpus corpus-sync corpus-discover corpus-enrich corpus-extend corpus-filter corpus-align corpus-filter-all corpus-tables corpus-validate deploy-corpus clean rebuild archive-analysis archive-manuscript archive-datapaper analysis-figures analysis-tables analysis-stats manuscript-render manuscript-figures datapaper-render datapaper-figures corpus-handoff
 
 .DEFAULT_GOAL := manuscript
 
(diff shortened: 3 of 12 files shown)


MinhHaDuong (commented):
## Review — ticket 0216 fast-path tier guards

Verified empirically (not just read): all three ticket-0216 exit-criteria behaviors reproduce correctly on this branch.

- `tests/test_fast_path_budget.py::test_fast_path_within_budget` fails on a synthetic `.test_durations.json` record with a 10s unmarked test, passes once the same record carries `markers: ["slow"]`, and `pytest.skip`s (not errors) with `.test_durations.json` absent.
- `make test-durations` (serial, `RECORD_TEST_DURATIONS=1 -p no:xdist`) writes the file; the same command under `-n 2` correctly does **not** write it (xdist-contention guard via `getoption("numprocesses")` works).
- Static-scan false positives: scanned every `tests/*.py` for the heavy-import regex hits and manually checked each — no false positive from a comment or a string literal (e.g. `tests/test_split_embeddings.py:71,73` contain `"import matplotlib"` / `"sentence_transformers"` as string literals inside `assert`, correctly *not* matched because the `^\s*` anchor requires the line to start with `from`/`import`).
- Full `tests/` run at `-m "not slow and not integration and not adherence"`: 925 passed / 10 skipped / 444 deselected / 1 pre-existing failure (`test_robustness_observability.py::test_step1_counter_attempted`, missing `data/catalogs/unified_works.csv` — unrelated to this diff, a data file absent from this worktree). `TestMarkerDiscipline` (subprocess-vs-slow hygiene) passes with the two new `EXCEPTIONS` entries.

### consider: module-granularity of the auto-mark sweeps far more previously-fast tests than the PR body's "6 flagged" framing suggests

`tests/_tier_autoscan.py::file_has_heavy_import` scans a whole module's source; `conftest.py::pytest_collection_modifyitems` then marks **every** test in that module `slow` if the heavy import appears anywhere in it — including inside a function-local, exception-guarded block. This is called out in the PR body as intentional ("all 7 dcor/torch/matplotlib-importing test files leave the fast path at once"), and I confirmed 7 files match. But the per-file blast radius is uneven and, in one case, large:

- `tests/test_null_model.py:983,1026` — `import torch` appears only inside two `TestBackendComparison` methods (`test_gpu_energy_formula_correct`, `test_gpu_mmd_formula_correct`), each wrapped in `try/except ImportError` + an immediate `pytest.skip("CUDA not available")` if no GPU — i.e., these two tests already cost near-zero on a non-GPU dev box. Because of module-granularity, this pulls **all 37 tests in the file** (34 of which never touch torch) off `make check-fast`. Reproduced: `pytest tests/test_null_model.py --collect-only -m "not slow and not integration and not adherence"` → `no tests collected (37 deselected)`, vs `--collect-only` (no filter) → `37 tests collected`.
- Smaller instances of the same pattern: `tests/test_embedding_sensitivity.py` (11 tests deselected; only `TestPCABreakPreservation` at line 106 — already explicitly `@pytest.mark.slow` — actually imports `dcor`, at line 130; the unrelated `TestHelpers` class, 5 pure-logic tests, is swept along), `tests/test_analytical_null.py` (7 deselected; `import matplotlib` is local to one plotting test), `tests/test_zoo_figure_polish.py` (5 deselected; same pattern).

None of this breaks the ticket's stated exit criteria (`make check` still runs everything, so there's no coverage loss — the PR body's claim holds), but it does mean the fast inner loop silently lost dozens of genuinely-fast, torch/dcor-free tests as a side effect of two CUDA-skip-guarded tests sharing a file with them. Worth a follow-up: hoist `test_gpu_energy_formula_correct` / `test_gpu_mmd_formula_correct` (and the `TestPCABreakPreservation`-adjacent `TestHelpers`) into files that don't share a module with heavy-import tests, to win back ~40 tests' worth of fast-loop coverage. Not a blocker for this PR — the mechanism works as documented and the trade-off is stated in the PR body, just under-quantified.

### nofollow: ratchet threshold (5.0s) vs. ticket's suggested "~3.0s"

Ticket 0216 body says "BUDGET_S=3.0" / "generous (~3s)"; the shipped `pyproject.toml` `[tool.fast_path_ratchet]` sets `budget_seconds = 5.0`. The PR body justifies this with measured data (serial fast-path max dropped to 1.30s, "budget 5.0s → ~3.8x headroom"), which is a stronger basis than the ticket's a-priori estimate. Not pursuing — the exit criteria only require "threshold lives in config," which it does, and the value is defensible.

### nofollow: static-scan regex has a theoretical semicolon-chained-import gap

`_HEAVY_RE` anchors on `^\s*(?:from|import)\s+`, so `import os; import torch` on one line would only match `os` (not heavy) and miss `torch`. Not pursuing: semicolon-chained imports don't appear anywhere in `tests/` or `scripts/` (grep confirms), and ruff's style rules would flag the construct anyway if introduced.

### Correctness spot-checks that passed

- `is_fast_path` / `fast_path_violations` (`tests/_tier_autoscan.py:67-85`) unit-tested for all three non-fast markers (slow/integration/adherence) individually — each correctly exempts an over-budget record.
- `pytest_runtest_logreport` (`tests/conftest.py`) reads `report.keywords`, which reflects marks *after* collection-time auto-marking, so the ratchet sees the effective tier, not the source-declared one — confirmed by design read, consistent with the auto-mark-then-ratchet layering described in the module docstring.
- `pytest_sessionfinish`'s xdist-worker guard (`hasattr(session.config, "workerinput")`) and distributed-controller guard (`getoption("numprocesses", None)`) both verified live (see above).
- Cold-start and doc-propagation exit criteria both satisfied: `.claude/rules/coding.md` documents both guards in the marker-tier table with the "no coverage lost" invariant restated.

No blockers found. The two enforcement layers behave as specified, the TDD red/green fixtures for the ratchet are real (not just described), and the auto-mark/ratchet split correctly divides "heavy import" from "heavy compute" as designed.

PR #984, opened 2026-07-10
test(0228): pin fast-path auto-mark hook ordering + document over-sweep granularity decision

MinhHaDuong:
**Ticket:** tickets/0228-harden-fast-path-auto-mark-pin-hook-orde.erg

Follow-up from 0216 (#978 /gaze review, tracker 0213). Hardens the collection-time heavy-import auto-mark that keeps the ~7s dcor/torch import tax off `make check-fast`.

## Action 1 — hook ordering pinned
The auto-mark depends on pytest firing user `pytest_collection_modifyitems` **before** its internal `-m` deselection. Verified under pytest 9.0.2 but previously unpinned — an ordering inversion would silently readmit heavy-import modules to the fast path with no failure elsewhere (the auto-mark fails open; it only *adds* work).

New `tests/test_fast_path_hook_ordering.py` (integration) spins a real collection in a throwaway dir whose conftest mirrors the production hook (same `file_has_heavy_import`, same `add_marker(slow)`), and asserts a synthetic heavy-import test is **deselected** on the fast-path selector and **selected** under `-m slow`. `--collect-only` keeps the lazy import from firing. Teeth confirmed by mutation (disabling `add_marker` fails the deselection assertion).

## Action 2 — over-sweep quantified, module granularity kept
Measured: 95 test functions (~120 collected items) across 7 heavy-import modules are auto-marked with no explicit marker of their own. But per-test refinement recovers almost nothing:
- **~104 items** live in heavy-*compute* modules (`test_divergence`, `test_null_model`, `test_embedding_sensitivity`) — each runs tens of seconds even with the heavy import excluded, so the layer-2 duration ratchet excludes them anyway.
- **`test_pipeline_io`** imports matplotlib at top level — module import already pays the tax; per-test marking can't help.
- Only **~12 light tests** in 2 modules are genuinely recoverable, and each still retains one matplotlib-touching test.

**Decision: keep module granularity.** Sound per-test lazy-import detection (through fixtures/helpers/transitive calls) is a fragile reimplementation of pytest whose false negatives would readmit the exact import tax the guard removes. No coverage is lost (over-swept tests still run in `make check`); one can run an over-swept test in the inner loop directly (`pytest tests/test_foo.py::test_bar`, no `-m`). Documented in `file_has_heavy_import` docstring.

## Verification
- `tests/test_fast_path_hook_ordering.py`: 2 passed; mutation-proven teeth.
- `tests/test_fast_path_budget.py` + new test: 19 passed, 1 skipped.
- `/verify-adherence`: PASS (117 adherence passed). ruff clean.
- Pre-existing unrelated env gap: `bibtexparser` absent → `tests/test_bib_doi_title.py` collection error (untouched module, not a regression).


MinhHaDuong (commented):
## /review-pr — approve (test-infrastructure, verified locally)

Scope is test files + ticket close only (`tests/_tier_autoscan.py` docstring, new `tests/test_fast_path_hook_ordering.py`, ticket `Closed: done`). No production code. Reviewed for correctness, consistency, scope, and red-team; findings below.

**Correctness (high confidence).** Verified on this branch:
- `tests/test_fast_path_hook_ordering.py`: 2 passed (1.23s).
- Mutation teeth reproduced: neutralizing `item.add_marker(pytest.mark.slow)` in the synthetic conftest fails `test_auto_mark_deselects_on_fast_path` (`assert 'test_heavy_module' not in {...}`). The claim holds.
- `tests/test_fast_path_budget.py` + new file: 17 passed, 1 skipped. `ruff check` clean on both touched files.

**Consistency (high confidence).** The synthetic conftest single-sources detection via the real `file_has_heavy_import`, and the static-scan approach matches the existing `test_script_hygiene.py` substring convention. `@pytest.mark.integration` tier is correct (spawns `subprocess`). Good.

**Scope (high confidence).** Changes only what the ticket asks; ticket closed in-PR per ticket discipline.

Minor:
- `consider:` `_collect()` asserts `returncode == 0`, but `pytest --collect-only -m <selector>` returns exit code 5 when a selector matches zero items. Both current selectors always collect ≥1 item (`test_light_module` on fast path, `test_heavy_module` under `-m slow`), so this is safe today — but a future selector that legitimately collects nothing would surface as `collection failed` rather than a clean empty set. No test attached; author may dismiss.
- `nofollow:` The quantified over-sweep numbers (95 functions / ~104 heavy-compute items / ~12 recoverable) are a point-in-time measurement dated 2026-07-10 in both docstring and ticket. They will drift as the suite grows, but they are a decision record with no CI enforcement — no staleness risk. Recorded, no action.

Verdict: **approve.** Note branch is 2 commits behind `origin/main` (both unrelated: tickets 0220/0269 harness migration) — rebase before merge per gate discipline.

PR #992, opened 2026-07-10
fix(0233): DERIVED producers makedirs before validate_io + class guard

MinhHaDuong:
Finishes the `$(DERIVED)`-producer makedirs class that ticket 0218 started, and adds a standing guard so the class stays closed.

## Problem
`validate_io()` only *checks* the output dir exists (it raises if absent) — it does not create it. `content/tables/` is git-tracked and always present, but `$(DERIVED)`=`data/derived/tables` is gitignored and regenerable, so it is absent on a clean tree. A producer that writes there without `os.makedirs` fails under `make -jN` / an isolated target build depending on which producer runs first (0218 fixed four; the roar sweep that filed this ticket found more).

## Change
- **8 producers** get `os.makedirs(os.path.dirname(io_args.output) or ".", exist_ok=True)` before `validate_io`:
  - the 3 named by the ticket: `analyze_embeddings`, `analyze_cocitation`, `build_het_core`
  - `compute_lexical` — a 4th the guard's auto-discovery surfaced that the manual sweep missed
  - `compute_analytical_null`, `compute_crossyear_zscore`, `compute_sensitivity_grid`, `compute_venue_concentration` — 4 more brought in by rebasing onto 0231's eviction work; the guard caught them going RED
- **Standing guard** `TestDerivedProducersMakedirs` (adherence tier): Makefile-driven auto-discovery of every target whose output resolves under `data/derived`, asserting the producing script calls `os.makedirs(os.path.dirname(io_args.output) ...)` **in the main() prologue** (between `parse_io_args()` and `validate_io`). Not a hardcoded list — a future `$(DERIVED)` producer that forgets the bootstrap is caught automatically. The prologue+output-target anchoring means a decoy `makedirs` elsewhere (e.g. `analyze_genealogy`'s module-level `makedirs(TABLES_DIR)`) can't mask a deleted real fix. `test_discovery_nonempty` pins all 12 known producers so a scan regression that drops one fails loudly. Green now; was RED for every offender, and mutation-verified: deleting any producer's real fix turns it RED.
- `simplify` cleanup in `compute_lexical.py`: dropped a now-redundant post-`validate_io` `makedirs` and its dead `TABLES_DIR` / `BASE_DIR` import.

## Verification
- `TestDerivedProducersMakedirs`: 13 passed, 3 skipped (skips write via `save_csv`, not `validate_io`).
- Adherence tier: 132 passed. Ruff: clean on all touched files.
- Teeth proven: reverting any producer's makedirs turns the guard RED.
- `validate_io`'s raise-on-missing-dir behaviour is unchanged (still pinned by `test_validate_io_checks_output_dir`). No Phase-1 trigger; DVC untouched.
- Out of scope / pre-existing (not caused by this diff): 3 `qa_*` subprocess tests fail in this worktree on a missing `corpus`-extra dep (`bibtexparser`); they pass in the corpus env.

**Ticket:** tickets/0233-derived-producers-makedirs-bootstrap.erg

### DIFF
--- a/scripts/analyze_cocitation.py
+++ b/scripts/analyze_cocitation.py
@@ -210,6 +210,9 @@ def run_robustness(G):
 
 def main():
     io_args, extra = parse_io_args()
+    # Output lands under data/derived/tables/ (gitignored, regenerable — ticket 0233);
+    # create it so validate_io's dir check passes on a clean tree.
+    os.makedirs(os.path.dirname(io_args.output) or ".", exist_ok=True)
     validate_io(output=io_args.output)
 
     parser = argparse.ArgumentParser()
--- a/scripts/analyze_embeddings.py
+++ b/scripts/analyze_embeddings.py
@@ -40,6 +40,9 @@ CLUSTERS_PATH = os.path.join(DERIVED_TABLES_DIR, "semantic_clusters.csv")
 
 def main():
     io_args, extra = parse_io_args()
+    # Output lands under data/derived/tables/ (gitignored, regenerable — ticket 0233);
+    # create it so validate_io's dir check passes on a clean tree.
+    os.makedirs(os.path.dirname(io_args.output) or ".", exist_ok=True)
     validate_io(output=io_args.output)
 
     global FIGURES_DIR
--- a/scripts/build_het_core.py
+++ b/scripts/build_het_core.py
@@ -379,6 +379,9 @@ def _select_with_quotas(s2):
 
 def main():
     io_args, extra = parse_io_args()
+    # Output lands under data/derived/tables/ (gitignored, regenerable — ticket 0233);
+    # create it so validate_io's dir check passes on a clean tree.
+    os.makedirs(os.path.dirname(io_args.output) or ".", exist_ok=True)
     validate_io(output=io_args.output)
 
     parser = argparse.ArgumentParser(description=__doc__)
--- a/scripts/compute_analytical_null.py
+++ b/scripts/compute_analytical_null.py
@@ -28,6 +28,7 @@ null distribution parameters, not the observed statistic).
 
 import argparse
 import math
+import os
 import sys
 
 import pandas as pd
@@ -156,6 +157,9 @@ def main() -> None:
     args = parser.parse_args(extra)
     method = args.method
 
+    # Output lands under data/derived/tables/ (gitignored, regenerable — ticket 0233);
+    # create it so validate_io's dir check passes on a clean tree.
+    os.makedirs(os.path.dirname(io_args.output) or ".", exist_ok=True)
     validate_io(output=io_args.output)
 
     cfg = load_analysis_config()
--- a/scripts/compute_crossyear_zscore.py
+++ b/scripts/compute_crossyear_zscore.py
@@ -19,6 +19,7 @@ Usage::
 """
 
 import argparse
+import os
 import re
 import sys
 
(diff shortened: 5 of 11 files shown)


MinhHaDuong (commented):
## /review-pr — multi-perspective panel (correctness · red-team · consistency · scope)

Guard test ran green in the worktree: `pytest -k Derived` → **13 passed, 3 skipped**. Ruff clean. The 8 producer script edits are correct (idiom identical, `import os` placed right, makedirs guards the real `io_args.output` dir in every case). Scope expansion 3→8 is justified and in-class (the extra 5 were surfaced by the auto-discovery guard / 0231 rebase, not creep). Nice work on the Makefile-driven discovery — it correctly resolves multi-`.mk` variable indirection.

One blocker on the standing guard itself, converged on independently by the correctness and red-team reviewers and reproduced directly.

### request-changes

**The per-script guard has no teeth for producers that already contain an unrelated earlier `os.makedirs(`.** `test_makedirs_before_validate_io` (tests/test_io_discipline.py:361-374) checks only `source.find("os.makedirs(") < source.index("validate_io(")` — first occurrence of *any* makedirs anywhere in the file, never checking the call targets `os.path.dirname(io_args.output)`. Two existing 0218-era producers already defeat it:
- `scripts/analyze_genealogy.py`: module-level `os.makedirs(TABLES_DIR, ...)` at line 40, real fix at 309 → deleting line 309 keeps `md < vi` green.
- `scripts/analyze_bimodality.py`: helper `os.makedirs(pole_dir, ...)` at line 257, real fix at 428 → same.

This directly contradicts the ticket exit criterion "reverting the makedirs on one producer turns the guard RED" and the PR body's "reverting any producer's makedirs turns the guard RED" — false for these two files, and a future producer copy-pasted from one of them inherits a silently-green guard while its output dir is never created. Suggested fix: anchor the check to the argument and adjacency, e.g. require `re.search(r"os\.makedirs\(\s*os\.path\.dirname\(io_args\.output\)", source)` appearing immediately before the `validate_io(` call, not merely "some makedirs earlier in the file."

### comment

- `verifiable:` `test_discovery_nonempty` (tests/test_io_discipline.py:352-360) hardcodes only the 3 new names. If the Makefile scan regresses and drops one of the other 5+ producers from `PRODUCERS`, the corresponding parametrization simply never runs — pytest reports fewer tests, not a failure, and nothing trips. Widen `expected` to the full known set or assert `len(found) >= N`.
- `consider:` `_derived_producer_scripts()` only scans the target line's prerequisite list for `scripts/*.py`; a future target that invokes the script from the recipe *body* instead of listing it as a prerequisite silently drops out of discovery (guard vacuously green). Today's convention always lists the script as a prerequisite, so dormant — worth a comment documenting the convention as load-bearing.
- `consider:` the makedirs idiom is now repeated 8×; folding it into `validate_io` would centralize it, but the ticket explicitly pins `validate_io`'s raise-on-missing-dir behaviour (`test_validate_io_checks_output_dir`), so deferring is the right call. Noted, not requested.
- `nofollow:` PR body says the formatter removed an unused `BASE_DIR` import from `analyze_cocitation.py`; the diff shows no such change there — that cleanup actually belongs to `compute_lexical.py`. Cosmetic provenance nit in the description only.
- `nofollow:` `?=` conditional Make assignments are not matched by the variable-resolution regex, but no `?=` exists in the Makefiles today.

Recommend tightening the guard (blocker) and widening `test_discovery_nonempty` in this PR before merge; the rest are optional.

PR #1049, opened 2026-07-15
fix(0262): repoint regression harness at post-0240 script paths

MinhHaDuong:
## Summary

- Fixes `ROOT`/`SCRIPTS_DIR` in `compute_regression_hashes.py` (was resolving one directory too deep after the script itself moved into `scripts/analysis/`) and repoints all 10 `REGISTRY` script paths to their post-0240 `scripts/analysis/`/`scripts/figures/` subdirectories.
- Fixes `_smoke_env()` to put `scripts/` + `libs/openalex-corpus/src` on `PYTHONPATH` for the subprocess — the reorg broke a previously-implicit coincidence where a script's own directory doubled as the `script_io_args` import root.
- Fixes `plot_fig_alluvial.py` and `plot_fig2_composition.py` to read `cluster_labels.json` alongside their `--input` CSV instead of a hardcoded `DERIVED_TABLES_DIR` — the former crashed as soon as Wave 2 could actually run (which it never had before, in this harness); the latter silently fell back to placeholder labels (caught in `/review-pr`'s red-team pass).
- Regenerates golden hashes for exactly the 2 scripts (`plot_fig_alluvial`, `plot_fig2_composition`) whose Wave 2 run had never previously completed — see the ticket's Invariants section for the full justification. The other 8 goldens are byte-identical.
- Adds `test_registry_script_paths_exist` as a fast-tier guard against future registry/script-path drift.

## Test plan

- [x] New guard test (`test_registry_script_paths_exist`) red before the fix, green after
- [x] `uv run pytest tests/test_regression.py` — 23/23 pass (10 regression + 13 infra/redirect)
- [x] `make check-fast` — 946 passed, 10 skipped
- [x] `make lint` — 169 passed, 5 skipped
- [x] `/verify-adherence` — PASS (one non-blocking nit noted in the review, not gating)

Ticket: none — 0262 was closed on-branch (see `tickets/closed/0262-...erg`'s Closed: header and log entry in this PR's diff).

### DIFF
--- a/scripts/analysis/compute_regression_hashes.py
+++ b/scripts/analysis/compute_regression_hashes.py
@@ -34,7 +34,7 @@ log = get_logger("compute_regression_hashes")
 # Paths
 # ---------------------------------------------------------------------------
 
-ROOT = Path(__file__).resolve().parent.parent
+ROOT = Path(__file__).resolve().parent.parent.parent
 FIXTURE_DIR = ROOT / "tests" / "fixtures" / "smoke"
 GOLDEN_PATH = FIXTURE_DIR / "golden_hashes.json"
 SCRIPTS_DIR = ROOT / "scripts"
@@ -50,7 +50,7 @@ REGISTRY: list[dict] = [
     # --- Wave 1: independent (no deps, run in parallel) ---
     {
         "name": "compute_breakpoints",
-        "script": "compute_breakpoints.py",
+        "script": "analysis/compute_breakpoints.py",
         "args": ["--output", "data/derived/tables/tab_breakpoints.csv"],
         "deps": [],
         "outputs": [
@@ -59,7 +59,7 @@ REGISTRY: list[dict] = [
     },
     {
         "name": "compute_breakpoint_robustness",
-        "script": "compute_breakpoints.py",
+        "script": "analysis/compute_breakpoints.py",
         "args": ["--output", "data/derived/tables/tab_breakpoint_robustness.csv",
                  "--robustness"],
         "deps": [],
@@ -69,7 +69,7 @@ REGISTRY: list[dict] = [
     },
     {
         "name": "compute_clusters",
-        "script": "compute_clusters.py",
+        "script": "analysis/compute_clusters.py",
         "args": ["--output", "data/derived/tables/tab_alluvial.csv"],
         "deps": [],
         "outputs": [
@@ -79,7 +79,7 @@ REGISTRY: list[dict] = [
     },
     {
         "name": "plot_fig1_bars",
-        "script": "plot_fig1_bars.py",
+        "script": "figures/plot_fig1_bars.py",
         "args": ["--output", "deliverables/_shared/figures/fig_bars.png"],
         "deps": [],
         "outputs": [
@@ -88,7 +88,7 @@ REGISTRY: list[dict] = [
     },
     {
         "name": "plot_fig1_bars_v1",
-        "script": "plot_fig1_bars.py",
+        "script": "figures/plot_fig1_bars.py",
         "args": [
             "--output", "deliverables/_shared/figures/fig_bars_v1.png",
             "--v1-only",
@@ -100,7 +100,7 @@ REGISTRY: list[dict] = [
     },
     {
         "name": "build_het_core",
-        "script": "build_het_core.py",
+        "script": "analysis/build_het_core.py",
         "args": ["--output", "tests/fixtures/smoke/catalogs/het_mostcited_50.csv"],
         "deps": [],
         "outputs": [
@@ -113,7 +113,7 @@ REGISTRY: list[dict] = [
     # harness never touches real deliverables/.
     {
         "name": "plot_fig2_breaks",
-        "script": "plot_fig2_breaks.py",
+        "script": "figures/plot_fig2_breaks.py",
         "args": ["--output", "deliverables/_shared/figures/fig_breaks.png",
                  "--input", "data/derived/tables/tab_breakpoints.csv"],
         "deps": ["compute_breakpoints"],
@@ -123,7 +123,7 @@ REGISTRY: list[dict] = [
     },
     {
         "name": "plot_fig2_composition",
-        "script": "plot_fig2_composition.py",
+        "script": "figures/plot_fig2_composition.py",
         "args": [
             "--output", "deliverables/_shared/figures/fig_composition.png",
             "--input", "data/derived/tables/tab_alluvial.csv",
@@ -135,7 +135,7 @@ REGISTRY: list[dict] = [
     },
     {
         "name": "plot_fig_alluvial",
-        "script": "plot_fig_alluvial.py",
+        "script": "figures/plot_fig_alluvial.py",
         "args": ["--output", "deliverables/_shared/figures/fig_alluvial.png",
                  "--input", "data/derived/tables/tab_alluvial.csv"],
         "deps": ["compute_clusters"],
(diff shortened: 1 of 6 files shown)


MinhHaDuong (commented):
## /review-pr — PR #1049 (ticket 0262)

Five-agent panel (Correctness, Consistency, Scope, Red team, Doc propagation), model sonnet, foreground. `make check-fast` re-run clean (946 passed, 10 skipped) — matches the PR's claim.

### Blocking finding (Red team, independently reproduced by the synthesizer)

**`scripts/figures/plot_fig2_composition.py` still has the bug this PR claims to have fixed elsewhere, and the regenerated golden bakes in the defect.**

The PR fixes `plot_fig_alluvial.py`'s hardcoded `DERIVED_TABLES_DIR` lookup for `cluster_labels.json` by deriving the path from `--input` instead. `plot_fig2_composition.py` has the identical dependency (same `compute_clusters` upstream, same co-located `cluster_labels.json`) but its REGISTRY entry passes no `--labels`, so it falls through to `load_cluster_labels()` — which reads a **module-level** `_CLUSTER_LABELS_PATH` derived from `DERIVED_TABLES_DIR`/`CLIMATE_FINANCE_DATA`, never the harness's tmp `output_root` where `compute_clusters.py` actually wrote the file for this run.

Reproduced directly:
```
CLIMATE_FINANCE_DATA=tests/fixtures/smoke uv run python -c \
  "from utils import load_cluster_labels; print(load_cluster_labels())"
# -> {0: 'Cluster 0', 1: 'Cluster 1', ...} + UserWarning: cluster_labels.json not found
```
`compute_regression_hashes.py`'s `run_and_hash()` always calls `_smoke_env()` (even under `--update-golden` — there is no separate "real" code path), so this fallback fired when the `fig_composition.png` golden was regenerated in this PR. The `raw_labels` returned feed `_format_tfidf_line()` at line 144, which renders as the figure's TF-IDF subtitle line — meaning the newly-accepted golden encodes literal "Cluster 0" / "Cluster 1" placeholder text as a manuscript figure's subtitle, not real TF-IDF terms.

This directly contradicts the ticket's own invariant claim: "regenerating them via `--update-golden` reflects the harness actually working for the first time, not an intentional analysis change." It does not — Wave 2 completing does not mean the figure it produced is correct, and nobody opened the PNG to check. This is exactly the failure mode the project's methodological-honesty norm (`feedback_manuscript_number_provenance`, `feedback_bytecheck_old_vs_new_not_golden`) exists to catch: a self-certifying `--update-golden` blessed a real defect into the baseline.

**Request**: apply the same `--input`-derived labels-path fix to `plot_fig2_composition.py` (or pass `--labels <path>` explicitly in its REGISTRY entry), re-run, visually inspect the resulting `fig_composition.png` for real TF-IDF terms (not "Cluster N"), and re-regenerate its golden only after that check.

### Non-blocking findings

- **verifiable (Consistency):** `_smoke_env()`'s PYTHONPATH construction duplicates `tests/_source_roots.py`'s `source_root_env()`/`source_root_pythonpath()`, the repo's existing shared helper for this exact purpose, already reused by other subprocess-launching tests. Two independent implementations of the same convention now exist; consider switching to the shared helper in a follow-up.
- **verifiable (Consistency):** `scripts/analysis/compute_regression_history.py`, a sibling registry-driven tool moved by the same reorg, has the identical unfixed bug class (shallow `ROOT`, stale flat `SCRIPTS` paths). Not CI-wired today (only referenced in `test_script_hygiene.py` allowlists), so it's latent rather than active breakage — worth a follow-up ticket, not blocking this PR.
- **consider (Correctness, low confidence):** `plot_fig_alluvial.py`'s new `labels_dir = os.path.dirname(io_args.input[0])` returns `""` if `--input` is ever passed as a bare filename with no directory component, silently resolving relative to cwd instead of erroring. No current call site hits this, so not blocking.

### Panel verdicts (dissent preserved)

| Agent | Verdict | Confidence |
|---|---|---|
| Correctness | approve | high (one low-confidence "consider" on the same area red team confirmed) |
| Consistency | comment | high |
| Scope | approve | high |
| Red team | **request-changes** | high |
| Doc propagation | approve | high |

Four of five agents approved; Red team's finding is confirmed by independent reproduction above and is the deciding factor. Scope, Consistency, and Doc-propagation findings are otherwise clean — the PR is tightly scoped to its ticket, doesn't touch unrelated files, and no stale docs reference the pre-fix state.

**Overall: request-changes.** The path/PYTHONPATH/ROOT fixes and the guard test are correct and well-tested; the `plot_fig2_composition.py` labels bug must be fixed before the golden regeneration for that script can be trusted.

MinhHaDuong (commented):
## Review panel — PR #1049 (5 perspectives: correctness, consistency, scope, red-team, doc-propagation)

**Overall: approve**, with one item to fix before merge and several non-blocking nits/follow-ups.

### Fix before merge

- **verifiable: ticket-close line vs. PR convention** (scope agent). The ticket was already closed on-branch (`tickets/0262-...erg` moved to `tickets/closed/0262-...erg` with a `Closed:` header), but the PR body's `**Ticket:**` line still points at the archived path instead of `Ticket: none`. Per the project's recorded convention (a ticket closed on-branch uses `Ticket: none`, not the archived path — `erg-pr-merge` expects to find and close the ticket at the *open* path), this risks a "no ticket found" bounce at merge time. Please update the `**Ticket:**` line to `Ticket: none`.

### Non-blocking findings

- **consider: no visual verification of the 2 regenerated golden hashes** (red-team). `plot_fig_alluvial` and `plot_fig2_composition`'s new goldens rest entirely on a hash-matches-itself claim plus a narrative ("Wave 2 never completed before"). Nothing in the PR shows the two regenerated PNGs were actually eyeballed. Given this repo's own precedent that hashing alone can't catch a wrong-but-plausible rendering, worth a quick visual check of `fig_alluvial.png`/`fig_composition.png` before merge, even though it's not a hard blocker.
- **consider: inconsistent missing-file behavior between the two sibling fixes** (correctness + red-team). `plot_fig2_composition.py`'s `_resolve_cluster_labels()` silently falls back to placeholder labels via `load_cluster_labels()` when `cluster_labels.json` isn't found alongside `--input`; `plot_fig_alluvial.py` hard-crashes (`FileNotFoundError`) on the same miss. Same bug class, two different failure modes — the silent one is the more dangerous of the two long-term (a future path regression there would produce wrong-but-valid-looking labels instead of failing loudly).
- **consider: duplicated cluster-labels-resolution logic** (consistency). Both scripts solve "find `cluster_labels.json` alongside `--input`" independently — one as a named helper, one inline. A shared helper (e.g. in `utils.py`, already imported by both) would remove the duplication and prevent the next reorg from needing the same fix twice.
- **consider: `test_registry_script_paths_exist` is existence-only, not identity-checked** (correctness + red-team). It confirms every REGISTRY script path resolves on disk, but wouldn't catch a REGISTRY entry pointed at a wrong-but-existing script (e.g. a copy-paste swap) — that would still pass the guard while silently comparing hashes for the wrong figure. Could be strengthened later (e.g. assert script basename relates to entry `name`) but not required for this fix.
- **nofollow: pre-existing staleness, out of scope for this PR** (doc-propagation). `STATE.md` still lists ticket 0262 as an open/current item, and `.claude/rules/script-io.md` names `plot_fig1_bars.py` without its post-reorg subdirectory. Neither is touched by this diff and both predate it — flagging for a future housekeeping pass, not this PR.

### Confirmed correct (no action needed)
- `ROOT = Path(__file__).resolve().parent.parent.parent` depth arithmetic is correct for the script's new location (`scripts/analysis/compute_regression_hashes.py` → repo root).
- The `_smoke_env()` `PYTHONPATH` conditional (`os.environ["PYTHONPATH"]` inside the truthy branch) is safe — Python only evaluates the taken branch, so there's no KeyError risk.
- The `golden_hashes.json` diff is scoped to exactly the 2 claimed entries; the other 8 are untouched.
- Ticket housekeeping format (`Closed:` header, closing log line) matches other closed tickets.
- No stale references to the old flat script paths found elsewhere in the repo (Makefile, docs, other tests already use the post-reorg subpaths).

### Process note
This repo has no CI configured (known/expected — `reference_repo_no_ci`), so the 23/23 `test_regression.py` and `make check-fast`/`make lint` results in the PR body are the author's self-reported local runs, not independently re-verified by this panel.

PR #1063, opened 2026-07-17
fix(0265): plot_ncc_alluvial reads cluster_labels.json alongside --input

MinhHaDuong:
## Summary

`scripts/figures/plot_ncc_alluvial.py` gated its primary `tab_alluvial.csv` read
on `--input` but read the secondary `cluster_labels.json` unconditionally from
the module-level `DERIVED_TABLES_DIR` constant. When `--input` redirects the data
elsewhere (test harness, companion-paper variant), the labels file was looked up
in the wrong place. This is the third instance of the bug shape fixed for
`plot_fig_alluvial.py` and `plot_fig2_composition.py` in ticket 0262 (PR #1049);
the `roar` sweep after that PR found this un-fixed sibling.

## Changes

1. **Fix** (`plot_ncc_alluvial.py`): when `--input` is given, derive
   `labels_dir = os.path.dirname(io_args.input[0])`; otherwise fall back to
   `DERIVED_TABLES_DIR`. Identical pattern to the `plot_fig_alluvial.py`
   reference fix. The default (no `--input`) path is unchanged, so the real
   `make` build (`--input data/derived/tables/tab_alluvial.csv`) behaves exactly
   as before. `tab_core_shares.csv` stays hardcoded + `os.path.exists`-guarded,
   matching the reference (intentional, out of scope).

2. **Standalone regression test** (`tests/test_ncc_alluvial_labels_path.py`,
   `@pytest.mark.integration`): runs the script with `--input` at a tmp dir
   holding both files and `CLIMATE_FINANCE_DATA` pointed at an empty tree, so a
   buggy fallback to `DERIVED_TABLES_DIR` crashes. Confirmed red on the buggy
   code (`FileNotFoundError` at the labels read), green after the fix.

3. **Regression harness** (`compute_regression_hashes.py` REGISTRY + golden):
   registered `plot_ncc_alluvial` (Wave 2, deps `compute_clusters`, `--input`
   redirected) so this bug class stays caught by the harness like its two
   siblings — closing the gap the ticket names (this script was invisible to the
   regression suite). Golden regenerated surgically: only the new
   `fig_ncc_alluvial.png` hash was added; existing baselines verified unchanged
   (`--check` green before and after) and deterministic across two same-machine
   runs.

## TDD

Red (test) → green (fix) → refactor (REGISTRY), one commit each.

## Verification

- `ruff check` clean on all changed files.
- New standalone test + `test_regression_plot_ncc_alluvial` + infra guards pass.
- `make check-fast` / `make lint`: the only failures are **pre-existing,
  causally disjoint** manuscript-workpackage reds — `test_cited_works_available`
  (`michalopoulos2017` unallowlisted, added in `0d096444`) and 7
  `test_manuscript_prose` content assertions. My diff touches none of
  `deliverables/`, `config/`, or the prose tests; all are present on
  `origin/main` independent of this branch.

**Ticket:** tickets/0265-plot-ncc-alluvial-hardcoded-cluster-labels.erg

### DIFF
--- a/scripts/analysis/compute_regression_hashes.py
+++ b/scripts/analysis/compute_regression_hashes.py
@@ -144,6 +144,19 @@ REGISTRY: list[dict] = [
         ],
     },
     {
+        # NCC-format alluvial variant. Registered so the cluster_labels.json
+        # path bug (ticket 0265 — read from --input dir, not DERIVED_TABLES_DIR)
+        # stays caught by the harness, like its plot_fig_alluvial sibling.
+        "name": "plot_ncc_alluvial",
+        "script": "figures/plot_ncc_alluvial.py",
+        "args": ["--output", "deliverables/_shared/figures/fig_ncc_alluvial.png",
+                 "--input", "data/derived/tables/tab_alluvial.csv"],
+        "deps": ["compute_clusters"],
+        "outputs": [
+            "deliverables/_shared/figures/fig_ncc_alluvial.png",
+        ],
+    },
+    {
         "name": "plot_fig_breakpoints",
         "script": "figures/plot_fig_breakpoints.py",
         "args": ["--output", "deliverables/_shared/figures/fig_breakpoints.png",
--- a/scripts/figures/plot_ncc_alluvial.py
+++ b/scripts/figures/plot_ncc_alluvial.py
@@ -216,18 +216,23 @@ def main():
     _apply_ncc_style()
 
     # --- Load data ---
+    # --input takes precedence for the alluvial CSV; cluster_labels.json is
+    # written by compute_clusters to the same directory, so derive its path the
+    # same way rather than hardcoding DERIVED_TABLES_DIR (ticket 0265).
     if io_args.input:
         alluvial_data = pd.read_csv(io_args.input[0], index_col=0)
+        labels_dir = os.path.dirname(io_args.input[0])
     else:
         alluvial_data = pd.read_csv(
             os.path.join(DERIVED_TABLES_DIR, "tab_alluvial.csv"), index_col=0,
         )
+        labels_dir = DERIVED_TABLES_DIR
     alluvial_data.columns = alluvial_data.columns.astype(int)
     period_labels = alluvial_data.index.tolist()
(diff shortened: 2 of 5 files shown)


MinhHaDuong (commented):
Correctness-only review (trivial-risk small-tier PR, per project policy — consistency/scope/red-team/doc-propagation skipped).

**Verified:**
- `labels_dir` derivation exactly mirrors the reference fix in `plot_fig_alluvial.py` (ticket 0262 / PR #1049): `os.path.dirname(io_args.input[0])` when `--input` is given, `DERIVED_TABLES_DIR` fallback otherwise — same idiom, no divergence.
- `tab_core_shares.csv` handling (line 241, guarded by `os.path.exists`) is untouched by the diff, as claimed — confirmed via `git diff origin/main...HEAD -- scripts/figures/plot_ncc_alluvial.py | grep tab_core_shares` (no hits).
- New test `tests/test_ncc_alluvial_labels_path.py`: confirmed red on pre-fix code (temporarily swapped in the old script body, got the expected `FileNotFoundError` on `DERIVED_TABLES_DIR/cluster_labels.json`) and green on the fixed code. Test logic is sound — it redirects `CLIMATE_FINANCE_DATA` to an empty tree so only a correctly-derived `labels_dir` can succeed.
- `golden_hashes.json` diff is a pure addition (`plot_ncc_alluvial` entry, 64-char hex hash, consistent with all sibling entries) — no other entries modified.
- New `REGISTRY` entry in `compute_regression_hashes.py` is well-formed; its `deps: ["compute_clusters"]` resolves to an existing registry entry.
- `make check-fast`-equivalent (`pytest -m "not slow and not integration and not adherence"`): 945 passed, 1 unrelated pre-existing failure (`test_cited_works_have_local_fulltext`, missing fulltext for `michalopoulos2017`) — this PR touches no `.bib`/manuscript files, so unrelated to this change.

No blocking findings.

nofollow: the new test is marked `@pytest.mark.integration` (subprocess-based, correctly tiered per project convention) so it doesn't run under the fast tier — consistent with how the reference PR's tests were tiered, not a defect here.

verdict: approve

PR #1129, opened 2026-07-27
ticket(0297): persist pass-2 language detections, with provenance

MinhHaDuong:
**Ticket:** tickets/0297-corpus-language-null-rate-regression.erg

The language null rate has sat at 4.0% against a <2% contract for weeks, and
the count never moved: **exactly 1,743 nulls before and after the corpus-v2
rebuild**, with the three new v2 sources (UNFCCC, OECD, ISTEX) contributing
zero of them. That stability was the clue. This is not a source that skips
language — it is a step that never lands.

## Root cause

`enrich_language.py` is cache-only by design: it never writes the works CSV
back, and `enrich_join.py` applies its caches to the monolith. Pass 1 obeys
that, saving to `language_resolved`. **Pass 2 did not.** `pass2_local_detect`
mutated the in-memory frame, `main()` logged `Pass 2 filled N`, and the frame
was dropped on return — its own comment says "Cache-only", but pass 2 wrote to
no cache. Roughly 1,500 langdetect results were computed and discarded on
every single run.

The leftover population matches exactly: of the 1,743 nulls, 1,579 have a
title and 1,306 have an abstract — precisely what pass 2 is for. Only 164 have
no title at all, the irreducible floor. By source: OpenAlex 1,427 (82% of all
nulls), bibCNRS 209 (**90% of that source's 233 rows**), SciSpace 66,
teaching 42.

**Why the tests missed it.** `TestPass2LocalDetect` asserts the in-memory
mutation, which always worked. The new `TestPass2Persistence` asserts the
persisted artifact.

## The fix, and one design decision

Pass 2 now writes to its **own** cache, deliberately not merged into pass 1's:

| Cache | Values |
|---|---|
| `language_resolved` | **sourced** from the OpenAlex API |
| `language_detected` | **inferred** by langdetect from title + abstract |

Merging them would make an inferred tag indistinguishable from a sourced one
inside a deposited dataset — in a data paper about data quality, whose referee
probed exactly this. Today every language value in the corpus is sourced;
this change introduces inferred ones, so it needs disclosure.

`enrich_join` applies sourced first, inferred only to what remains null, and
records which in a new `language_provenance` column (`source` / `openalex` /
`detected:langdetect`), declared in the deposit contract alongside the
existing `abstract_provenance` and `keywords_provenance` — the precedent this
follows.

## Expected effect

~4.0% → ~0.4%, every filled value carrying how it was obtained.

## Note on running it

`make corpus` guards on `main`, so the rebuild that realises this cannot run
from the branch. The caches are being warmed now (`enrich_language.py` writes
only to `enrich_cache/`, which is gitignored and not a DVC output), so the
rebuild after merge is cheap. **`lang_english_pct` and `tab_languages` will
move** — this needs to land before the response letter quotes them, not after.

## Verification

- `tests/test_enrich_language.py` — 41 passed, incl. 5 new persistence tests
  red-tested before the fix
- 176 passed across the enrich/deposit/codebook suites
- ruff clean

### DIFF
--- a/scripts/_deposit_variables.py
+++ b/scripts/_deposit_variables.py
@@ -133,6 +133,13 @@ DEPOSIT_VARIABLES: list[Variable] = [
              "or `generated:lexicon`; empty elsewhere", _KEYDOCS,
              required=False, group=_PROV,
              allowed_values="extracted, generated:lexicon, empty"),
+    Variable("language_provenance", "string, nullable",
+             "How the language code was obtained: `source` (carried by the "
+             "source catalog), `openalex` (backfilled from the OpenAlex API), "
+             "or `detected:langdetect` (inferred from title and abstract); "
+             "empty where no language could be established", _ENRICH,
+             required=False, group=_PROV,
+             allowed_values="source, openalex, detected:langdetect, empty"),
     Variable("source_count", "integer",
              "Number of sources that contributed the record (sum of the "
              "provenance flags)", _MERGE, group=_PROV,
--- a/scripts/harvest/enrich_join.py
+++ b/scripts/harvest/enrich_join.py
@@ -148,25 +148,56 @@ def _apply_abstract_caches(df, cache_dir):
 
 
 def _apply_language_cache(df, cache_dir):
-    """Normalize language codes and fill missing from cache."""
-    df["language"] = df["language"].apply(normalize_lang)
-    lang_cache = _load_csv_cache(
-        os.path.join(cache_dir, "language_resolved.csv"), "key", "language")
+    """Normalize language codes and fill missing from the two language caches.
 
-    applied = 0
+    Order matters: `language_resolved` holds values *sourced* from OpenAlex,
+    `language_detected` holds values *inferred* by langdetect from title and
+    abstract (ticket 0297). Sourced wins; inference only fills what is still
+    null afterwards, and every inferred value is disclosed in
+    `language_provenance` so a deposited language tag never hides how it was
+    obtained. Mirrors abstract_provenance / keywords_provenance.
+    """
+    df["language"] = df["language"].apply(normalize_lang)
+    resolved = _load_csv_cache(
+        os.path.join(cache_dir, "language_resolved.csv"), "key", "language")
+    detected = _load_csv_cache(
+        os.path.join(cache_dir, "language_detected.csv"), "key", "language")
+
+    # Anything already present before either cache is applied came with the
+    # record from its source catalog.
+    provenance = pd.Series("", index=df.index, dtype=object)
+    provenance[~df["language"].apply(_is_missing)] = "source"
+
+    def _lookup(cache, doi, sid):
+        lang = cache.get(doi, "") if doi else ""
+        return lang or cache.get(sid, "")
+
+    applied_resolved = applied_detected = 0
     for idx in df.index:
-        if _is_missing(df.at[idx, "language"]):
-            doi = normalize_doi(df.at[idx, "doi"])
-            sid = str(df.at[idx, "source_id"])
-            lang = lang_cache.get(doi, "") if doi else ""
-            if not lang:
-                lang = lang_cache.get(sid, "")
-            if lang:
-                df.at[idx, "language"] = lang
-                applied += 1
-    log.info("Language cache: applied %d (cache has %d entries)",
-             applied, len(lang_cache))
-    return applied
+        if not _is_missing(df.at[idx, "language"]):
+            continue
+        doi = normalize_doi(df.at[idx, "doi"])
+        sid = str(df.at[idx, "source_id"])
+
+        lang = _lookup(resolved, doi, sid)
+        if lang:
+            df.at[idx, "language"] = lang
+            provenance.at[idx] = "openalex"
+            applied_resolved += 1
+            continue
+
+        lang = _lookup(detected, doi, sid)
+        if lang:
+            df.at[idx, "language"] = lang
+            provenance.at[idx] = "detected:langdetect"
+            applied_detected += 1
+
+    df["language_provenance"] = provenance
+    log.info(
+        "Language caches: applied %d sourced (%d entries) + %d detected (%d entries)",
+        applied_resolved, len(resolved), applied_detected, len(detected),
+    )
+    return applied_resolved + applied_detected
 
 
 def _apply_summaries(df, cache_dir):
--- a/scripts/harvest/enrich_join.py
+++ b/scripts/harvest/enrich_join.py
@@ -18,7 +18,7 @@ import os
 
 import pandas as pd
 from openalex_corpus.text import normalize_doi
-from pipeline_text import normalize_lang
+from pipeline_text import is_valid_iso639_1, normalize_lang
 from utils import CATALOGS_DIR, get_logger, save_csv
 
 log = get_logger("enrich_join")
@@ -163,10 +163,19 @@ def _apply_language_cache(df, cache_dir):
     detected = _load_csv_cache(
         os.path.join(cache_dir, "language_detected.csv"), "key", "language")
 
-    # Anything already present before either cache is applied came with the
-    # record from its source catalog.
+    def _needs_filling(val):
+        """Missing, or present but not a real ISO 639-1 code.
+
+        A code the codebook calls ISO 639-1 but that no standard recognises
+        ('ua', 'sh') is not usable data; pass 2 already re-detects those, and
+        treating them as present would discard the correction (ticket 0297).
+        """
+        return _is_missing(val) or not is_valid_iso639_1(val)
+
+    # Anything usable before either cache is applied came with the record
+    # from its source catalog.
     provenance = pd.Series("", index=df.index, dtype=object)
-    provenance[~df["language"].apply(_is_missing)] = "source"
+    provenance[~df["language"].apply(_needs_filling)] = "source"
 
     def _lookup(cache, doi, sid):
         lang = cache.get(doi, "") if doi else ""
@@ -174,7 +183,7 @@ def _apply_language_cache(df, cache_dir):
 
     applied_resolved = applied_detected = 0
     for idx in df.index:
-        if not _is_missing(df.at[idx, "language"]):
+        if not _needs_filling(df.at[idx, "language"]):
             continue
         doi = normalize_doi(df.at[idx, "doi"])
         sid = str(df.at[idx, "source_id"])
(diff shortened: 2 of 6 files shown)


MinhHaDuong (commented):
Reviewed the diff (`origin/main...a4549de`) plus the full current bodies of `enrich_language.py`, `enrich_join.py`, `pipeline_text.py`, `schemas.py`, `dvc.yaml`, and `.gitignore`. Ran the touched test modules and ruff against a clean checkout of the PR head — all green (53/53 in `test_enrich_language.py` + `test_join_enrichments.py`, 34 passed/1 skipped in `test_variables_table.py` + `test_doc_contract_consistency.py`, ruff clean on all 4 changed files).

**Correctness — looks sound.**
- Two-cache split (`language_resolved` sourced vs `language_detected` inferred) and the sourced-wins-then-inferred-fallback order in `_apply_language_cache` is correct and well tested (`TestPass2Persistence`, `TestInvalidCodesAreReplaced`).
- `_needs_filling` correctly re-checks `is_valid_iso639_1` so a bogus existing code (`ua`, `sh`) is treated as fillable, closing the second bug described in the ticket body (invalid codes previously survived the join even after pass 2 corrected them in-cache).
- `is_valid_iso639_1` is imported from `pipeline_text` in `enrich_join.py` and re-exported from the same module via `utils` in `enrich_language.py` — no risk of the two call sites diverging on what counts as valid.
- Cache persistence semantics match pass 1's existing pattern (load whole dict, mutate, save whole dict back) — accumulates across runs, never drops previously-detected entries even when a row is later resolved a different way.
- `enrich_cache/` (and hence the new `language_detected.csv`) is already covered by the wholesale gitignore/dvcignore entry and the DVC stage only declares the stamp file as output, so no `dvc.yaml`/`.gitignore` update was needed — checked, and the PR correctly didn't touch either.
- New `language_provenance` column is only introduced downstream in `enrich_join.py` (enriched_works.csv), not at the `refined_works.csv` Phase 1→2 boundary, so `schemas.py`'s `RefinedWorksSchema` correctly doesn't need updating either (verified no `EnrichedWorksSchema` exists for the monolith stage).

**Minor / non-blocking:**
1. `enrich_language.py`'s module docstring still says "Cache: data/catalogs/enrich_cache/language_resolved.csv" (singular) — now stale with two caches. Small doc nit, not worth a blocking round.
2. Edge case not covered by a test: a row whose original `language` is an invalid code (e.g. `ua`) with no hit in either cache keeps the invalid junk value in `language` but gets `language_provenance = ""` (i.e. "empty" per the codebook's `allowed_values`). That reads as "no language could be established" when the field in fact still holds a bogus code. Extremely small population (ticket says ~5 such rows historically) and pre-existing data quality issue, not a regression — flagging only because the provenance column's own contract ("empty where no language could be established") is technically not quite true for this sliver of rows.
3. `_apply_language_cache`'s docstring/design note is good and explains the "why" well; nothing else to add there.

No blocking issues found. The design decision to keep the two caches separate (rather than merging into `language_resolved`) is the right call given the stated goal of not letting an inferred tag masquerade as sourced data in the deposit.

MinhHaDuong (commented):
## Multi-perspective review — PR 1129 (ticket 0297)

Ran the full five-perspective panel. Verified via: `gh pr diff`, `git show FETCH_HEAD:<path>` for each changed file in full (not just hunks), an archived checkout of the branch tree run through the project's own `.venv` (pytest + ruff), and a direct read of `data/catalogs/enriched_works.csv` on disk (82MB, dated Jul 24) to check the corpus's actual current null rate.

### ticket_ref_mismatch — blocking, procedural

The PR body's `**Ticket:**` line reads `tickets/0297-corpus-language-null-rate.erg`. The real file is `tickets/0297-corpus-language-null-rate-regression.erg` (confirmed: `ls tickets/ | grep '^0297'` returns exactly one match, and it is not the string in the PR body). `erg-pr-merge` closes tickets by matching this literal path — as written, the auto-close will not find the ticket and the ticket will be left open after merge. Fix: correct the `**Ticket:**` line before merge.

### Correctness — comment (one exit-criteria gap, otherwise sound)

- Pass 2 persistence: fixed correctly. `pass2_local_detect(df, cache=None)` now writes `{key: lang}` into a caller-supplied dict keyed by normalized DOI, else `source_id` — exactly the key convention `enrich_join._apply_language_cache` resolves by. Verified with the branch's own new tests (`TestPass2Persistence`, 5/5) plus a full `test_enrich_language.py` run in an archived checkout: 43 passed.
- `_apply_language_cache`: sourced (`language_resolved`) is looked up before inferred (`language_detected`) on every row that needs filling, so "sourced wins" holds. Provenance is computed correctly per branch: pre-existing valid code → `source`; filled from resolved → `openalex`; filled from detected → `detected:langdetect`; still null after both → `""` (empty), matching the codebook wording "empty where no language could be established."
- Invalid-ISO fix (2nd commit): `_needs_filling()` now covers `_is_missing(val) or not is_valid_iso639_1(val)`. `normalize_lang()` passes any 2-letter code through unvalidated (by design — that's *why* `ua`/`sh` reach the join at all), and `is_valid_iso639_1` checks against a 183-entry frozenset I independently checked excludes both `ua` and `sh` (neither is a real/current ISO 639-1 code) and doesn't include deprecated forms like `iw`/`in` either — consistent, not overly permissive.
- **Exit-criteria gap (untagged, real):** the ticket's stated exit criterion is "Language null rate < 2% on the current corpus, test green" — i.e. `tests/test_corpus_acceptance.py::test_language_null_rate_below_2_percent`. I read `data/catalogs/enriched_works.csv` directly off disk right now: **1,743/43,179 = 4.04% null, and it has no `language_provenance` column** — identical to the pre-fix numbers quoted in the PR body. That test currently either skips (no local data) or fails (stale committed data); it is not green yet. The PR is explicit about this ("rebuild ... cannot run from the branch," "caches are being warmed now") and the sequencing (code merges → `make corpus`/`make corpus-sync` on padme → doudou) matches the ticket's stated invariant, so this isn't a code defect — but it means the ticket's own exit criterion is *not yet satisfied* at merge time and needs a tracked follow-through (the rebuild + a green run of that specific test) before the ticket can be considered actually done, not just code-complete.

### Consistency — approve, one scope note

- `language_detected.csv` uses the identical on-disk shape as `language_resolved.csv` (`key,language` columns, same `load_cache`/`save_cache` in `enrich_language.py`, same `_load_csv_cache(path, "key", "language")` reader in `enrich_join.py`) and lives in the same `enrich_cache/` (gitignored, non-DVC, per the incremental-cache convention in the project's architecture notes).
- `language_provenance` follows the `abstract_provenance`/`keywords_provenance` pattern (string/nullable, `required=False`, `group=_PROV`, `allowed_values="..., empty"` trailing sentinel) — the analogy holds structurally. One difference worth naming: the precedent pair is scoped to `_KEYDOCS` (curated key documents only), while `language_provenance` is `_ENRICH` (applies corpus-wide). That's a correct difference given language backfill isn't keydocs-only, not an inconsistency, but it means "mirrors abstract_provenance/keywords_provenance" in the docstring is true in shape, not in scope.
- `consider:` `scripts/schemas.py`'s `RefinedWorksSchema` declares explicit `Column(...)` entries for `abstract_provenance` and `keywords_provenance` but gets none for `language_provenance` in this diff. The schema is `strict=False` ("allow extra columns, forward-compatible") so nothing breaks, but the PR is otherwise careful to extend precedent-following declarations (deposit contract) and silently skips this one.

### Scope — comment

- The core fix (persistence + provenance) is fully traceable to ticket 0297.
- The second commit ("treat an invalid ISO code as missing in the join") is bundled but not scope creep: it's the same failure mode (a value pass 2 corrects that never reaches the deposit), discovered while warming the cache for this exact ticket, and it's small (one predicate change + 2 tests). Legitimate same-ticket fix, not a separate concern smuggled in.

### Red-team — comment

- Fresh checkout / no warmed cache: `_load_csv_cache` returns `{}` gracefully when a cache CSV is absent (`if not os.path.exists(path): ... return {}`), so first-run behavior degrades to "nothing filled, all still needing detection" rather than crashing. Confirmed by running the archived branch tree's test suite with no `enrich_cache/` present — no crash, `test_language_null_rate_below_2_percent` skips cleanly (fixture-level skip on missing `enriched_works.csv`) rather than erroring.
- `consider:` `pass2_local_detect` does not skip rows whose key is already present in the passed-in `cache` — it re-runs `detect_language()` on every not-yet-valid row on every invocation, then overwrites the cache entry. Since `langdetect` is seeded for determinism, this doesn't produce a correctness bug (same input → same output), but it means the "persist so we don't recompute" fix doesn't actually save the ~1,500 recomputations on a second run — only saves the *discard*, not the *recompute*. Not required by the ticket, but slightly undercuts the "cheap rebuild after merge" framing in the PR body if run repeatedly.
- `nofollow:` `language_detected.csv` (and `language_resolved.csv`) only ever gains entries, never prunes keys for rows that leave the corpus (dedup, source removal). Harmless accumulation, not a correctness issue at current corpus scale.
- `verifiable:` no test exercises the "still-null after both caches" path for `language_provenance` (should be `""`) or the case where a row's language stays an uncorrected invalid code because neither cache has a fix for it — both are implied-correct by code reading (confirmed above) but untested. Test coverage gap, not a bug.

### Doc-propagation — approve

- `deliverables/_shared/tables/codebook.md` is a generated artifact (`Makefile:288`, built from `scripts/_deposit_variables.py` via `export_codebook.py`) — correctly left untouched; it regenerates with the new `language_provenance` row on next build. No hand-maintained doc needs a manual edit here that wasn't made.
- No other reference to the old single-cache/no-provenance language behavior was found elsewhere in the repo (`git grep` for `abstract_provenance`/`keywords_provenance` as the closest precedent shows all touch points already covered: `_deposit_variables.py`, `schemas.py` partially — see consistency note above, `catalog_merge.py`, `catalog_keydocs.py` — none of which need a language-specific update).

### Verification run (this review)

- `ruff check` on all 5 touched files: clean.
- `pytest tests/test_enrich_language.py`: 43 passed (archived branch checkout, project's own `.venv`).
- `pytest tests/test_join_enrichments.py tests/test_language_table.py`: 19 passed.
- `pytest tests/ -k "deposit or codebook or schema_contract or phase_layout"`: 17 passed, 2 skipped, 1 unrelated failure (`test_no_committed_symlink_under_data` — fails only because the archived tree has no `.git`; not a PR defect, artifact of my test method).
- Direct read of the live `data/catalogs/enriched_works.csv`: 4.04% null, no `language_provenance` column — confirms the exit criterion is not yet met on-disk (expected, per the PR's own note on sequencing).

PR #1130, opened 2026-07-27
ticket(0321): extract the TF-IDF label vocabulary out of a Tier-3 entry point

MinhHaDuong:
**Ticket-ref:** tickets/0321-main-is-red-8-ruff-findings-and-a-tier-3.erg

Second half of 0321 (the lint half is #1128). Resolves the dual-role import by
extraction, per the author's call.

`analyze_global_map` imported `LABEL_STOPWORDS` and `_collapse_acronyms` from
`compute_clusters`, a **Tier-3 entry point**. The `scripts/` reorg (epic 0240)
rests on Tier-3 files being import-leaves — nothing may import them by flat
name, or moving one between phase directories breaks its importer. That is the
invariant `test_no_tier3_entry_point_is_imported` guards, failing on `main`
since `10a8c029`.

## Extraction, not reclassification

The guard offers both exits. `RECLASSIFIED_DUAL_ROLE` is for genuinely
reusable computational libraries wearing a thin `main()` —
`compute_divergence`'s dispatch registry, the null-model drivers.
`compute_clusters` is an entry point that happened to own two label helpers;
reclassifying it would have used the escape hatch instead of fixing the thing.

## Why all three names move together

They are one unit. Both consumers label groups of works by TF-IDF
distinctiveness over the same corpus:

- `compute_clusters` → semantic k-means clusters → `fig_sem_composition`
  panel subtitles
- `analyze_global_map` → Louvain citation communities →
  `fig_global_map_direct` node labels

A term suppressed in one and kept in the other makes the two figures disagree
about which terms characterise a group of the same works. `_label_vocabulary`
documents that shared meaning rather than just holding the constants.

## Verified as a pure move

Not eyeballed — the definitions were rebuilt from `origin/main` in an isolated
namespace and compared:

- `LABEL_STOPWORDS` — equal, 60 terms
- `ACRONYM_EXPANSIONS` — equal, 8 patterns
- `collapse_acronyms` — identical output on five samples covering every
  pattern plus a no-op input

`re` left `compute_clusters` with the only function that used it. The
underscore-private spelling was dropped at both call sites rather than
aliased — two call sites is not worth a compatibility shim.

## Verification

- `tests/test_label_vocabulary.py` (new) — both consumers import it, neither
  keeps a local copy that could drift, neither imports the other
- `tests/test_script_classification.py` green — the guard this fixes
- 76 passed across classification + hygiene + the new file
- `ruff check scripts/` clean (the remaining `conception/` findings are #1128's)

### DIFF
--- a/docs/repo-layout.md
+++ b/docs/repo-layout.md
@@ -164,13 +164,16 @@ required; the flat classification already makes every mover import-leaf.
 
 Nine files were entry-point-AND-imported. Three had a genuine helper leaked from
 an output-producing entry point → **extracted to a neutral flat `_`-module (the
-0250 pattern), byte-identical by construction**, freeing the entry point to move:
+0250 pattern), byte-identical by construction**, freeing the entry point to move.
+A fourth case (0321) arrived later by the same route and took the same
+resolution — the pattern is the standing answer, not a one-off audit result:
 
 | Entry point (now Tier-3) | Extracted helper(s) → flat module | Repointed importer(s) |
 |---|---|---|
 | `summarize_core_venues` | `canonical_venue`/`venue_type`/`institution_group` → `_venue_naming.py` | `compute_venue_concentration`, `export_core_venues_markdown` |
 | `build_het_core` | `is_global_south`/`is_non_english` → `_corpus_predicates.py` | `analyze_multilingual` |
 | `build_teaching_yaml` | `_dedup_course_names` → `_course_dedup.py` | `analyze_syllabi` |
+| `compute_clusters` | `LABEL_STOPWORDS`/`ACRONYM_EXPANSIONS`/`collapse_acronyms` → `_label_vocabulary.py` | `analyze_global_map` (ticket 0321) |
 
 The other six were **reclassified Tier-2** (see above). One correction to the old
 audit: `qa_near_duplicates` has no `__main__` and its docstring documents a
--- a/scripts/_label_vocabulary.py
+++ b/scripts/_label_vocabulary.py
@@ -1,0 +1,59 @@
+"""Shared vocabulary for TF-IDF cluster and community labelling (ticket 0321).
+
+Two consumers label groups of works by their distinctive TF-IDF terms and must
+agree on what counts as distinctive: `compute_clusters` (semantic k-means
+clusters, whose top terms become the fig_sem_composition panel subtitles) and
+`analyze_global_map` (Louvain citation communities, whose top terms label the
+fig_global_map_direct nodes). A term suppressed in one and kept in the other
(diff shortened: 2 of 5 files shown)


MinhHaDuong (commented):
Reviewed the diff as a general code-review pass (correctness, style, edge cases).

**Summary:** Clean, well-scoped extraction. Verified by reading the actual PR-head content (not just the diff):

- `scripts/_label_vocabulary.py` is a faithful move of `LABEL_STOPWORDS`, `ACRONYM_EXPANSIONS`, and `collapse_acronyms` (renamed from `_collapse_acronyms`) — matches the PR description's claim of an equal 60-term/8-pattern set and identical `collapse_acronyms` behavior.
- Both consumers (`compute_clusters.py`, `analyze_global_map.py`) now import from the new `_`-prefixed module and no longer keep a local copy — checked with `git grep` across the full tree at the PR head for any stale `from compute_clusters import` / `_collapse_acronyms` references outside the new test file; none found.
- `import re` was correctly dropped from `compute_clusters.py` (no remaining direct `re.` use there) and correctly kept in `analyze_global_map.py`, which still calls `re.sub` directly at line ~138 for keyword cleaning — good catch on not over-pruning.
- Import ordering in both consumers is isort-clean (alphabetical, `_label_vocabulary` sorts correctly among the flat-surface imports).
- No leftover blank-line artifacts at the deletion site in `compute_clusters.py`.
- New module follows the existing Tier-3-avoidance convention pinned by `tests/test_script_classification.py::test_no_tier3_entry_point_is_imported` (underscore-prefixed flat module = library surface); no `docs/repo-layout.md` update needed since that doc references the convention generically rather than enumerating modules by name.
- `tests/test_label_vocabulary.py` is solid: pins both "consumer imports the shared module" and "no local re-definition that could silently drift," plus a `test_no_consumer_imports_from_the_other` guard against reintroducing the exact dual-role hazard this PR fixes.

**Nits (non-blocking):**
- `tests/test_label_vocabulary.py::_assigned_names` only walks `ast.Assign`/`ast.FunctionDef` at module top level — an `AnnAssign` or `AugAssign` reintroduction of one of the guarded names would slip past `test_consumer_keeps_no_local_copy`. Not a live risk given the current file contents, just a theoretical gap in the guard's coverage.
- The new module's docstring is a bit dense (three paragraphs of "why" before the code) for a 59-line file, but it's appropriately doing the job of explaining a non-obvious constraint (why two label vocabularies must stay in sync), so I wouldn't cut it.

No correctness bugs or missed edge cases found. This reads as a genuine pure-move refactor, not a disguised behavior change.

PR #1137, opened 2026-07-27
fix(0325): the variables table drops the negation — escape LaTeX and Markdown specials

MinhHaDuong:
**Ticket:** tickets/0325-variables-table-drops-the-negation-latex.erg

The data paper documents how to rebuild the refined subset from the deposited
CSV, and prints that recipe twice. Both sources were correct; neither
published copy was.

| Artifact | Published, before |
|---|---|
| PDF Table 4 (`@tbl-variables`) | `‘df[ df[’is_flagged’] \| df[’is_protected’]]‘` |
| `codebook.md`, in the Zenodo deposit | `df[~df['is_flagged']` |
| §3 prose | `df[~df['is_flagged'] \| df['is_protected']]` — correct |

The table published the **complement**: a reader following it selects the
flagged, unprotected works that were *removed*. The codebook published the
recipe **cut in half at the pipe** — that one was not in the ticket; it is the
same defect in the other markup language, found while checking action 3.

## Root cause

`render_markdown_table()` emits a raw-LaTeX `longtable`, so pandoc never
inspects those cells and the emitter alone is responsible for escaping them.
`esc()` handled `&  %  #  _` and passed the rest through. In LaTeX, `~` is a
non-breaking space and a backtick is an opening quote. `render_codebook()`
writes a Markdown pipe table and left `|` raw, which ends the cell.

## Change

`esc()` → `latex_inline()`: the full special set, applied as a single
`str.translate` pass — a sequential replace loop re-escapes the braces and
backslashes its own earlier substitutions introduce — with backtick spans set
as `\texttt` so a code literal reads as code. `markdown_cell()` escapes the
pipe for the codebook. The LaTeX mapping mirrors pandoc's, so a description
sets identically whichever pipeline carries it.

One finding beyond the ticket: raw LaTeX also bypasses pandoc's smart
punctuation, so the table alone printed `...` where the surrounding prose
prints `…`. Handled, and the rendered test now fails on any further such
divergence rather than letting the table drift typographically.

## Verification

Built the real data-paper PDF. `pdftotext` extraction, Table 4 and §3:

```
df[~df['is_flagged'] | df['is_protected']]
df[~df['is_flagged'] | df['is_protected']]
```

The codebook, rendered through pandoc's GFM reader, now returns the recipe in
one cell. Before/after was produced with the same toolchain in both cases, not
argued from the source.

## Tests

The ticket's own lesson is that a source-level test cannot see this: the `.md`
artifact carried a faithful tilde the whole way to the PDF. So the load-bearing
guard is `tests/test_rendered_table_fidelity.py`, which typesets every contract
description twice in one document — once through the emitter's raw LaTeX, once
as ordinary Markdown down Quarto's own pipeline, the path §3's prose takes —
and compares the extracted PDF text character for character. The oracle is the
sibling pipeline, not a second copy of the escaping table. The codebook gets
the same treatment via a GFM render.

Fast guards in `test_variables_table.py` cover what a source-level test
legitimately can: escape sequences round-trip back to the contract's own
strings, no live special survives into the raw block, unbalanced backticks
fail the build.

## Notes for review

- **`make lint` is green on the merged tree** (166 passed). The two failures at
  the branch's original base — `test_ruff_clean` and
  `test_no_tier3_entry_point_is_imported` — were on files this diff never
  touches and are fixed on current `origin/main`. Seven `test_corpus_acceptance`
  failures are DVC corpus data absent in the worktree, not in the primary
  checkout.
- **Both artifacts were a row short of their own contract on `main`.** 0297
  added `language_provenance` without re-running either emitter, so `make`
  dirtied a clean tree. Both regenerated; table and codebook now agree at 33
  rows.
- **The codebook's `language` missingness moves 4.0% → 0.2%.** I first withheld
  this, reasoning the number came from a corpus rebuild whose `dvc.lock` is
  uncommitted. Review corrected me: `Makefile:288` makes `codebook.md` depend on
  `_deposit_variables.py`, which this PR changes, so withholding withheld
  nothing — it only left a hand-edited generated file for the next `make` to
  overwrite. Regenerated. Both codebook headers report the same 43,179-row
  corpus, so this is 0297's language backfill being measured, not a different
  corpus.
- **`markdown_cell()` escapes prose and code spans separately.** Round 2 showed
  the earlier blanket refusal was overbroad. Checked through real pandoc, the
  three candidate rules fail differently: escaping the pipe alone (what shipped)
  truncates `a \| b` to `a \`; escaping the backslash everywhere renders
  `` `\d{4}` `` as `\\d{4}`; escaping each side by its own rule passes both.
  Pinned as rendered cases — the parametrized GFM test fails against each wrong
  rule, not against an expected escape string.
- **Deposit provenance is now pinned.** `origin/main` gained `6677d12c`
  ("record the corpus regeneration that cleared the language[-null gap]") while
  this PR was open; the branch is merged up to it, so `dvc.lock` now pins
  exactly the `extended_works.csv` the committed codebook was measured from
  (`a4595cd6…`), verified by hash. `codebook.md` reproduces byte-identically.
- **Author call:** the deposited `codebook.md` carries `\|` in its raw text.
  Rendered consumers are correct, but copy-pasting the recipe out of the raw
  `.md` yields invalid Python. Escaping is unavoidable in a pipe table; the
  alternative is rewording the description to avoid `|`. Left as is.
- Incidentally silences a pre-existing `SyntaxWarning: invalid escape sequence
  '\m'` from the docstring at `_deposit_variables.py`.

`make check` on the final tree: 1751 passed, 7 failed — all seven are
`test_corpus_acceptance`, DVC corpus data absent from the worktree, present in
the primary checkout. `make lint` 166 passed, `make check-fast` 1147 passed.

Adherence gate: PASS. Two review rounds, no blocker; round 1's correctness agent
independently reproduced the published defect from `origin/main`'s emitter and
confirmed the new tests fail against it.


MinhHaDuong (commented):
## `/review-pr` — five-agent panel

Panel: correctness, consistency, scope, red team, doc propagation (each a fresh context, `sonnet`). Synthesized verdict: **comment** — no blocker. One agent dissented at `request-changes`; its argument is preserved verbatim below.

The fix itself is sound and independently reproduced. The correctness agent reconstructed the pre-fix `esc()` from `origin/main`, ran it through a real `quarto`/TinyTeX build, and got back the published defect — `‘df[ df[’is_flagged’]` — then confirmed the new tests fail against it. They are regression guards, not change-detectors. The red team rebuilt `tab_variables.md` standalone at A4/2.5 cm through real `xelatex`: zero `Overfull \hbox`, the recipe wraps cleanly at bracket boundaries, so the wide-page probe in the fidelity test is not hiding a column overflow in the real paper.

`make check` on this branch: **1750 passed, 7 failed, 77 skipped**. All seven are `test_corpus_acceptance::TestFileExistence` — DVC corpus absent from this worktree, exactly as the PR body states. `tests/test_rendered_table_fidelity.py` runs and passes here in 4.7 s (quarto + pdftotext + pandoc all present), so the headline guard is live on this machine.

Findings below are the deduplicated survivors. Several agent hypotheses were investigated and dropped; they are recorded at the end so they are not re-litigated.

---

### Findings

**`verifiable:` [doc propagation, consistency] `codebook.md` and `tab_variables.md` now disagree, and nothing detects it.**
```
grep -c '^\\texttt{' deliverables/_shared/tables/tab_variables.md   → 33
grep -c '^| `'       deliverables/_shared/tables/codebook.md        → 32
```
The emitter, run against the current contract, produces 33 rows including `language_provenance`:
```
$ PYTHONPATH=scripts:libs/openalex-corpus/src uv run python -c \
    "from _deposit_variables import render_codebook; \
     md=render_codebook({}, n_rows=1); \
     print(sum(l.startswith('| \`') for l in md.splitlines()))"
33
```
So the committed `codebook.md` is a hand-edited generated artifact that no longer matches its generator. The trade-off is deliberate and well argued in the PR body and in `5b61fbd0`'s message — but nothing durable records it: no comment in `render_codebook()`, no note at `Makefile:288`, no line in ticket 0325, no follow-up ticket. Once the PR merges the reasoning lives only in a PR description nobody re-reads.

Compounding it: `Makefile:288` makes `codebook.md` depend on `scripts/_deposit_variables.py`, which this PR changed. The next plain `make` by anyone unaware of the corpus-lane pinning will rebuild it against `$(EXTENDED)` and publish the 0.2 % missingness number the author deliberately withheld. That is Make behaving correctly on real staleness, not a defect introduced here — but the deliberate half-refresh will not survive contact with the build.

Cheapest durable fix, in this PR: a one-line note above the `codebook.md` rule (or in `render_codebook()`'s docstring) saying the on-disk artifact trails the contract until the corpus-lane regeneration lands, with the ticket reference. A follow-up ticket referenced from 0274 would also do.

**`verifiable:` [doc propagation] Ticket 0274 does not list 0325 as a child.**
`tickets/0325-*.erg`'s own log says "Child of 0274 — resubmission-blocking", but 0274's Children list (0275–0289, 0300, 0301) omits it. `tickets/AGENTS.md`'s tracking-ticket convention requires the tracker to name every child, and 0274's exit criterion is "all children closed" — an unlisted child cannot be checked against that list at integration review.

**`verifiable:` [doc propagation] Ticket 0325's Verification and Exit-criteria boxes are all unticked** (`tickets/0325-...erg:80-94`), while the PR body claims concrete evidence for each. Tick them with the evidence before merge.

**`consider:` [red team] `markdown_cell()` has a blind spot for a backslash already adjacent to a pipe** (`scripts/_deposit_variables.py:243`). A source value containing `\|` becomes `\\|`, which GFM reads as a literal backslash followed by a live cell separator — content silently dropped, reproduced end-to-end through `pandoc -f gfm`. `tests/test_variables_table.py:312` (`test_every_row_has_exactly_five_cells`) splits on `(?<!\\)\|` and shares the same blind spot, so it passes on the input that corrupts the render.

Triaged as `consider:`, not a blocker, because it is unreachable from the contract today — no description or type contains a backslash:
```
$ ... print([v.name for v in DEPOSIT_VARIABLES if '\\' in v.description or '\\' in v.type])
[]
```
It is nevertheless the ticket's own defect class (a delimiter in payload text, silently truncating a published recipe) surviving in the function written to close it. `re.sub(r'\|', r'\\|', ...)` on an already-escape-aware pass, or an assert forbidding backslashes in contract strings, would close it in two lines.

**`consider:` [consistency] `render_codebook()` still uses raw `v.description` while `render_markdown_table()` now uses `describe(v)`** (`scripts/_deposit_variables.py:401` vs `:333`). Verified: the emitter's codebook output carries the "(absent from corpus builds predating this pipeline stage)" suffix on **0** rows; `tab_variables.md` carries it on several. Pre-existing on `main` and not a regression — but this PR introduced `describe()` precisely to give the two markup languages one source of description text, and then applied it to one of the two. Applying it to both is a one-word change; deciding *not* to (the codebook has a missingness column that may make the suffix redundant) deserves a comment.

**`consider:` [correctness, red team, agreed] The LaTeX map has no entry for `"`.** A straight double quote in a future prose description would set as same-direction quotes in the raw-LaTeX copy against pandoc's curly pair in the Markdown copy. No contract value contains one today (verified), and `test_rendered_table_fidelity.py::test_every_description_renders_as_written` would fail loudly if one arrived — which is the docstring's stated design. Recording it so the next reader knows the boundary was chosen, not missed.

**`consider:` [scope] The `...` → `\ldots{}` substitution puts smart-punctuation emulation inside an escaping function.** The scope agent's read: it is load-bearing for the ticket's Action 4 (character-identical copies), the docstring at `:222-224` explicitly scopes it to what the contract exercises, and the fidelity test converts any future divergence into a loud failure rather than silent drift. The alternative not considered aloud — disabling smart punctuation, or forbidding `...` in contract strings with a validation assert — would keep the layer clean. Worth a conscious "yes, this layer" from the author rather than an implicit one.

**`nofollow:` [red team] `test_rendered_table_fidelity.py` skips rather than fails when quarto/pdftotext are absent or the render fails.** True, and there is no CI at all in this repo (`.github/workflows/` does not exist), so enforcement rests entirely on whoever runs `make check` having the toolchain. Recorded, not actioned: it is a repo-wide condition this PR did not create, and demanding a hard failure would make the suite unrunnable on a machine without TinyTeX.

**`nofollow:` [correctness] `test_no_live_special_reaches_the_pdf`'s lookbehind `(?<!\\text)~` is dead** — the preceding `.replace(r"\textasciitilde{}", "")` already removed every match it could exclude, so it is equivalent to a bare `~` search. Cosmetic; the assertion loses no strength.

**`nofollow:` [correctness] `plain_text()`'s `\\texttt\{([^}]*)\}` mis-decodes a code span containing a literal brace** (reproduced). Test tooling only, would surface as a false failure on legitimate future content rather than a missed defect.

**`nofollow:` [scope] The `language_provenance` row in `tab_variables.md` belongs to ticket 0297, not 0325.** Isolated in its own commit (`06f186af`) with a stated rationale — the stale artifact dirtied a clean tree and `make` would have produced it regardless. Acceptable as-is; a stricter reviewer could ask for a separate housekeeping PR.

**`nofollow:` [scope] The `r"""` raw-docstring change silencing a pre-existing `SyntaxWarning: invalid escape sequence '\m'` rides in `074a3c53` unmentioned in that commit's message.** Trivial and safe, on a docstring the commit was already editing; noted only because "one change per commit, message explains why" is the standing rule.

**`nofollow:` [doc propagation] Two pre-existing stale counts in R&R response material,** neither touched by this diff, both reusable verbatim in the resubmission: `deliverables/data-paper/revision-rdj26561/r1-19-csv-structure.md:7` says "26 columns" against a 33-variable contract; `ed03-variables-table.md:8-10` enumerates four provenance columns and omits `language_provenance`. Out of scope here — flagged because the resubmission will carry them.

---

### Dissent, preserved

The red team returned **request-changes**, on the strength of the `markdown_cell` backslash truncation and its matching test blind spot: *"a genuine, demonstrated test-evasion of every new guard in this PR."* The synthesis downgraded it to `consider:` on reachability — no contract string contains a backslash, so no published artifact is affected today. That downgrade is a judgment about exposure, not about the defect: the red team's reproduction stands, and the author may reasonably prefer to close a two-line hole in the function whose whole purpose is to stop exactly this failure mode.

### Hypotheses investigated and dropped

Recorded so the next panel does not spend on them again. Apostrophes, `--` and `---` in prose are ligature-substituted identically by both paths (the render engine's `Mapping=tex-text` matches pandoc's smart typography) — the two copies agree, verified by build. `\texttt{}` renders braces, tilde, caret, `$` and non-ASCII correctly in code spans, verified through real `quarto render` + `pdftotext`. `\|` inside a GFM code span round-trips correctly, verified with `pandoc -f gfm -t html`. `\texttt{10--20}` renders as a literal `10--20`, not an en-dash. The longtable column does not overflow at realistic width. `export_deposit.py`'s `transform()` / `check_columns()` are untouched and still agree with the contract, so the ticket's deposit-contract invariant holds. `uv run ruff check` is clean on all four changed files; line length is not gated (`E501` absent from the project's `select`). Test tiering is correct — the render test is properly `integration`, excluded from `check-fast` and from the duration ratchet. `build/build_datapaper_archive.sh` and `build/templates/README-datapaper.md` only `cp` `codebook.md` as an opaque file, so the new `\|` cannot break a downstream parser. `data-paper.qmd` hardcodes no variable count. No stale "pipe table" phrasing survives outside the one the PR fixed.

---

Nothing here blocks the merge. The three `verifiable:` items are minutes of work and two of them (the tracker link, the ticket checkboxes) are hygiene the merge gate will want anyway; the codebook-divergence note is the one that earns its keep, because the reasoning is currently one merge away from being lost.

MinhHaDuong (commented):
## `/review-pr` — round 2, focused panel

Round 1's five-agent panel covered `c3a46845..06f186af`. One commit landed since — `e30015a5`, written in response to that review — so this round targets the delta and verifies the round-1 findings actually closed. Four agents (correctness, data provenance, red team, scope+consistency), each a fresh `sonnet` context. Synthesized verdict: **comment** — no blocker.

`make check-fast` on the branch tip: **1146 passed, 11 skipped** (21.6 s). The PR's own suites, run under the project path roots: **36 passed, 1 skipped** — the skip is `TestAgainstShippedCorpus::test_shipped_csv_matches_contract`, gated on the DVC corpus absent from this worktree. `ruff check` clean on all changed files. Test tiering verified by collection: 32 of 37 run in the fast tier; the 5 deselected are the three subprocess-spawning fidelity tests (correctly `integration`) plus the corpus test (`slow`).

### Round-1 findings — closed

- **codebook/table row disagreement (33 vs 32) — resolved, and better than a note.** The recommended fix was a comment recording the deliberate half-refresh; `e30015a5` instead regenerated. Verified: with the real corpus symlinked in, `export_codebook.py` reproduces the committed `codebook.md` **byte-identically** (`diff` exit 0, 43,179 rows measured). Both artifacts now carry 33 rows. This is a genuine generator output, not a hand-edited one.
- **The red team's `request-changes` dissent (`\|` → `\\|` GFM truncation) — answered.** `markdown_cell()` now raises. The new `test_backslash_is_a_build_error` discriminates: simulated against the pre-fix function it returns `'a \\\\| b'` instead of raising, so it fails red on the old code.
- **Tracker 0274 now lists 0325** (`tickets/0274-...erg:55`); **0325's Verification and Exit-criteria boxes ticked** and each evidenced by a green test.
- **`describe()` vs raw `v.description` in `render_codebook()` — re-examined, downgraded to `nofollow:`.** Not an information loss. Of the 8 optional variables, six carry real measured percentages in this build (the "absent from corpus builds predating this pipeline stage" caveat would not apply to them), and the two genuinely absent — `semantic_outlier_dist`, `in_v1` — already show `Missing: n/a`, which conveys the same fact per-build rather than as boilerplate. The table needs the static caveat because it has no missingness column; the codebook does not.

### New findings

**`verifiable:` [correctness + provenance, found independently by two agents] The branch tip's `dvc.lock` does not describe the corpus that produced the committed codebook — but `origin/main` already does, and this branch never touches the file.**

```
HEAD:dvc.lock        extended_works.csv  md5 ceff029c269749b4c97701bc7b7d567b  84921207
origin/main:dvc.lock extended_works.csv  md5 a4595cd619138d31231d260ec42dce8b  85412413
```
The committed `codebook.md` was verified to reproduce byte-identically from the corpus whose real hash is `a4595cd6…` — i.e. `origin/main`'s state, locked by `6677d12c chore(0297): record the corpus regeneration that cleared the language nulls`. On the branch tip in isolation, `dvc pull && make codebook` would restore `ceff029c…` and rebuild the old 4.0 % / 32-row codebook, silently reverting `e30015a5`.

`git diff --name-only origin/main...HEAD` confirms `dvc.lock` is not among the eight files this branch touches, so the merged tree takes `origin/main`'s lock and is self-consistent. `mergeStateStatus` is CLEAN. **No action required beyond the standard pre-merge currency check** — but do not merge against a base that predates `6677d12c`.

Residual, worth correcting while it is still cheap: `e30015a5`'s message justifies the number with "same 43,179-row corpus in both headers". A backfill fills nulls without changing the row count, so matching row counts are *consistent with* one corpus, not proof of it. The accurate statement is the hash chain above. This is the provenance trail of a Zenodo-deposited artifact, so the history is worth getting right.

**`verifiable:` [red team] The new backslash guard is overbroad — it refuses input the renderer provably handles.**

`scripts/_deposit_variables.py:248` raises on any backslash in the raw cell text, before backtick parsing, so it cannot distinguish the dangerous case (`\|`, which re-escapes to `\\|` and truncates) from the safe one. Reproduced against the real GFM renderer:

```
$ printf '| a | b |\n|---|---|\n| `C:\\Users\\name` | text |\n' | pandoc -f gfm -t html
…<td><code>C:\Users\name</code></td>…
```
A backslash inside a code span round-trips exactly as written. Yet:
```
>>> markdown_cell(r'a regex like `\d{4}`')
ValueError: backslash has no safe pipe-table escaping: 'a regex like `\\d{4}`'
```
Consequence: no future contract description can document a regex, a Windows path, or a LaTeX command by name — the build stops permanently on input that was never unsafe. It fails safe (raises rather than corrupts) and nothing in the contract triggers it today, which is why this is not a blocker. But it is an author call rather than an obvious win: is "zero backslashes, ever" the intended permanent policy, or should the guard target the `\|` adjacency? The docstring argues no single rule is right on both sides of a code-span boundary — that argument is sound for *escaping*, and does not by itself justify refusing the cases where no escaping is needed.

**`consider:` [red team] `latex_inline()` silently misrenders the CommonMark double-backtick idiom.**
```
>>> latex_inline('use `` `nested` `` spans')
'use \\texttt{} \\texttt{nested} \\texttt{} spans'
```
The intended single code span reading `` `nested` `` splits into three alternating segments. The `unbalanced backtick` guard does not catch it — four backticks give five parts, an odd count, so the parity check passes. Unreachable today (no contract string contains `` `` ``), and the rendered-fidelity test would catch it loudly if one arrived, since the Markdown copy would set it correctly. Recorded so the next panel does not rediscover it.

**`consider:` [scope] Tracker entries for 0320 and 0321 ride along in a LaTeX-escaping PR.**
`tickets/0274-...erg:53-55` gains three children, only one of which (0325) is this PR's. 0274's own 08:11Z log already named 0320 and 0321 as filed, so the Children list was simply out of sync and this is syncing rather than inventing — the same defect class round 1 flagged for 0325 itself. Three mechanical lines in a bookkeeping file; noted only because one-change-per-commit is the standing rule.

### Attacks tried and refuted — add to the dropped list

All reproduced through the real toolchain on this machine (`quarto render` → LaTeX → `pdftotext -layout`, `pandoc -f gfm`), not source reasoning:

- **`&` inside a `\texttt{}` code span, and bare `&` in prose, do not break the longtable's column structure.** Injected `` "See `a|b~c&d` and profit & loss and `x&y`" `` as a description and built it: extracted text `See a|b~c&d and profit & loss and x&y`, row stayed a clean three columns. This was the most promising remaining structural attack; it fails because a raw `&` is always translated to `\&`, so no bare `" & "` substring ever reaches the row.
- `\textbar{}` and `\textasciitilde{}` inside `\texttt{}` render as literal `|` and `~` (the round-1 list covered braces/tilde/caret/`$`/non-ASCII in `\texttt{}`, but not the pipe).
- A single embedded newline in a description does not break the `\\` row terminator; a double newline (paragraph break) inside the `p{8.2cm}` cell builds clean and stacks two lines within the same cell — a content-fidelity nuisance at most, not structural corruption.
- The test helper's `table_rows()` / `.split(" & ")` parsing is not evadable by escaped `&`/`|`/`~` in a cell — verified on the injected payload above, still exactly three cells.
- `TestCodebookEscaping`'s `re.split(r"(?<!\\)\|", …)` cell counting no longer has the round-1 blind spot: the only input that could produce an ambiguous `\\|` run is now unreachable past the `markdown_cell` guard.
- Ticket 0297 corroborated as the cause of the missingness move (`tickets/closed/0297-corpus-language-null-rate-regression.erg`: opened at a 4.1 % null rate, action "backfill via language detection or the source API", closed by PR #1129) — consistent with 4.0 % → 0.2 % and with the new `language_provenance` allowed values.
- Full-file diff of `codebook.md` against `origin/main`: exactly three changed lines — the `language` percentage, the new `language_provenance` row, and this PR's own `\|` escape. The 43,179-row header line is byte-identical. Nothing else moved, which is what the "0297's backfill, not a different corpus" story predicts.
- No new stale downstream figures. The two pre-existing ones round 1 flagged (`r1-19-csv-structure.md:7` "26 columns"; `ed03-variables-table.md:8-10` omitting `language_provenance`) are confirmed *worse* against a 33-variable contract, still out of scope here, and still headed for the resubmission. `STATE.md:15` is stale on 0297 (says "fixable to ~0.4 %", actual 0.2 %, ticket now closed) — outside this diff.
- Ticket-close hygiene: exactly one `**Ticket:**` line (0325). 0274 is not named in closing form and correctly stays open. `erg check tickets/` passes.

### Disposition

Nothing blocks. The one item that would change the merged artifact is not in this diff at all — merge onto a base at or after `6677d12c`, which `origin/main` already is. The backslash guard's breadth is the only genuine judgment call left, and it is a policy question for the author, not a defect: it fails loudly rather than publishing a truncated cell, which is the direction this ticket exists to enforce.

PR #1141, opened 2026-07-27
feat(0323): narrate langdetect-inferred language tags in the data paper

MinhHaDuong:
**Ticket:** tickets/0323-data-paper-narrate-langdetect-inferred-l.erg

**Prose PR — for the author to arbitrate, not to auto-merge.**

## The sentence added (§2.4, verbatim)

New paragraph, placed immediately after the abstract-flaws paragraph that is its precedent:

> Language codes are often missing or non-standard: enrichment backfills them from the OpenAlex API, then infers what remains with langdetect from title and abstract, and `language_provenance` marks each tag's origin. The inference covers {{< meta lang_detected_n >}} refined works ({{< meta lang_detected_pct >}}%), so a user who needs catalog values alone can drop them.

Renders as: *"The inference covers 1,331 refined works (4.0%), so a user who needs catalog values alone can drop them."*

Shape mirrors the precedent (`data-paper.qmd`, abstract paragraph): flaw, then pipeline response, then the disclosure column. Wording keeps the tags from reading as authoritative — "infers", "the inference covers", and a named escape hatch, rather than a bare share that reads like a measurement.

## Vars added

All in `scripts/analysis/compute_vars.py::corpus_stats` (measured on `refined_works.csv`), and registered in `DOC_VARS["data-paper"]`:

| Var | Value | Meaning |
|---|---|---|
| `lang_detected_n` | 1,331 | works whose `language_provenance` starts with `detected` |
| `lang_detected_pct` | 4.0 | same, as a share of refined works |
| `lang_non_english_n` | 2,061 | non-English layer, excluding unclassified (replaces a hand-typed literal, below) |

Prefix match rather than equality on `detected:langdetect`: the value names the detector, so a second detector would extend it rather than replace it.

## Denominator chosen, and why

**The refined corpus (33,344), not the enriched one (43,179).** Everything the clause sits next to is a refined-corpus figure: `lang_english_pct` comes from `load_refined_works()`, and `export_language_table.py` is fed `$(REFINED)` by the Makefile and captions itself *"Language distribution in the refined corpus"*. Quoting the enriched count (1,620 / 43,179 = 3.8%) would mix denominators inside one section. The enriched-corpus provenance distribution stays in `codebook.md`, which describes the deposited file — and which this PR regenerates, because it had drifted and did not yet carry the row (review round 1).

## `lang_english_pct` movement — flagged

It moves, and not for the reason the ticket anticipated.

`compute_vars` computes `lang_english_pct` from `load_refined_works()`, which reads `refined_works.feather` **whenever it exists** — no staleness check. After the 2026-07-27 rebuild that cache was three days old (2026-07-24). It had **no `language_provenance` column at all**, and still carried pre-0297 `language` values:

| | stale cache | current CSV |
|---|---|---|
| `lang_english_pct` | 89.86 | **93.79** |
| non-English layer | 2,025 | **2,061** |
| unclassified | 1,356 | **9** |
| rows whose `language` differs | — | **1,350** |

So the corpus rebuild was not reaching the prose at all: the cache was swallowing it downstream of every existing guard. `lang_detected_n` was simply uncomputable through the mandated loader path. Fixed in `scripts/pipeline_loaders.py` (`_handoff_cache_is_current`): a Feather cache older than its CSV is no longer served, and the loader logs the `make corpus-handoff` remedy. Both handoff caches are covered — they come off one Makefile pattern. mtime comparison, not a hash or schema check, because the Makefile already expresses the dependency that way; the loader now enforces the contract the build declares instead of inventing a second one. An absent CSV is not staleness (archives may ship either file).

Consequence for the author: **`lang_english_pct` renders 93.8, where a Phase-2 run before this PR rendered 89.9.** 93.8 is the correct figure for the frozen corpus.

The corpus was **not** touched — no `make corpus*`, no `dvc repro`, nothing written under `data/catalogs/`. `refined_works.feather` is left stale on disk on purpose; the next Phase-2 build refreshes it through the existing Make rule.

## Second prose change, same invariant

> ~~the non-English layer counts 3,381 works~~ → the non-English layer counts {{< meta lang_non_english_n >}} works

`3,381` was an **enriched**-corpus figure typed into a **refined**-corpus sentence, and the rebuild moved the refined layer to **2,061** without moving the prose — a 64% overstatement live in §2.3. Not droppable collateral: the new count guard fails while any such literal remains. The ordering claim that follows it ("led by Portuguese, Spanish, German, and French") still holds — pt 418, es 308, de 295, fr 256.

## Test

Per the ticket's Test section, the §2–§3 literal guard in `tests/test_citation_verification.py` is extended — and it needed widening, not just new entries. The existing guard sees **percentages only**, which is why a count rotted under it. `compute_vars._int` formats every corpus count as `f"{n:,}"`, so a hand-typed count is lexically identical to a generated one.

- `test_method_and_data_sections_carry_no_hand_typed_count` — comma-grouped thousands plus bare 4+-digit integers, after stripping years, DOIs, URLs, and inline code. Red on `origin/main` with `['3,381']`; green here. Documented limit: a count below 1,000 typed without a separator still escapes. A wider net would allowlist every year and section number and break on each date bump.
- `tests/test_loader_cache_freshness.py` (4 tests) — the stale cache is not served; a **current** cache still is (a correctness fix that disables the speed layer is a regression, not a fix); a cache with no CSV beside it is served; the citations cache is covered too.

- `tests/test_language_table.py` (+2) — `arz`/`sco` bucket as unknown, and the prose path and the table path partition an identical code list identically (review round 1).

Gates: `make check-fast` 1144 passed · `make lint` 166 passed, 0 failed (`test_ruff_clean` and `test_no_tier3_entry_point_is_imported` both green here) · `/verify-adherence` PASS.

## Action 3 — the response letter

Checked, **not edited** (the ticket's scope is prose + vars; the letter is a submission record).

**R1-13 is the wrong home.** It answers citation-link accuracy, reference counts, and GROBID — nothing about metadata-field provenance. The natural homes are:

- **R1-11** (folding R1-07/R1-10, *OpenAlex limitations and pipeline responses*) — its answer says §2.4 gives "a clear statement of each OpenAlex limitation, and the pipeline step that addresses it". This PR adds exactly one such pair, so the claim is now stronger and the answer can name it.
- **R1-16c** (*non-English coverage wording*) — the remark where the referee pushed back on overstated language coverage. The layer count in that discussion has changed (3,381 → 2,061), and the tags behind it are now disclosed as partly inferred.

**Separate finding in R1-13, worth a look before the letter goes out.** Its closing sentence still reads *"the existing 300-link audit against Crossref (99.0% confirmed, 95% CI [97.1%, 99.7%])"*. That is the pre-0320 figure — the point estimate that fell outside its own interval. `tab_citation_verification.csv` and the paper both now say **97.0%, 95% CI [94.4%, 98.4%]**. The reference-count figures in the same paragraph (median 29, mean 42.8, p95 126, max 1,536, 20.3% / 5,159 works) all still match the artifact. Already owned by open ticket **0283** (logged 2026-07-27, all three files), so nothing filed here.

---

**Review round 1 addressed** — see [the reply comment](https://github.com/MinhHaDuong/climate-finance-het/pull/1141#issuecomment-5091295633): the prose/table count disagreement fixed at the cause (one shared normaliser, `lang_non_english_n` 2,063 → **2,061**), `reproducibility.md`'s stale loader contract corrected, `codebook.md` regenerated, and the scope dissent answered.


MinhHaDuong (commented):

Substantial PR: prose + vars + a behavior change to a loader every Phase 2 script uses. All five agents ran, each pinned below the coder tier. Verdicts: **4 × comment, 1 × request-changes (scope)**. The dissent is preserved verbatim below — the author arbitrates.

Every numeric claim in the PR body was independently recomputed against `data/catalogs/refined_works.csv` (33,344 rows) and **all check out**: `lang_detected_n` 1,331 / 4.0%, `lang_non_english_n` 2,063, `lang_english_pct` 93.79, pt/es/de/fr 418/308/295/256, and the on-disk `refined_works.feather` confirmed to lack `language_provenance` entirely. The red-on-`origin/main` claim for the new count guard reproduced exactly (`['3,381']`).

### Gates

`make check` → **1742 passed, 8 failed, 77 skipped** (285 s). All 8 failures are environmental and untouched by this diff:
- 7 × `tests/test_corpus_acceptance.py::TestFileExistence/TestCorpusAlign` — `data/catalogs/` in this worktree holds only `.dvc` pointers.
- 1 × `tests/test_bias_flag.py::test_bias_flag_produces_different_output` — `LLVM ERROR: Symbol not found: __gufunc__..._dcor...` from the dcor/numba stack, torch absent in this venv.

Fast tier 1142 passed · adherence 166 passed · `ruff check` clean on all four touched Python files · `tests/test_doc_vars_completeness.py` 6/6 (it pins DOC_VARS ↔ `{{< meta >}}` bidirectionally, so a missing var aborts the build rather than shipping an empty macro — the panel's highest-priority attack, closed).

---

### Worth fixing in this PR

**1. `verifiable:` — `lang_non_english_n` and `@tbl-languages` disagree by 2 rows, and the code comment claims they don't.** [correctness, doc propagation, independently re-measured]

`scripts/analysis/compute_vars.py:254-255` classifies with `~lang.str.lower().isin(["en","english","unknown"])`. `scripts/figures/export_language_table.py` routes through `normalise_language`/`normalize_lang`, which buckets any code it cannot map to 2-letter ISO 639-1 as unknown. Two rows carry raw `arz` (Egyptian Arabic) and `sco` (Scots):

```
$ python -c "...refined_works.csv, usecols=['language']"
compute_vars non_en = 2063
arz    1
sco    1
```

The table therefore puts them in **Unclassified = 11**, giving a non-English total of **2,061**, while the prose prints **2,063**. The comment "matching the separate 'Unclassified' line of @tbl-languages" is off by two. Immaterial to the rounded headline figures, referee-visible as a printed count that does not sum to the table beside it. Fix either way: reuse `normalise_language` for classification, or drop the correspondence claim from the comment.

*Dissent noted:* red team argued no divergence is possible because `enrich_join.py:160` pre-normalizes every `language` value. The measurement above contradicts that — `arz`/`sco` reach the CSV raw.

**2. `verifiable:` — a published deliverable still documents the loader contract this PR replaces.** [doc propagation]

`deliverables/_shared/_includes/reproducibility.md:84`:

> The Phase 2 loaders (`pipeline_loaders.py`) prefer Feather when present and fall back to CSV transparently.

That include renders into `deliverables/corpus-report/corpus-report.qmd`. "When present" is exactly the semantics `_handoff_cache_is_current` removes. One clause ("…when present and current; a cache older than its CSV is treated as stale") keeps the deliverable honest, and it sits squarely in this diff's blast radius.

---

### Flagged, not for this PR

**3. `verifiable:` — `codebook.md` is stale, has no `language_provenance` row at all, and the response letter promises the referee it cannot be.** [doc propagation, consistency — the panel's highest-value finding]

`deliverables/_shared/tables/codebook.md` was last regenerated at `c8cbbdaf` (2026-07-24), predating 0297's addition of `language_provenance` to `scripts/_deposit_variables.py` (`a0f05ab1`, 2026-07-27). Regenerating from the current `extended_works.csv`:

```diff
- | `language` | ... | 4.0% | ...
+ | `language` | ... | 0.2% | ...
+ | `language_provenance` | string, nullable | source, openalex, detected:langdetect, empty | 0.2% | ...
```

Two consequences. (i) This PR's denominator argument states "the enriched-corpus provenance distribution stays in `codebook.md`" — it is not there. (ii) `response-letter.md` R1-19 tells the referee the codebook "is generated from the same declared column contract… so it cannot drift from the data." Presently untrue, and nothing enforces it — `test_datapaper_archive_layout.py` checks the file exists, not that its content is current against `_deposit_variables.py`. Cheap to fix (one Makefile target) and worth doing before the letter is sent; a content-currency test would close the class.

**4. `verifiable:` — R1-13's stale audit figure already has an owner.** [scope]

`deliverables/data-paper/revision-rdj26561/response-letter.md:154` reads "99.0% confirmed, 95% CI [97.1%, 99.7%]"; `tab_citation_verification.csv` gives 97.0, CI [94.3977, 98.4138]. The PR body flags this and declines to fix it — correctly: `tickets/0283-datapaper-response-letter-resubmission.erg` already logs it at `2026-07-27T09:52Z`, naming all three affected files (`response-letter.md`, `summary-of-revisions.md`, `citation-quality-response.md`), and 0283 is open. **No new ticket needed.** R1-16c does not cite the 3,381/2,063 count, so this PR's prose change contradicts nothing in the letter.

**5. `consider:` — the count guard has two measured gaps beyond the one it discloses.** [red team]

Run against a scratch qmd, `_measurable_counts` returns `[]` for both `33 344` (space-grouped) and `1.5 million` — escapes the docstring does not mention. In the other direction, on isolated strings: `0000-0002-1825-0097` → `['0000','0002','1825','0097']`, an ISSN → `['1234','5678']`, a page range `3381--3390` → `['3381','3390']`. None of these shapes is live in §2-§3 today (verified by reading the section; the two Zenodo DOIs, both `1990--2024` year ranges, and the allowlisted `8192` all strip correctly). Incremental hardening of a net that already catches the regression it was built for — not a blocker.

**6. `consider:` — `_handoff_cache_is_current` has an unguarded false-freshness direction.** [correctness, contra red team]

`os.utime`-ing a CSV *older* than its Feather (as `rsync`/`tar` preserving historical mtimes, or a rollback, would) makes the function return `True` and the loader serve a possibly-divergent cache silently. None of the four new tests exercises that direction. Red team's counter-evidence: `.dvc/config` sets no `cache.type`, so DVC's default `copy` stamps checkout-time mtimes and can only make the CSV look *newer* — no false-fresh path in the real workflow, reachable only by an explicit `touch` without a rebuild. I side with red team on likelihood; the residual is operator error, and `>=` at the boundary correctly mirrors Make's own up-to-date semantics.

**7. `consider:` — TOCTOU between `os.path.exists(FEATHER)` and `os.path.getmtime`.** [red team] A concurrent delete/rebuild in that window raises an uncaught `OSError` instead of falling back to the CSV. Narrow; Make runs these sequentially. Not reproduced.

**8. `consider:` — `STATE.md:15-16` still poses 0297's timing as an open author call** ("do it before the letter or after resubmission, not between"), but 0297 is closed and this PR documents the rebuild as already done, `lang_english_pct` already moved. Next housekeeping pass should close the loop — including whether that constraint was honored. [doc propagation]

**9. `nofollow:` — `_handoff_cache_is_current`'s docstring is an 18-line postmortem** in a file whose other docstrings run 2–5 lines of contract, and the same incident rationale (89.86 vs 93.79, "three days old") is triplicated into the `compute_vars.py` comment and the test module docstring. Valuable content, three places to update in sync. [consistency]

**10. `nofollow:` — `export_language_table.py`'s docstring and default `ENRICHED_PATH` still say "enriched corpus"** while `Makefile:278-279` overrides with `--input $(REFINED)`. Latent trap for a future direct invocation; pre-existing, untouched by this diff. [red team]

---

### Dissent: scope says request-changes

**scope agent, verbatim:**

> `$(REFINED_FTH): $(REFINED)` already exists in the Makefile (`Makefile:222-223`), rebuilding the Feather from the CSV whenever the CSV is newer, via `make corpus-handoff`. Running that one command — with **zero code change** — would have refreshed the stale cache and made `lang_detected_n`/`lang_english_pct` compute correctly through the *unmodified* loader. The PR's own warning message even names this as the remedy. So the ticket's number was not "literally uncomputable" — it was one Make invocation away. That reframes this from "(a) blocking prerequisite" to "(b) a valuable but separable hardening of the loader," which is exactly the pattern flagged in project memory `feedback_scope_discipline`.

**red team, contra, verbatim:**

> `Makefile`'s `$(COMPUTED_STATS)` grouped target (line 305, builds `compute_vars.py`'s output) depends only on `$(REFINED)` (the CSV) — **not** on `$(REFINED_FTH)`/`corpus-handoff`. So plain `make analysis-stats`/`make stats` never forces a Feather rebuild before `compute_vars.py` runs; a stale Feather left from an earlier build is exactly what this fix intercepts, independent of whether a script was invoked "directly."

I verified the crux myself: `Makefile:305` lists `$(REFINED)` and never `$(REFINED_FTH)`; `corpus-handoff` is documented as **optional** at `Makefile:17` and again at `reproducibility.md:84`. So `make corpus-handoff` fixes today's number and leaves the defect armed for the next rebuild — the loader change is a structural fix for a dependency no rule owns, which makes the bundling defensible. Against that: it is a semantics change to a loader read by ~25 Phase 2 scripts, riding a prose PR the author is asked to arbitrate, and the commits being split (`test` / `fix` / `feat`) is only partial mitigation.

**A third option, if the author wants the separation without losing the fix:** add `corpus-handoff` as a prerequisite of `$(COMPUTED_STATS)` — one Makefile line, and arguably the fix the dependency graph is actually missing. The loader guard then becomes belt-and-braces rather than the sole owner of the invariant, and can travel in its own ticket.

**11. `consider:` — tracker `0274` does not list `0323` among its children** (nor 0320/0321, filed the same day), though 0323's log says "child of 0274." Pre-existing; per ticket convention the tracker should enumerate every child. [scope]

**12. `consider:` — one register nit in the new §2.4 sentence.** "so a user who needs catalog values alone can drop them" is the only clause in §2.4 that talks about what a reader might *do* with a column rather than what the pipeline did to it; the section otherwise addresses the audience once, formally ("Users should be aware…"). Not an AI tell, and the sentence otherwise mirrors the abstract-flaws precedent exactly — same colon-led flaw → response → disclosure skeleton, same mid-prose backticked column name (`near_duplicate_group` sets that precedent in the same section). Author's call. [consistency]

**Cleared without findings:** var naming and `DOC_VARS` sort position; all three vars consumed exactly once, none orphaned; `low_memory=False` is parity with `load_refined_citations`, which already had it on `origin/main`; test tiering correct (`tests/_tier_autoscan.py` does not count pandas as heavy); `_log.warning` genuinely surfaces (reproduced with and without `utils` imported); `.str.startswith("detected")` safe on the all-NaN dtype path and correctly forward-compatible with a second `detected:*` detector; no stale `3,381`/`89.8x` literal anywhere in the tree; the `3,381` → var swap is load-bearing (the new guard does flag it); `**Ticket:**` names only 0323, correctly leaving tracker 0274 open.

🤖 `/review-pr` — 5 agents, `model: sonnet`

MinhHaDuong (commented):
## `/review-pr` — five-perspective panel (round 2)

Correctness · Consistency · Scope · Red team · Doc propagation, each in a fresh context. Substantial-risk classification: the PR carries a prose change *and* a repo-wide Phase-2 data-loading semantics change.

**Panel verdicts — dissent preserved verbatim, the author arbitrates:**

| Perspective | Verdict |
|---|---|
| Correctness | approve |
| Consistency | comment |
| **Scope** | **request-changes** |
| Red team | comment |
| Doc propagation | approve |

`make check`: **1745 passed, 7 failed, 77 skipped** in 219 s. All 7 failures are `test_corpus_acceptance.py::TestFileExistence` / `TestCorpusAlign` asserting `data/catalogs/*.csv` exists — this review worktree has an empty `data/catalogs/` (only `.dvc` pointers; `data/` is not in `.worktreeinclude`). Unrelated to the diff, which touches no acceptance test. Every test the PR adds or modifies is green: 23 passed on `test_loader_cache_freshness.py` + `test_language_table.py` + `test_citation_verification.py`, and `test_doc_vars_completeness.py` green with the three new vars.

---

### One item to resolve before merge (Scope, high confidence)

**Ticket `0315-feather-cache-mtime-invalidation.erg` is open and already owns this PR's loader fix — and is referenced nowhere.**

Created 2026-07-24, three days before 0323, from the corpus-v2 rebuild incident ("bit twice on 2026-07-24"). Its Actions and Exit criteria describe `_handoff_cache_is_current` almost verbatim:

> 1. In pipeline_loaders, when a Feather sidecar is older than its source CSV, ignore it…
> 2. Cover load_refined_works / load_refined_citations / any loader with a Feather fast path.
> - [ ] A rebuilt CSV can never be shadowed by an older Feather sidecar

`scripts/pipeline_loaders.py:301-333` (commit `be59f505`) satisfies all of it — both Feather-reading loaders are covered, and the panel independently confirmed only two exist (`grep read_feather scripts/` matches `pipeline_loaders.py` alone; `load_refined_embeddings`/`load_embeddings` read `.npz`). Its Test section ("touch the CSV newer than the Feather; loader returns CSV content") is `tests/test_loader_cache_freshness.py::test_a_cache_older_than_the_corpus_is_not_served`.

But `grep -rn 0315` over `tickets/`, `scripts/`, `deliverables/`, `STATE.md` returns nothing outside the ticket file. Not in the PR body, not in any of the six commits, not in 0323's body, and not in round 1 — which debated at length *whether* this fix belongs here without either side knowing a ticket already claimed it. `erg-pr-merge` closes only the tickets named in `**Ticket:**` lines, so 0315 survives the merge describing finished work, and the next picker takes it.

This reframes the round-1 scope dispute rather than reopening it: the fix is *ticketed work*, not unticketed creep, which strengthens the rebuttal. Remedy is cheap — add a second `**Ticket:** tickets/0315-feather-cache-mtime-invalidation.erg` line (the workflow rule permits multiple lines "when they genuinely land together", which one commit here does) and log the landing on 0315.

---

### Substantive, for the author's arbitration (Red team, high confidence)

**`verifiable:` The new sentence's escape hatch is imprecise about `openalex`.** `deliverables/data-paper/data-paper.qmd:124` promises "a user who needs catalog values alone can drop them" — them being the `detected:langdetect` rows. But this PR's own `codebook.md` draws the line elsewhere:

> `source` (carried by the source catalog), `openalex` (**backfilled from the OpenAlex API**), `detected:langdetect` (inferred…)

Dropping only the detected rows leaves the `openalex` backfill (850 works enriched-corpus) inside the set the reader believes is catalog-only — third-party API language calls, unaudited, which the codebook explicitly declines to call "carried by the source catalog". The sentence's own preceding clause names the backfill correctly ("enrichment backfills them from the OpenAlex API, then infers what remains"), so the imprecision is only in the closing promise. Two readings are defensible — OpenAlex *is* a catalog, so "catalog values" may be intended broadly — which is exactly why this is the author's call, not a fix to apply. `language_provenance` is genuinely in the deposit contract (`scripts/_deposit_variables.py:136`), so whichever wording is chosen, the mechanism it promises exists.

---

### Minor findings

**`verifiable:` Two docstrings still carry the pre-round-1 figure as if current.** `tests/test_citation_verification.py:133` ("moved the refined layer to 2,063") and `:182` ("3,381 against a measured 2,063"). The round-1 shared-normaliser fix landed on **2,061**, stated correctly in the same PR at `scripts/pipeline_text.py:253` and `tests/test_language_table.py:63`, and confirmed by `data-paper-vars.yml:54`. Repro: `grep -rn "2,063\|2,061" tests/ scripts/`. Comments only, no behavioural effect — but the PR is now internally inconsistent about its own headline number. (Consistency)

**`verifiable:` The count guard fires on a legitimate `2100`.** The year strip is `\b(?:19|20)\d{2}\b`, which covers 1900–2099 only. Executed against the guard's exact logic:

```
'Projections run through 2100 under SSP2-4.5.'  -> ['2100']
'embedding dimension is 1536'                   -> ['1536']
'a sample of 837 works'                         -> []          # documented limit
'see Section 2.3 and 8192-token window'         -> []          # allowlisted
```

A climate-finance paper citing an IPCC/SSP horizon year in §2–§3 breaks CI on a number that cannot rot. `_ALLOWED_COUNTS` is the escape hatch and the test says so, so this is a maintenance tax rather than a defect — worth one line in the docstring naming the 2100 case, since the sibling year-strip reads as if it covers all years. (Red team)

**`verifiable:` `reproducibility.md` omits the absent-CSV branch.** Line 84 now says the loaders "prefer Feather when the cache is at least as recent as the CSV, and otherwise read the CSV". `_handoff_cache_is_current` has a third branch the doc does not mention: `if not os.path.exists(csv_path): return True` — with no CSV to compare against, the cache is trusted unconditionally. That is a deliberate and well-reasoned choice (archives may ship either file, per the docstring), and it is the one case where the doc's rule and the code's rule diverge. One clause fixes it. (Doc propagation)

**`consider:` `test_prose_counts_bucket_codes_exactly_as_the_table_does` is true by construction.** `tests/test_language_table.py:69-90`. Since `export_language_table.normalise_language` is now a one-line delegate to `normalize_lang_display`, the equality assertion holds even if both bucketed every code wrong. It is still a useful *re-divergence* guard — and the second half (asserting `compute_vars` mentions the shared symbol) is the part with teeth — but the docstring's "One normaliser, not two look-alikes" reads as validating the buckets, which the sibling tests (`test_locale_normalised_underscore`, `test_iso639_3_code_with_no_two_letter_equivalent_is_unknown`) actually do. Docstring wording only. (Correctness)

**`consider:` The mtime guard is defeated by a DVC `cache.type` change.** Reproduced end-to-end in a throwaway repo: under the committed config (no `cache.type`, so `copy`) a `dvc checkout` that changes content resets mtime to now, so a checkout can only make the guard *too cautious* — false-stale, safe slow path. Under `cache.type = symlink|hardlink`, `os.path.getmtime()` follows to the cache object's original add-time, and a checked-out **older** CSV can read as older than a newer Feather built from newer content — the guard then serves the mismatched cache, silently, which is the exact failure it exists to prevent. Nobody has flipped that config, but `Makefile:176` documents copy-cost pain on this data (~1.7 GB), which is the pressure that leads there. One sentence in the docstring ("assumes DVC's default copy cache; symlink/hardlink caching preserves add-time mtimes and would defeat this") would keep the assumption from being rediscovered the hard way. (Red team)

**`consider:` The CSV fallback path coerces only two columns, and the new tests use 2-row fixtures.** `load_refined_works` explicitly coerces `year` and `cited_by_count`; every other column rides pandas' CSV inference, and this path is now reachable far more often than "first run". The `low_memory=False` added in the same hunk is itself evidence of a real dtype ambiguity at production scale. The panel probed the most exposed consumers and found no live bug — `from_grey`'s reader (`compute_pre2007_coverage.py::_is_grey`) already normalises defensively, and `in_v1` relies on pandas round-tripping its own bool spelling — so nothing to fix here, but the new tests' synthetic `WORKS` dict (2 rows) cannot exercise scale-dependent dtype divergence, and no test covers Feather-vs-CSV dtype parity on the real file. (Red team)

**`consider:` `normalise_language` is now dead indirection.** `scripts/figures/export_language_table.py:70` is a one-line British-spelled delegate to the American-spelled shared symbol. The name predates this PR and two tests import it, so keeping it is defensible; removing it costs two test-import edits. (Consistency)

**`nofollow:` `export_language_table.py`'s module docstring and `ENRICHED_PATH` default still say "enriched corpus"** while `Makefile:279` always passes `--input $(REFINED)` and the generated caption says "refined corpus". Running the script without `--input` would compute from the enriched corpus and caption it as refined. Pre-existing and outside the diff — flagged because this PR spent real effort reasoning about exactly this refined-vs-enriched denominator distinction, so it is the natural moment to notice it. Worth its own ticket rather than a fix here. (Red team, Doc propagation)

**`nofollow:` `low_memory=False` (`pipeline_loaders.py:352`)** aligns with the sibling `load_refined_citations` call and is unmentioned in the PR body's change list. Audit trail only. (Consistency)

**`nofollow:` `STATE.md:15`** still carries an "0297 open author call" note; 0297 closed and the corpus rebuilt before this PR. Out of scope here. (Doc propagation)

---

### Verified, not merely repeated

The panel re-derived the PR body's own claims rather than taking them:

- **Denominator.** `n = len(df)` from `load_refined_works()` (`compute_vars.py:225,229`) — the refined corpus, as claimed. All three new vars are in `DOC_VARS["data-paper"]` (`:104-107`), and `main()` hard-exits if a declared key is never computed ("Rendering would produce ?meta:X placeholders"), so an empty macro cannot ship silently. `test_doc_vars_completeness.py` is bidirectional and parametrised over `DOC_VARS.keys()`, so it covers the three additions.
- **`.str.startswith("detected")`.** The only values ever written to `language_provenance` are `source` / `openalex` / `detected:langdetect` / `""` (`scripts/harvest/enrich_join.py:178-204`); no other value begins with "detected". Prefix match is safe and the stated rationale holds.
- **RED on `origin/main`.** Extracted `origin/main:deliverables/data-paper/data-paper.qmd` and ran the guard's exact logic: returns `['3,381']`. Green on the branch. The claim is reproducible.
- **`en_US`.** `normalize_lang_display` splits only on `-`, but delegates to `normalize_lang`, which splits on `_` (`pipeline_text.py:234-237`) — so `en_US` → `en`, and the pre-existing `test_locale_normalised_underscore` pins it. Not a hole.
- **Ticket 0283 owns the stale R1-13 Crossref figure.** Its log entry of 2026-07-27T09:52Z names the defect in the same three files at the same lines. The "checked, not edited" claim also holds: the diff touches nothing under `revision-rdj26561/`, and R1-16c carries no numeric literal for the non-English layer.
- **No superseded literal survives in a live document.** Repo-wide grep for `89.9`, `89.86`, `3,381`, `3381`: the only remaining hit is `tickets/closed/0320-*.erg:66`, where the old figure is correct as history.
- **`codebook.md` is generator-faithful.** Wording, allowed-values, and group placement (`language_provenance` between `keywords_provenance` and `source_count`) match `DEPOSIT_VARIABLES` verbatim; `Makefile:288` declares the right target and prerequisites. Byte-identical regeneration is **not** verifiable here — `extended_works.csv` is a DVC pipeline-stage output and is absent from this worktree. Stated rather than guessed.
- **Commit decomposition** is six single-purpose commits (`68abc076` test → `be59f505` fix-loader → `f1d17a28` feat-prose → `91f1cdc8` chore → `124b490d` fix-normalizer → `649e1f5c` docs-regen), not a lump.

---

**Panel verdict: comment.** No correctness defect found in the code. One traceability item to resolve before merge (link ticket 0315), one prose-precision question that is the author's to arbitrate ("catalog values alone" vs the `openalex` backfill), and three cheap `verifiable:` cleanups (the 2,063 docstrings, the `2100` guard note, the absent-CSV clause in `reproducibility.md`). Per the project's prose-workpackage rule this PR does not auto-merge — the merge decision is the author's.

🤖 `/review-pr` — five agents, `model: sonnet`, fresh contexts

PR #1149, opened 2026-07-27
fix(0341): escape free bibliographic text in interactive-corpus hover boxes

MinhHaDuong:
**Ticket:** tickets/0341-interactive-corpus-hover-text-interpolat.erg

## What

`scripts/figures/plot_interactive_corpus.py` builds its Plotly hover boxes as raw markup — the `<b>`, `<br>` and `<i>` tags are deliberate — and interpolated `title_short`, `first_author` and `journal_short` straight into that template at three sites. Free bibliographic text carrying an `&` or a `<` therefore reached the tooltip as markup.

The two sibling HTML figure scripts (`plot_genealogy_html.py`, `plot_alluvial_html.py`) already escape free text. This was one script departing from an established convention, so the fix adopts it — with one deliberate exception documented below, which review round 2 forced.

**Escaped the interpolated values, not the template.** The structural tags survive as markup; a test pins that.

## The defect was live, not theoretical

Measured against the current corpus (2,644 core rows): **135 rows carry `&`, `<` or `>` in title or journal.** Samples:

- `Agriculture Ecosystems & Environment`
- `Gender & Development`, `Trends in Ecology & Evolution`
- `Outsourcing CO <sub>2</sub> within China` — a title that arrives from OpenAlex *already carrying literal `<sub>` tags*, previously injected raw into the hover box

## `quote=False`: where this script must not follow its siblings

Review round 2 caught a regression the first version of this PR introduced, invisible to all 13 then-green tests because none exercised a double quote.

Plotly does not hand hover `text` or a trace `name` to the browser's HTML parser. It runs its own `convertEntities`. From the shipped bundle (`plotly.min.js`, 6.6.0):

```js
Vat = {mu:"μ", amp:"&", lt:"<", gt:">", nbsp:"\xA0", times:"×", plusmn:"±", deg:"°"},
Gat = /&(#\d+|#x[\da-fA-F]+|[a-z]+);/g
function oL(e){return e.replace(Gat,function(t,r){ ... n = Vat[r]; return n||t })}
```

`quot` is absent — zero occurrences of `quot:` in the bundle — so an unknown name falls through verbatim. `html.escape`'s default `quote=True` therefore put a literal `&quot;` in the tooltip for **23 of the 2,644 core titles**, e.g. `Assessing "Dangerous Climate Change"`. Those rendered correctly *before* this branch touched them.

Apostrophes were never affected: `html.escape` emits the numeric `&#x27;`, which the `#x` branch above does decode. 148 core titles carry one.

The trap was adopting the sibling convention verbatim. `plot_genealogy_html.py` and `plot_alluvial_html.py` emit SVG for the browser's own parser, which knows the full entity set — `quote=True` is right there. Plotly's eight-entry mini-decoder is the special case. So the fix is `quote=False`, and the module carries a long docstring explaining why, because the next reader's instinct will be to "fix" it back.

Leaving `"` and `'` literal is safe: these strings reach the page as JSON payload values, not as HTML attribute values.

Re-verified over the corpus: **0 of 2,644 rows now emit an entity Plotly cannot decode.**

## The accepted cost

Escaping is not pure gain here, and the first version of this description wrongly implied it was. **24 of the 2,644 core titles carry markup that Plotly's hover renderer whitelists** — mostly `<sub>` arriving from OpenAlex. The headline example above, `Outsourcing CO <sub>2</sub> within China` (608 cites), rendered a genuine subscript before this branch and now shows the literal tag.

Keeping the behaviour anyway: that `<sub>` is a metadata artifact, not markup the figure author chose, and it appears in some records and not others. Escaping data-borne tags uniformly beats rendering whichever ones a source happens to emit. Net is 135 rows repaired against 24 flattened. Restoring subscripts would mean selectively unescaping a tag whitelist after escaping — a separate feature, not filed.

Nothing security-relevant rides on this either way: Plotly's own JSON serialization already made script breakout impossible.

## Design decision: why a separate helper module

The builder went into a new private module `scripts/figures/_hover_text.py` rather than a function inside the plotting script. `plot_interactive_corpus.py` runs its entire pipeline at module scope — it calls `parse_args()` and reads the DVC-gated corpus on import — so an inline builder cannot be imported by a test without executing the pipeline.

The alternatives were worse:

| Option | Why not |
|---|---|
| Re-indent ~370 lines into a `main()` guard | That is the refactor this ticket is not, on a script with no prior test coverage |
| Reimplement the builder inside the test | Tests a copy, not the shipped code. `test_alluvial_unit.py::TestBuildArgv` already had to do this for exactly this reason — a precedent worth not extending |

The chosen form follows existing repo convention: `scripts/figures/` already holds private side-effect-free helper modules (`_community_registry.py`), imported as `from _community_registry import ...` by `plot_fig_global_map.py` and `plot_fig_traditions.py`, and imported by name in tests via `sys.path`. Collapsing three near-duplicate inline blocks was a bonus, not the motive; the motive was a unit-testable seam.

## Second sink, found by the pre-PR adherence pass

`name=f"C{cid}: {label}"` — Plotly renders a trace `name` as markup too, and the cluster label is free text from the generated `cluster_labels.json`. Same defect class, same file, one screen below the hover sites. Fixed here rather than deferred: one line, in the file this ticket owns.

Escaped at the sink, not at the assignment, so `label` stays raw in the variable and the guard can be a real negative ratchet — a regression to the raw interpolation fails the test.

## Behaviour preservation

All call sites reproduce the prior line order and content exactly, including the unassigned-cluster trace's omission of the ISTEX line, now explicit at the call site with a comment instead of implicit in a copy-pasted block. Verified by building hover text for all 2,644 core rows: every row builds, no crash, no undecodable entity.

## Tests

New `tests/test_hover_text_escaping.py`, 19 tests, fast tier (0.15 s — no heavy dependency):

- escaping of `&` and `<` in title, author, journal, and the ISTEX URL
- **double quote and apostrophe left literal**, and `&`/`<` still escaped alongside them
- structural `<b>`/`<br>` survive unescaped
- exact line order and content of the four-line box; the `<i>Cited in manuscript</i>` and ISTEX lines in order
- **`int()` coercion of `year` and `cited_by_count`**, with a Python float and a numpy scalar — production feeds `float64` off pandas, and without these the tooltip reads `Zomer (2008.0)` on every point
- source ratchets: each of the three HTML-emitting figure scripts reaches escaping; the three defect sites no longer interpolate a raw row field; the legend does not interpolate a raw label; `quote=False` holds at both Plotly sinks and nowhere else

Two tests were verified by mutation rather than assumed: stripping both `int()` calls leaves the other 17 green and fails exactly the two new ones.

The ratchet class docstring states plainly what greps cannot prove — a regression reformatted as `row.title_short` would evade them. The behavioural tests over the shipped builder carry the real guarantee; the greps only stop a verbatim revert.

## Verification

| Gate | Result |
|---|---|
| `make check-fast` | **1166 passed, 11 skipped**, 13.4 s |
| `make lint` | **166 passed, 6 skipped**, 15.4 s (1 warning, pre-existing module-length smell, unrelated files) |
| `/verify-adherence` | **PASS**, no mechanical failures |

Three `/review-pr` rounds ran. Round 2 caught the `&quot;` regression; round 3 caught the untested `int()` coercion and the ticket-record gap below.

**This repo has no CI** (no `.github/workflows`, by decision in ticket 0321) — nothing runs the suite on push or on this PR. The results above are local runs, reported here because they are the only gate.

### Figure build check: NOT run

The ticket's third verification bullet ("the figure still builds") could **not** be executed. The script needs two Phase-2 derived artifacts that are gitignored and absent from both this worktree and the primary checkout:

- `data/derived/tables/semantic_clusters.csv`
- `data/derived/tables/cluster_labels.json`

Regenerating them means running the clustering stage over `embeddings.npz` — heavy, and out of scope. The round-2 reviewer went further and confirmed the script is an unwired orphan (`docs/repo-layout.md:208-212`, no Make target), so that bullet is unverifiable on any machine, not only this one.

What *is* verified: the changed code path runs over every real core row without error, the rendered entity set is correct against Plotly's actual decoder, and escaping behaviour is pinned by unit tests. The script has no Make target — it sits in `test_script_hygiene.py`'s `OUTPUT_EXEMPT` as an interactive tool — so nothing downstream regressed.

**Merge-gate note for the author:** the `**Ticket:**` line above closes 0341 on merge with that verification box unticked. The ticket now records why, next to the box.

## Ticket record

0341's exit criterion read "escape free text **the same way**" — which `quote=False` deliberately falsifies. The ticket now carries the arbitration on its log (why the siblings' call is right for them and wrong here), the accepted `<sub>` cost, and an amended criterion stating the rule that holds: escape free text before it reaches markup, with the exact call chosen per output target. Without that, the closed ticket would have read as evidence *against* the code it shipped.

## Scope

Touches four files: the script, the new helper, the new test, and the ticket record. The two sibling scripts were **not** modified — they already escape correctly, and their `quote=True` is correct for their own output target.

Also fixes three stale lines in the script's header docstring (the pre-0240 `Usage:` and `Output:` paths, and a "~1,176 core papers" count that is 2,644 under the same filter).

Related follow-up: tickets/0351-interactive-corpus-figure-is-an-unwired.erg — the figure script has no Make target, which is why this PR could not tick "the figure still builds"; filed as a wire-in-or-retire decision, not resolved here.


MinhHaDuong (commented):
## Multi-perspective review — PR #1149 (ticket 0341)

Five reviewers (correctness, consistency, scope, red team, doc propagation), all on the sibling tier. **No blockers.** Panel verdicts: 1 approve, 4 comment.

The fix is right and the refactor is behaviour-preserving. What follows is the minor tier plus one dissent worth recording.

### Gates re-run on the head commit (`8e2e3f0`)

| Gate | Result |
|---|---|
| `make lint` | 166 passed, 6 skipped, 13.0 s (1 pre-existing module-length warning, unrelated files) |
| `make check-fast` | 1160 passed, 11 skipped, 12.7 s |
| `make check` | 1770 passed, 77 skipped, **7 failed**, 6 min 22 s |
| `tests/test_hover_text_escaping.py` | 13 passed, 0.12 s |

The `make check` failures are all `tests/test_corpus_acceptance.py::TestFileExistence` / `TestCorpusAlign` — `AssertionError: <file> missing at data/catalogs/…` for `corpus_audit.csv`, `embeddings.npz`, `citations.csv`, `llm_relevance_cache.csv`, `refined_embeddings.npz`, `refined_citations.csv`. The worktree's `data/catalogs/` holds only `.dvc` pointer files; the data lives in the primary checkout. Environmental, not caused by this PR — it touches no corpus artifact. The two gates the PR body quotes reproduce exactly.

### Verified claims

- Behaviour preservation at all three rewired sites: line order, content, and both conditionals are identical to the deleted inline blocks. `int(year)` / `int(cited_by_count)` and the `if istex_url:` truthiness moved unchanged; upstream `core` filters guarantee non-NaN on both integer columns, and `make_istex_url` always returns a `str`.
- The "no ISTEX line on the unassigned trace" comment is accurate, not a preserved bug dressed as intent — red team traced it to the file's first commit (`115cc54`).
- The sibling scripts really do use `import html as html_mod` + `html_mod.escape` on every free-text field (`plot_genealogy_html.py:230-240,269`; `plot_alluvial_html.py:234-236`). The "same convention" claim is not overstated, and neither sibling was touched.
- The module extraction is justified: `plot_interactive_corpus.py` executes `parse_args()`, `load_analysis_config()`, and `pd.read_csv()` at module scope with no `__main__` guard, so an inline builder is genuinely unimportable. `_hover_text.py` matches the `_community_registry.py` precedent, and its `_`-prefix earns automatic exemption from the hygiene registries (`test_script_hygiene.py`) — no allowlist edit was owed.
- No remaining unescaped free-text sink in the file. Layout titles and period annotations are hardcoded; `customdata` carries `doi_url` through Plotly's JSON serialization, a different sink class.

### Dissent — preserved verbatim, the author decides

The two reviewers who looked hardest at the source-grep ratchets disagree about what they guard.

**Correctness:** "I verified empirically (not just reasoned) against `/tmp/plot_old.py` (the pre-fix version pulled from `main`) that the exact literal patterns the tests search for … are present in the old file and absent in the new file — confirmed the grep patterns use single-quote `row['...']` matching the old code's actual quoting, so the ratchet is not accidentally quote-mismatched into always-passing. … Genuine ratchet."

**Red team:** "The three `TestAllHtmlFigureScriptsEscape` ratchet tests are pure substring greps and were confirmed bypassable … `row['title_short']` → `row.title_short` (attribute access) or `row[\"title_short\"]` (double quotes) evades `test_interactive_corpus_no_raw_free_text_interpolation` for all three fields. … `test_each_script_reaches_escaping` checks only whether `html_mod.escape` or `build_hover_text` appears **anywhere in the whole file** — a single surviving call … keeps this passing while every hover-text field is unescaped again."

Both are right about different threat models: the ratchet catches a literal revert, not a paraphrase.

### Minor findings

**`verifiable:` Stale `Output:` line in the docstring the PR already edited** *(consistency, doc propagation — both high confidence)*
`scripts/figures/plot_interactive_corpus.py:13` still reads `Output: figures/interactive_core_corpus.html`. The real path is `deliverables/_shared/figures/interactive_core_corpus.html` (`FIGURES_DIR`, line 47; `output_path`, line 407). The PR fixed the adjacent `Usage:` line (`scripts/` → `scripts/figures/`) and missed its neighbour. Same file, same docstring, one line.

**`verifiable:` The exit-criterion ratchets pass on a reformatted regression** *(red team)*
Evaluated the tests' own assert predicates against hostile rewrites of the current source:

```
guard catches attribute-access form (row.title_short):  False
guard catches local-variable form (t = row["title_short"]): False
```

Non-blocking because the real guarantee lives in `TestEscaping` / `TestStructure`, which exercise `build_hover_text` behaviourally. But `TestAllHtmlFigureScriptsEscape` is the class named after the exit criterion, and as written it enforces less than its docstring claims. An AST walk over f-string contents, or asserting `build_hover_text` is the sole interpolation site for those fields, would close it.

**`consider:` Pre-escaped entities would double-escape** *(red team)*
`build_hover_text(journal="Energy Research &amp; Social Science")` returns `&amp;amp;`. Measured against the current corpus to check reachability: **0 of 2,644 core rows** carry an HTML entity in `title` or `journal` (143 carry raw `&`/`<`/`>` — 39 titles, 104 journals, consistent with the PR body's 135 rows). Not live today. Worth knowing that `scripts/pipeline_text.py` documents double-encoded entities as a known artifact class in this corpus, so the input is not impossible, only absent.

**`consider:` NaN `istex_url` is truthy** *(red team)*
`_hover_text.py:40` — `if istex_url:` renders a literal `ISTEX PDF: nan` line for a NaN. Unreachable from the three current call sites (`make_istex_url` returns a plain string), latent in a now-shared helper. Identical to the pre-PR behaviour, so no regression.

**`consider:` Type-hint style is an outlier among sibling private helpers** *(consistency)*
`_hover_text.py:16-24` uses keyword-only args with full annotations; every existing private helper in `scripts/` and `scripts/figures/` (`_community_registry.py`, `_divergence_io.py`, `_pre2007_traditions.py`, `_venue_naming.py`, `_course_dedup.py`) uses bare positional args with no hints. Compliant with `coding-python.md`, arguably better — but it is a deviation from the local pattern, worth a deliberate answer either way.

**`consider:` `script-io.md` migration not taken, no follow-up filed** *(consistency)*
The rule says a touched script gets migrated to `parse_io_args` / `--output` "in the same commit if the change is small, or open a follow-up ticket". `plot_interactive_corpus.py` was touched for 60 lines and still hardcodes its output path (lines 47-49, 407). Out of 0341's scope, but neither branch of the rule was taken.

**`consider:` One redundant test** *(scope)*
`test_helper_uses_the_sibling_escape_convention` greps `_hover_text.py` for the escape call; the six `TestEscaping` tests already prove the behaviour more strongly.

**`nofollow:` Two disclosed scope riders** *(scope)*
The legend-label fix (`plot_interactive_corpus.py:251`) is a fourth site the ticket does not name, and the docstring path fix is a second unticketed change. Both are one line, both are argued in the PR body, both are the ticket's own defect class or adjacent to it. The stricter reading of `feedback_scope_discipline` would have made the legend a sibling ticket; the panel does not think it is worth the round-trip here.

**`nofollow:` `first_author` unguarded against NaN** *(correctness, red team)*
`plot_interactive_corpus.py:236` — unlike `title_short` / `journal_short`, `first_author` has no `.fillna("")`. Pre-existing, unchanged by this PR, renders an escaped `nan` in both old and new code.

### Merge-gate note, not a review finding

Ticket 0341's third verification bullet — "The figure still builds" — cannot honestly be ticked. Confirmed independently: `data/derived/tables/cluster_labels.json` and `semantic_clusters.csv` are absent (only `cluster_labels_clean.json` is present), and the script fails earlier still on the DVC-gated `refined_works.csv` from within a worktree. The PR body states this plainly, which is the right call. But the PR carries a closing `**Ticket:**` line, so `erg-pr-merge` will close 0341 unconditionally on merge with that box unticked. Either accept it — the script has no Make target and is listed in `OUTPUT_EXEMPT`, so nothing downstream regressed — or run the render on a machine with the derived tables before merging.

---

*Panel: correctness (approve), consistency (comment), scope (comment), red team (comment), doc propagation (comment). No reviewer requested changes.*

MinhHaDuong (commented):
## Multi-perspective review — PR #1149, **round 2** (head `aab63fff`)

Five reviewers on the sibling tier (correctness, consistency, scope, red team, doc propagation), re-run against the round-1 fix commit. Panel verdicts: 3 approve, 2 comment.

Round 1's two `verifiable:` findings are **confirmed fixed**. The panel found **one new defect that round 1 missed, and it is a regression this PR introduces** — verified independently below, not taken on the reviewer's word.

### Gates re-run on `aab63fff`

| Gate | Result |
|---|---|
| `make check-fast` | **1160 passed, 11 skipped**, 11.5 s |
| `make lint` | **166 passed, 6 skipped**, 10.8 s (1 pre-existing module-length warning, unrelated files) |
| `tests/test_hover_text_escaping.py` | **13 passed**, 0.34 s |

### Round-1 fixes verified

- **`Output:` docstring path** — `plot_interactive_corpus.py:13` now reads `deliverables/_shared/figures/interactive_core_corpus.html`, traced to `FIGURES_DIR` (line 47) and `output_path` (line 407). Correct.
- **Narrowed `TestAllHtmlFigureScriptsEscape` docstring** — re-read all four methods against the new text; it no longer overclaims. Naming what the greps cannot prove was the right response to the round-1 red-team demo.

---

## Blocker

### `verifiable:` `html.escape()`'s default `quote=True` breaks 23 core titles in the rendered figure *(red team, high confidence — independently reproduced)*

This is a **behaviour regression introduced by this PR**, in the exact code path it changes. Before the PR a `"` in a title rendered as `"`. After it, it renders as the literal six characters `&quot;`.

Plotly does not parse the hover `text` / trace `name` as HTML. It whitelists a few tags and decodes entities through its own eight-entry table. From `.venv/…/plotly/package_data/plotly.min.js`:

```js
var Vat = {mu:"μ", amp:"&", lt:"<", gt:">", nbsp:"\xA0", times:"\xD7", plusmn:"\xB1", deg:"\xB0"},
    Gat = /&(#\d+|#x[\da-fA-F]+|[a-z]+);/g;
function oL(e){ return e.replace(Gat, function(t,r){ var n;
  return r.charAt(0)==="#" ? n=Hat(...) : n=Vat[r], n||t }) }
```

`quot` is absent from `Vat` (confirmed: the string `quot:` occurs nowhere in `plotly.min.js`), so the lookup yields `undefined` and the replacer returns the original token unchanged. `html.escape(s)` with the default `quote=True` emits `&quot;` for `"` — which therefore survives to the screen verbatim.

Reachable now, measured against the current corpus:

```
core rows (cited_by_count >= 50):        2644
title_short containing a double quote:     23
```

e.g. `Assessing "Dangerous Climate Change": Required Reduction of Carbon Emissions…`, `Constructing a Climate Change Logic: … the "Tragedy of the Commons"`.

Apostrophes are **fine** — `html.escape` emits the *numeric* `&#x27;`, which `Gat`'s `#x[\da-fA-F]+` branch decodes via `String.fromCodePoint`. So this is `"`-only: 23 titles, 0 journals, 0 authors.

**Why adopting the sibling convention verbatim was the trap.** `plot_genealogy_html.py` and `plot_alluvial_html.py` write raw SVG/HTML consumed by a *real* browser parser, where `&quot;` decodes correctly — the default `quote=True` is harmless there, and `plot_genealogy_html.py:240` even passes it explicitly for an attribute. Plotly's mini-decoder is the special case. The convention was right for `&<>` and wrong for `"`.

**Fix**: `html_mod.escape(str(x), quote=False)` in `_hover_text.py` (the four escape calls) and at the legend sink `plot_interactive_corpus.py:251`. These strings land in text nodes, never in attributes, so attribute-quote safety buys nothing. Add a `"`-in-title case to `TestEscaping` — no test in the file currently exercises a double quote in any field, which is why 13 green tests did not catch it.

---

## Minor findings

**`verifiable:` The docstring's "~1,176 core papers" is stale, in the header block this PR has now edited twice** *(consistency + doc propagation, both high confidence — independently reproduced)*

`plot_interactive_corpus.py:3` reads "~1,176 core papers (cited_by_count >= 50)". That parenthetical is exactly reproducible, and it yields **2,644**, not 1,176:

```
refined_works.csv, cited_by_count >= 50  ->  2644 rows
```

This PR's own commit message (`bfeeb044`) states "2,644 core rows in the current corpus". `CITE_THRESHOLD` is still 50 (`config/analysis.yaml:12`), so this is corpus growth, not a threshold change — the "hand-typed figures rot at every corpus rebuild" pattern. Round 1 flagged the sweep for fixing one docstring line and missing its neighbour; this is the third neighbour, two lines above the one that was fixed. Either update it or drop the number and cite the threshold alone.

**`consider:` The test module's `sys.path.insert` leaks for the whole pytest session** *(red team, high confidence; latent)*

`tests/test_hover_text_escaping.py:20` prepends `scripts/figures` to `sys.path` at import time with no teardown; the entry was empirically confirmed to persist ahead of the `scripts` pythonpath root for every module collected afterwards. No basename collision exists today (77 files in `scripts/figures/` vs 46 in `scripts/`, checked). But — and this is why it is only `consider:` — **this is the established repo idiom, not a new one**: nine sibling tests do exactly the same (`test_analytical_null.py:13`, `test_bootstrap.py:22`, `test_corpus_table_export.py:14`, `test_divergence.py:26`, `test_global_map.py:17`, `test_language_table.py:12`, `test_subsampling_ribbon.py:23`, `test_zoo_figure_polish.py:15`, `test_zoo_null_ci.py:12`). Fixing it here alone would create a second convention. If it is worth fixing, it is worth a sweep ticket using `monkeypatch.syspath_prepend`.

**`consider:` Four near-duplicate escaping tests** *(scope)*

`test_ampersand_in_title_is_escaped` / `test_ampersand_in_journal_is_escaped` and `test_angle_brackets_in_title_are_escaped` / `test_angle_brackets_in_author_are_escaped` differ only in the kwarg name — every field funnels through one `html_mod.escape` call in one function. They collapse to two `@pytest.mark.parametrize`d tests over `(field, char)` with no loss. Together with the round-1-flagged `test_helper_uses_the_sibling_escape_convention`, that is 3 of 13 tests carrying no distinct guarantee. Cheap and honest as written, so this is a nit — but if the blocker above is fixed, the `"` case should be *added* to a parametrized matrix rather than as a sixth near-clone.

**`nofollow:` `_hover_text.py` needs no registry entry — claim verified, not taken on faith** *(doc propagation)*

The PR says the `_` prefix earns automatic hygiene exemption. Reading the actual logic, that is right for the *naming* check (`test_all_scripts_have_conforming_prefix`: `if f.name.startswith("_") … continue`) but the exemptions that actually matter here are structural, not name-based: `TestOutputFlag` (`test_script_hygiene.py:1206-1214`) and `TestArgparsePresence` (`:219-236`) both skip any script with no `if __name__ == "__main__":` guard. `_hover_text.py` has none. No allowlist, Makefile, docs index, or coverage entry is owed — same standing as `_community_registry.py`, which is likewise unlisted. Repo-wide grep for `plot_interactive_corpus` / `interactive_core_corpus` / `_hover_text` turned up no stale reference.

**`nofollow:` PR body claim "135 rows carry `&`, `<` or `>`" now measures 139** *(doc propagation, medium confidence)*

Data drift between the author's measurement and this review, not a code defect. Noted only so the number is not re-cited later as a fixed fact.

---

## Merge-gate notes, not review findings

**The ticket's "figure still builds" bullet is unverifiable by design, not merely blocked.** The PR's excuse checks out — I confirmed `data/derived/tables/semantic_clusters.csv` and `cluster_labels.json` (the exact paths at `plot_interactive_corpus.py:52-53`) are absent from **both** the worktree and the primary checkout; only `cluster_labels_clean.json` is there, and the same-named `semantic_clusters.csv` in `data/catalogs/` is a legacy pre-split artifact at a path the script never reads. Nothing was suppressed. But `docs/repo-layout.md:208-212` independently documents this script as one of four **"unwired orphans"** — no `__main__` guard, no Make target, "nothing invokes them" — corroborated by a zero-hit grep for `plot_interactive_corpus` across `Makefile` and `scripts/**/*.mk`. So that ticket line can never be ticked as written, on any machine, until the script is wired. Worth recording on 0341 rather than leaving the box to look like an oversight.

**Branch is 27 commits behind `origin/main`.** `mergeStateStatus` is CLEAN and the changed files are untouched upstream, so per the force-push-denied clause in `rules/git.md` this is a currency nit, not a blocker — but re-run the gates after integrating if the base moves further.

---

### Verdict: **request-changes**

One blocker, one line of fix, one test case. Everything else on this PR is in good shape: the escaping fix is right in substance, the extraction is justified, behaviour preservation was verified at all three sites in round 1 and holds, and both round-1 fixes landed cleanly. The `&quot;` gap is precisely the kind of defect that survives a green suite — the tests pin `&`, `<`, `>` and the URL case, and the one character Plotly's decoder cannot handle was never exercised.

*Panel: correctness (approve), consistency (comment), scope (approve), red team (comment), doc propagation (approve). No reviewer requested changes; the escalation to request-changes is the synthesis call, on the ground that the finding is a regression introduced by this PR (`rules/review-pr` severity table) and the fix is one keyword argument.*

MinhHaDuong (commented):
## Multi-perspective review — PR #1149, **round 3** (head `f009c316`)

Five reviewers on the sibling tier (correctness, consistency, scope, red team, doc propagation). Panel verdicts: **2 approve, 3 comment. No blockers.**

Round 2's `&quot;` regression is **confirmed fixed and independently re-verified** — two reviewers read the shipped Plotly bundle rather than trusting the docstring. The remaining findings are all in the *safety net around* the fix, not the fix. Every `verifiable:` item below was reproduced by the synthesizer, not taken on a reviewer's word.

### Gates re-run on `f009c316`

| Gate | Result |
|---|---|
| `tests/test_hover_text_escaping.py` | **17 passed**, 0.09 s |
| `make check-fast` | **green**, exit 0 |
| `make lint` | **166 passed, 6 skipped**, 23.2 s (1 pre-existing module-length warning, unrelated files) |
| `make check` | **1774 passed, 77 skipped, 7 failed**, exit 0 on the libs suite |

The 7 failures are the same `tests/test_corpus_acceptance.py::TestFileExistence` / `TestCorpusAlign` set round 1 reported — `refined_works.csv`, `corpus_audit.csv`, `embeddings.npz`, `citations.csv`, `llm_relevance_cache.csv`, `refined_embeddings.npz`, `refined_citations.csv` missing under `data/catalogs/` because the review worktree holds only DVC pointers. Environmental; this PR touches no corpus artifact.

---

## Verified correct

Recorded because three of these were the PR body's own load-bearing claims, and each was checked against the artifact rather than the prose.

- **The Plotly entity table is exactly as documented.** Both correctness and red team disassembled `.venv/lib/python3.12/site-packages/plotly/package_data/plotly.min.js`: `Vat={mu,amp,lt,gt,nbsp,times,plusmn,deg}`, regex `/&(#\d+|#x[\da-fA-F]+|[a-z]+);/g`. `quot` genuinely absent; the numeric branch genuinely decodes `&#x27;`; `gt` present, so `>`→`&gt;` round-trips. `quote=False` is right.
- **Legend `name=` reaches the same renderer.** `convertEntities` is called from `convertToTspans`, Plotly's single shared text path for hover labels, trace names, titles and annotations alike. The legend escape is correct, not superfluous.
- **The sibling claim holds — it is not a second latent defect.** `plot_genealogy_html.py` and `plot_alluvial_html.py` contain *zero* Plotly usage (`grep -n "plotly\|go\.\|hoverinfo"` → 0 matches in both). They build raw SVG consumed via `el.dataset.tooltip` → `tooltip.innerHTML`, a genuine two-stage browser parse where `quote=True` is required. `plot_genealogy_html.py:240` even passes explicit `quote=True` for an attribute value. The divergence is real engineering, not an oversight.
- **Script-breakout is impossible, independently of this PR.** Red team built a trace with `text=["<b>Evil </script><script>alert(1)</script></b>"]` and **no escaping at all**; `fig.to_html()` still emitted `</script>` exactly once. `plotly/io/_json.py:60-71` `_safe()` rewrites `<`→`<`, `>`→`>`, `/`→`/` across the whole JSON payload. The escaping in this PR is a *display-correctness* fix, not a security fix — worth knowing when weighing finding 2.
- **No double-escaping in the live corpus.** 0 of 33,344 rows in `data/catalogs/refined_works.csv` carry a pre-existing `&amp;` or numeric entity in title/journal/first_author.
- **Behaviour preservation at all three rewired sites**: line order, four-line skeleton, both conditionals, and the `int()` coercions are identical to the deleted inline blocks. `core` is filtered on `axis_score.notna() & year.notna()`, and `cited_by_count >= CITE_THRESHOLD` excludes NaN, so neither `int()` can raise. `make_istex_url()` returns `str` always, never NaN — the falsy-NaN trap does not apply.
- **`2,644` matches the authoritative artifacts**: `STATE.md:36` and `deliverables/_shared/tables/tab_citation_coverage_periods.csv:27` (`core_n,2644.0`). No other doc still states the stale `~1,176`; that sweep is complete. Docstring `Usage:`/`Output:` paths match the real `output_path`.
- **Mutations the suite *does* catch**: reverting to `quote=True` (4 failures), unescaping `first_author` alone (1), reordering the ISTEX/manuscript lines (1), escaping the structural tags (11).

---

## Findings

### 1. `verifiable:` Dropping the `int()` coercion is invisible to all 17 tests *(red team, high confidence — reproduced)*

`scripts/figures/_hover_text.py:54,56`. Every test passes plain Python `int` literals; production feeds numpy `float64` (`works["year"] = pd.to_numeric(...)`). Reproduced in a scratch tree at `/tmp/mut1149`:

```
mutation: {int(year)} -> {year},  {int(cited_by_count)} -> {cited_by_count}
output:   '<b>T</b><br>A (2013.0)<br>J<br>Cited by: 608.0'
pytest tests/test_hover_text_escaping.py -q  ->  17 passed in 0.09s
```

A silent `(2013.0)` / `Cited by: 608.0` on every point in the figure, green suite. One test passing `year=2013.0, cited_by_count=608.0` closes it.

### 2. `verifiable:` 24 highly-cited titles lose genuine subscript rendering — an undiscussed cost *(red team, high confidence — recounted by the synthesizer)*

OpenAlex embeds real structural markup in some titles, and Plotly's hover renderer whitelists `{sup, sub, s, u, b, i, a, span, em}`. Escaping neutralizes it. Recounted directly against `data/catalogs/refined_works.csv`, core subset (`cited_by_count >= 50`, n = 2,644 — matches the PR's own figure): **24 core titles carry a whitelisted tag.**

```
608  Outsourcing CO <sub>2</sub> within China
470  The European Union Emissions Trading System reduced CO <sub>2</sub> emissions…
362  Presidential Address: Sustainable Finance and ESG Issues—<i>Value</i>versus<i>Values</i>
358  Terrestrial gross primary production: Using NIR <sub>V</sub> to scale…
342  Quantifying CO<sub>2</sub> Emissions From Individual Power Plants From Space
```

The PR body cites the first of these as *evidence for* the fix. It is also the clearest case against: before this branch it rendered a real subscript in the tooltip; after it, the literal characters `<sub>2</sub>`.

The trade is still net positive — 135 rows with a corrupting `&`/`<`/`>` repaired against 24 rows with markup flattened — and given the `_safe()` finding above, nothing about security rides on it. But it is a real rendering regression on some of the corpus's most-cited papers, presented in the PR as pure gain. **Ask: one line in the ticket log naming the trade-off.** Not a merge blocker.

### 3. `verifiable:` The grep ratchets are bypassable exactly as their own docstring concedes *(red team, high confidence — reproduced)*

`tests/test_hover_text_escaping.py:224-296`. Reproduced at `/tmp/mut2-1149` — reintroducing raw free text at a *new* hover sink:

```python
jshort = row["journal_short"]
hover_texts[-1] += f"<br>Also in: {jshort}"
```

→ **17 passed.** All five ratchets and all twelve behavioural tests green, with an unescaped journal name (many carry a literal `&`) back in the markup.

Round 1 already forced the class docstring to stop overclaiming, and it now states this plainly. Recording the working PoC so the limitation is on file as demonstrated rather than merely asserted — no action required.

### 4. `verifiable:` Ticket 0341's exit criterion is now false by design, and the ticket carries no record of the arbitration *(consistency + scope + doc propagation — three-way convergence)*

`git diff origin/main...pr1149 -- tickets/0341-*.erg` is **empty**. The ticket still reads:

> Actions: Apply `html.escape()` … **matching the sibling scripts**
> Exit criteria: All three HTML-emitting figure scripts escape free text **the same way**

The shipped code deliberately does the opposite — `_hover_text.py:19-30` says `quote=False` "is where this module must NOT follow the siblings," and `test_plotly_sinks_never_escape_quotes` pins that divergence as correct. The `--- log ---` section has no entry after creation; no verification checkbox is ticked.

`tickets/AGENTS.md` puts "the reasoning behind a choice" in the ticket body, not only in commits and PR text. `erg-pr-merge` closes 0341 unconditionally on the `**Ticket:**` line, so it archives with an exit criterion reading as unmet and no explanation on file — and a future reader of `tickets/closed/0341-*.erg` would flag today's `quote=False` as a regression.

**Ask before merge:** a log entry on 0341 recording (a) the `quote=False` arbitration and why, (b) that "the same way" is superseded — suggest *"all three escape free text at every interpolation site, with the quote mode appropriate to each sink's consumer"* — and (c) that "the figure still builds" is unverifiable here, not merely deferred.

### 5. `consider:` **Dissent — tier marker on the ratchet class.** Two reviewers disagree; recorded verbatim for the author

**Consistency (high confidence):** `TestAllHtmlFigureScriptsEscape` is a pure source-grep ratchet and carries no marker (`grep -n "pytest\|mark\." tests/test_hover_text_escaping.py` → 0 matches), so it runs in the fast tier. Every comparable ratchet in this repo is adherence-marked: `tests/test_io_discipline.py:351` and `tests/test_script_hygiene.py:36`, both `pytestmark = pytest.mark.adherence`; `coding-python.md`'s tier table files "hygiene / contracts" under adherence explicitly.

**Doc propagation (high confidence):** the file carries no marker and *correctly* lands in the fast tier — pure Python, no subprocess, no heavy import, 0.07 s.

**Synthesizer's note, offered as a tiebreak rather than a verdict:** both are right about their own half, because the file is *mixed*. `TestEscaping` and `TestStructure` are genuine behavioural fast tests that belong in the keystroke loop; only `TestAllHtmlFigureScriptsEscape` is a ratchet. A module-level `pytestmark` — the form both cited precedents use — would wrongly evict the behavioural tests from `make check-fast`. If this is adopted, the precise form is a class-level `@pytest.mark.adherence` on the ratchet class alone. Nothing enforces it either way: `test_script_hygiene.py::test_lint_tests_are_adherence` only catches files invoking ruff/mypy. At 0.09 s for the whole file the cost is nil, so this is convention purity, not budget.

### 6. `consider:` `quote=False`'s safety rationale is sink-shaped, not enforced *(red team, medium confidence)*

The docstring's "JSON payload values, not HTML attribute values" is true for every sink in this diff, and verified. But Plotly's whitelist includes `<a>` and `<span style="…">`, which do take attributes, and the file already has an adjacent click-to-DOI pattern (`customdata` → `window.open`). An `<a href="…">` hover variant added later would silently inherit `quote=False`. The `test_plotly_sinks_never_escape_quotes` ratchet would then *enforce the wrong thing* — it asserts `quote=False` on every `html_mod.escape(` line in these two files, so the correct `quote=True` for a future attribute sink would fail the suite. Worth a sentence in the module docstring naming the boundary.

### 7. `consider:` `2,644` is now hardcoded in three places *(consistency + scope)*

`plot_interactive_corpus.py:4`, `_hover_text.py:17`, `tests/test_hover_text_escaping.py:95`. Replacing a hedged `~1,176` with a hard, unhedged number triples the rot surface — the same class this project's own memory flags (*hand-typed figures rot at every corpus rebuild*). Comment-only, no test asserts on it, so it degrades gracefully. Red team already measured the related `23 of 2,644 core titles carry a "` claim at **25** today: drift after four days.

### 8. `consider:` `test_plotly_sinks_never_escape_quotes` is line-shaped *(scope, medium confidence)*

`tests/test_hover_text_escaping.py:252-267` requires `quote=False` on the *same source line* as each `html_mod.escape(`. A formatter wrapping a long call across lines would put the kwarg on a different line and fail on unchanged behaviour. Not live today — every current call is single-line.

### 9. `nofollow:` Pre-existing, unaffected by this diff

- `first_author` has no `.fillna("")` unlike `title_short`/`journal_short` (9 of 2,644 core rows are empty → `"nan (2013)"` in the box). Identical in pre-0341 code; outside the ticket's escaping scope.
- `plot_interactive_corpus.py` still uses bare `argparse` with no `--output`, unmigrated to `script-io.md`. Untouched by this diff; already in `test_script_hygiene.py`'s `OUTPUT_EXEMPT`.
- `normalize_doi()` does no DOI-shape validation before `customdata` → `window.open`. 0/33,344 rows non-standard today; predates this ticket.
- `_hover_text.py`'s 27-line docstring dwarfs sibling `_community_registry.py`'s 5. No rule caps it, and the length documents the one thing a future maintainer would otherwise "fix" back wrongly. Justified.

### 10. `nofollow:` Confirmed genuinely unverifiable, not skipped

"The figure still builds" cannot be ticked anywhere in this repo. `data/derived/tables/semantic_clusters.csv` and `cluster_labels.json` are absent from both the worktree and the primary checkout; `grep -rn "plot_interactive_corpus\|interactive_core_corpus" Makefile scripts/analysis/*.mk` returns nothing — no target exists, matching `docs/repo-layout.md:209`'s "unwired orphans." A synthetic smoke test would need fixtures across `embeddings.npz`, `semantic_clusters.csv`, `cluster_labels.json` and `main.bib` at once, well past the ticket's stated scope. No propagation debt either: `_community_registry.py`, the cited precedent, is likewise absent from `repo-layout.md`'s table, which is scoped to the pre-0254 top-level `scripts/*.py` era.

---

## Summary

The fix is correct and now verified against the actual Plotly runtime from two independent directions. Nothing here blocks the merge.

Two asks before it lands, both cheap:

1. **A log entry on ticket 0341** (finding 4) — it closes unconditionally with a superseded exit criterion and no record of the `quote=False` arbitration, and finding 2's trade-off belongs in the same entry.
2. **One test pinning the `int()` coercion** (finding 1) — the one gap where a live silent regression is a single deleted call away.

Findings 5-8 are convention and future-proofing; the author's call. Finding 3 is on file as a demonstrated limitation the code already documents honestly.

PR #1152, opened 2026-07-27
fix(0339): escape free-text venue names in both Markdown table emitters

MinhHaDuong:
**Ticket:** tickets/0339-markdown-pipe-table-emitters-interpolate.erg

Both venue-table emitters interpolate a **journal name** — free text straight from the corpus — into a Markdown pipe table without escaping. `refined_works.csv` carries ten bilingual `journal` values joined with a literal `|`, so this is one corpus rebuild away from firing, not hypothetical. It is latent today only because no pipe-bearing venue sits in the current top-N.

`export_tab_venues.py` feeds `@tbl-venues` in the rendered manuscript (both `manuscript.qmd` and `manuscript-Gide.qmd`), so this instance reaches a published page.

## What the defect looks like

Not a crash, and not even a malformed row. GFM truncates an overflowing row to the header's declared column count, so `export_core_venues_markdown.py` emits a row with all three declared cells:

| Venue | Papers | Type |
|:--|--:|:--|
| International Development Policy | Revue internationale | 2 |

Well-formed, and wrong: the second half of the venue name is in the Papers column, the paper count is in Type, and the type is gone. Only reading the cell values back distinguishes it from a correct row.

## Changes

- **New `scripts/_markdown_table.py`** — the escaping's single home, lifted out of `_deposit_variables.py`. Three call sites import it.
- Both emitters route their venue column through it; numeric and curated-label columns are left alone.

## Two judgement calls

**Module choice.** A flat `_`-prefixed module under `scripts/`, following `_venue_naming.py` verbatim (ticket 0254, the 0250 pattern): shared pure helpers sit on the flat library surface so entry points stay free to move by phase. `pipeline_text.py` was the other candidate and is the wrong one — it normalises *corpus* text (encoding artifacts, DOIs, language codes) and pulls in pandas, ftfy and langdetect to do it. Escaping markup is a different concern with no dependencies at all.

**Whether the escaper may raise.** `markdown_cell()` raises `ValueError` on an unbalanced backtick. That is correct where it was written — its input is a codebook description authored in this repo, where an odd backtick is a typo whose intent is genuinely ambiguous.

It is wrong for these emitters, and the crash risk is the lesser reason. The real problem is that the two callers hold different input *contracts*: a journal name is not Markdown. It has no code spans, so a name carrying two *balanced* backticks — no exception raised — would be silently typeset half in monospace with the backticks dropped. A merely non-raising variant of the same function would still get that wrong; the raise is a symptom, not the disease.

So the split is by contract, not by error policy:

- `markdown_cell()` — input **is** Markdown. Keeps the prose/code-span rule and keeps the raise.
- `markdown_text_cell()` — input is plain text. Escapes backslash, pipe and backtick uniformly. Never raises.

The escaping rules live in one module either way, which is what the ticket's "one escaping implementation" asks for. `markdown_text_cell` deliberately stops at those three characters: they are the ones a single occurrence misrenders, whereas emphasis and link syntax need a matched pair. Escaping all CommonMark ASCII punctuation would be equally correct to a renderer and would churn every shipped row (`Well\-known Journal`) for no gain. That scope limit is now pinned by a test.

## Verification

**No-op check — ran, on real data.** Both tables were regenerated from the current corpus by pre-change code (a throwaway worktree at `origin/main`) and post-change code, on identical inputs:

```
cmp old/tab_venues.md new/tab_venues.md              → IDENTICAL
cmp old/tab_core_venues_top10.md new/…              → IDENTICAL
```

No escape fires on today's data — the only backslash in the regenerated `tab_venues.md` is the caption's pre-existing `$\geq$`.

One honest caveat: the *committed* `tab_venues.md` does differ from a fresh regeneration (a few pole counts moved). That drift is identical under both versions — it is `tab_pole_papers.csv` having been regenerated since the artifact was committed, and is untouched by this PR. I have not refreshed the artifact, since that is corpus churn unrelated to this ticket.

**Tests.** Red-first, on the rendered page rather than the emitted source (0325's lesson). The tests drive the real emitters as subprocesses rather than re-deriving a row in-test, so they still fail if an emitter stops calling the escaper — verified by mutation.

An earlier revision of these tests asserted `row.count("<td") == 5`, which the adherence review correctly called vacuous: the count is 5 whether or not the venue is escaped. Replaced with a full cell-tuple assertion, and re-verified by mutation that it now catches the unescaped emitter.

## Local test results — no CI in this repo

This repo has no workflows (ticket 0321): **nothing runs on push or merge.** All gates were run locally in the PR worktree.

| Gate | Result |
|---|---|
| `make lint` | **166 passed, 6 skipped** |
| `make check-fast` | **1160 passed, 11 skipped** (+13 new unit tests) |
| `make check` | **1772 passed, 7 failed, 77 skipped** |
| `/verify-adherence` | **PASS** (4 nits, all fixed in `28afd7d2`) |
| `/review-pr` | **comment**, no blockers (findings fixed in `9b86989b`) |

The 7 `make check` failures are all `test_corpus_acceptance.py` file-existence checks for DVC-managed corpus files not pulled into this worktree (`refined_works.csv`, `embeddings.npz`, …). They are environmental, not caused by this branch: pointed at the real corpus in the primary checkout, the same module is **44 passed, 12 skipped, 0 failed**.

## Scope

Does not implement ticket 0340's standing class guard — that lands in its own PR.

## Review round 1

`/review-pr` returned **comment** — no blockers. All findings addressed in `9b86989b`:

- **Makefile prerequisites** — `tab_variables.md` and `codebook.md` declared `_deposit_variables.py`, which captured `markdown_cell` transitively until this refactor moved it; the two venue rules never covered it. All four now declare `scripts/_markdown_table.py`.
- **No oracle without pandoc** — the rendered tests are `integration`-tiered and skip wholesale where pandoc is absent, which in a repo with no CI left the escaping uncovered on such a machine. Added a fast-tier source-inspection guard over both emitters, verified by mutation.
- **Unescaped newline** — a raw newline ends a pipe-table row and no backslash can hold it. CR/LF now fold to a space. Only line breaks, so shipped rows do not churn.
- **`-f gfm` vs Quarto's `-f markdown`** — noted, 0 exposure today; not changed.
- **`[text](url)` from third-party metadata** — this is the documented and now test-pinned scope limit. Escaping brackets would alter the canonical `[missing]` value for no gain, since a bracket pair without a following `(` renders literally.

Two **pre-existing** gaps surfaced by the Makefile change, both caught by existing guards rather than by inspection:

- `_venue_naming.py` was missing from `build_analysis_archive.sh`, so the archived `export_core_venues_markdown.py` could not have imported it. Fixed.
- That guard parses `SCRIPTS=( ... )` with a non-greedy regex stopping at the first closing paren, so **any parenthesis in that array silently truncates the shipped script list**. My first comment tripped it. Worked around by keeping the array paren-free and documenting it in place; the brittle parser is left alone as out of scope for this ticket, and is worth its own ticket.

### Open design question for the author

The panel split on shape, not behaviour. Both reviewers agree the behavioural split is necessary. One holds that `markdown_cell(text, literal=True)` would give identical behaviour under a single name and satisfy the ticket's "one escaping implementation" more literally.

I kept two named functions: a boolean parameter that switches the input *contract* — and silently disables an exception — reads at the call site as a mode flag rather than as a different kind of input, and `markdown_text_cell(venue)` says at a glance what `markdown_cell(venue, literal=True)` makes you look up. Happy to collapse it if you prefer the single name. Either way the ticket's verification bullet is worth rewording to "one escaping *module*".

Related follow-up: tickets/0352-archive-script-list-guard-truncates-sile.erg — the archive script-list guard reads only a prefix of the SCRIPTS array; filed, not fixed here.


MinhHaDuong (commented):
## `/review-pr` — five-agent panel, verdict **comment**

No blockers. The fix is correct and mutation-verified by two independent reviewers; one finding (Makefile prerequisites) is a real regression this PR introduced and is a four-word fix.

**Gate run in the PR worktree** (this repo has no CI): `make lint` 166 passed / 6 skipped · `make check-fast` exit 0 · `make check` **7 failed, 1768 passed, 77 skipped** — the 7 are `test_corpus_acceptance.py` file-existence checks for DVC-managed catalogs not pulled into a throwaway worktree (`refined_works.csv`, `embeddings.npz`, …). Environmental, exactly as the PR body documents.

### Verified sound (so the author knows the ground was covered)

- **Escaping is empirically correct.** Adversarial inputs round-tripped through real `pandoc -f gfm -t html`: trailing backslash before the delimiter, pre-existing `\|`, doubled backslash before a pipe, unpaired backtick, empty string — all render as the original value with the correct cell count.
- **Mutation-verified, twice, independently.** Reverting `markdown_text_cell(...)` to the bare value in either emitter turns the matching test in `test_rendered_venue_table_fidelity.py` red with a genuine content mismatch (5 cells → 4, venue split at the pipe). Restored, green.
- **TDD is genuine, not claimed.** Commit `974f88c1` checked out in isolation fails both tests on a real content mismatch, not a fixture error. `28afd7d2` honestly documents replacing the vacuous `row.count("<td") == 5` assertion caught by adherence review.
- **No missed call sites.** Two reviewers swept the repo's pipe-table emitters independently and converged: `export_language_table.py`, `export_citation_coverage.py`, and `export_corpus_table.py` interpolate only curated label dicts, ISO codes, or computed numerics. 0339's exit criterion is met by this PR alone; deferring 0340's *standing guard* is clean, and 0340 declares `Blocked-by: 0339`.
- **The extraction is behaviour-preserving.** `_GFM_TEXT`/`_GFM_CODE` and the split/translate logic are byte-identical to `origin/main:scripts/_deposit_variables.py`. `test_variables_table.py` (33 passed) and `test_rendered_table_fidelity.py` (9 passed) still exercise `markdown_cell` through the re-import.
- **Investigated, no defect:** NaN can't reach a cell (`value_counts()` drops it; the other emitter does `fillna("")` → `"[missing]"`); the LaTeX path is clean (pandoc re-escapes AST text nodes independently — `a\\b` → `a\textbackslash b` under both readers, no double-escape); zero imports in the new module, no circularity; `markdown_cell`'s sole caller `render_codebook()` iterates hardcoded `Variable(...)` literals, so no corpus text can reach the raising function; mypy and ruff clean; test tiering correct.

---

### Findings

**1. `verifiable:` — Makefile prerequisites no longer cover the escaping rule.** *(doc propagation; high confidence; confirmed independently)*

`Makefile:283` and `Makefile:288` declare `scripts/_deposit_variables.py` as a prerequisite for `tab_variables.md` and `codebook.md`. Those rules used to capture the escaping rule transitively, because `markdown_cell` lived in that file. After the move they do not — editing `scripts/_markdown_table.py` rebuilds neither target:

```
283: deliverables/_shared/tables/tab_variables.md: scripts/figures/export_variables_table.py scripts/_deposit_variables.py
288: deliverables/_shared/tables/codebook.md: scripts/figures/export_codebook.py scripts/_deposit_variables.py $(EXTENDED)
272: deliverables/_shared/tables/tab_venues.md: scripts/figures/export_tab_venues.py scripts/utils.py $(REFINED) ...
330: deliverables/_shared/tables/tab_core_venues_top10.md: scripts/figures/export_core_venues_markdown.py scripts/analysis/summarize_core_venues.py scripts/utils.py $(MOSTCITED)
```

The convention is live in these very rules (`scripts/utils.py`, `scripts/_deposit_variables.py` are declared), so this is a gap the refactor opened, not a pre-existing habit. Adding `scripts/_markdown_table.py` to all four closes it. Related pre-existing gap, not this PR's fault: `_venue_naming.py` is imported by `export_core_venues_markdown.py` and is undeclared at `:330`.

**2. `verifiable:` — without pandoc, the only oracle for this fix silently disappears.** *(red team; high confidence; reproduced)*

`tests/test_markdown_table.py` covers the helper, not the emitters. `tests/test_rendered_venue_table_fidelity.py` is the sole end-to-end test and `pytest.skip`s on `shutil.which("pandoc") is None`. Reproduction: both emitters mutated to interpolate the raw value, pandoc shadowed off PATH →

```
uv run python -m pytest tests/test_markdown_table.py tests/test_rendered_venue_table_fidelity.py -q
  → 9 passed, 2 skipped, 0 failed        (pandoc hidden, emitters broken)
  → 2 failed "the venue split the row"   (pandoc restored, same mutation)
```

`make check-fast` deselects `integration` regardless of pandoc, and pandoc is declared nowhere — absent from `pyproject.toml`, `Makefile`, and `uv.lock` (verified). With no CI in this repo, a machine without pandoc degrades to green rather than to a warning. A fast-tier source-inspection test asserting each emitter's interpolation site calls `markdown_text_cell(` — this repo's own convention for CLI-flag presence — would close it for the cost of six lines. Arguably 0340's job rather than this PR's; worth deciding explicitly rather than by default.

**3. `verifiable:` — the rendered-fidelity harness is now duplicated across two test files.** *(consistency; high confidence)*

`_render_row` (`tests/test_rendered_venue_table_fidelity.py:49-61`) and `_cells` (`:64-74`) re-implement logic already inline in `tests/test_rendered_table_fidelity.py` — the pandoc invocation at `:119-120` and `:162-163`, the `re.sub(r"\s+", " ", …)` + `<tr>` findall at `:122-123` and `:164-165`. The two files also diverge in idiom for the identical pandoc check: a module-scoped fixture in the new file, three inline `shutil.which` skips in the old. The ticket names the old file as "the pattern"; sharing one helper module would honour that more than re-deriving it.

**4. `consider:` — two functions vs. the ticket's "one escaping implementation". Dissent preserved.** *(scope + consistency)*

The contract argument holds on the merits, and both reviewers said so: reusing `markdown_cell` on a journal name carrying two literal backticks would silently typeset half the name as code, and its `ValueError` would fail a manuscript build on one stray corpus character. Neither defect is hypothetical for free text.

The dissent is about shape, not behaviour. Scope reviewer: `markdown_cell(text, literal=True)` yields the identical behavioural split under one exported name and satisfies `tickets/0339-…erg:83` ("One escaping implementation, imported by all three call sites") literally, where two functions and three translation tables do not. PR body: splitting by input contract is the clearer boundary and the escaping rules still live in one module. Author's call — but the ticket's verification bullet should be reworded to match what shipped rather than left silently unsatisfied.

**5. `consider:` — the test oracle's reader is not the manuscript's reader.** *(red team; medium confidence)*

Tests assert against `pandoc -f gfm`; Quarto renders `manuscript.qmd` with pandoc's `markdown` reader (no `from:` override in `deliverables/manuscript/_quarto.yml`). They agree on every character the corpus currently carries, and diverge only for CommonMark type-1 raw-HTML-block starters (`script`/`pre`/`style`/`textarea`):

```
echo '| Efficiency | Journal<script>x</script>Name | 5 |' | pandoc -f gfm      -t latex  → JournalxName
echo '| Efficiency | Journal<script>x</script>Name | 5 |' | pandoc -f markdown -t latex  → Journal
```

Zero live exposure: the `journal` column has 0 values containing `<` or `>` across 23,975 non-null entries. Pointing the test's render call at `-f markdown` would make the oracle the real pipeline.

**6. `consider:` — embedded newline or CR breaks the row and is not escaped.** *(red team; medium confidence)*

`markdown_text_cell` escapes `\`, `|`, and backtick only. A literal `\n` in a venue value splits the GFM row into two rows (fabricating a phantom row); `\r` merges words with no space. Reproduced under both readers. Zero current exposure — `refined_works.csv` has no such value, since it would need a quoted multi-line CSV field. Same defect class the ticket is about, one character over; a whitespace normalisation before escaping closes it cheaply.

**7. `consider:` — unescaped `[text](url)` in a venue name becomes a live hyperlink in the PDF.** *(red team; low confidence)*

`pandoc -f markdown -t latex` renders `Journal [Foo](http://evil.example)` as `Journal \href{http://evil.example}{Foo}`. This is the documented, deliberate scope limit and the reasoning for it (not churning shipped rows) is sound. Flagged only because the venue field is third-party bibliographic metadata rather than authored content — one line in `_markdown_table.py`'s docstring naming that trust boundary alongside the existing rationale.

**8. `consider:` — documentation density inverts the sibling the module cites as its pattern.** *(consistency + scope, converging)*

`scripts/_markdown_table.py` is 89 lines: ~15 executable, the rest docstring and comment (~70% non-code). `scripts/_venue_naming.py`, cited at `_markdown_table.py:10-13` as the precedent, is 78% code. Separately, `_split_spans()` (`:47-52`) is a new private helper with exactly one caller, extracted during what the ticket specified as a lift — behaviour-preserving, but churn on moved code rather than reuse.

**9. `nofollow:` — ticket checkboxes unticked.** All five boxes at `tickets/0339-markdown-pipe-table-emitters-interpolate.erg:80-93` are still `- [ ]`; the 0325 precedent ticks them before close. `erg-pr-merge` closes regardless.

**10. `nofollow:` — committed `tab_venues.md` drift.** Pre-existing (last regenerated `9973559c`, 2026-07-22), disclosed in the PR body, and identical under both old and new code — its input `tab_pole_papers.csv` is a gitignored derived intermediate. Leaving it is correct scope discipline; refreshing it here would be unrelated corpus churn.

---

**Recommendation:** fix finding 1 before merge (four words in four Makefile lines, and it is a regression this PR created). Decide finding 2 explicitly — here or in 0340. Everything else is author's-call polish.

*Panel: correctness · consistency · scope · red team · doc propagation.*

MinhHaDuong (commented):
## `/review-pr` — PR 1152 (`t0339-markdown-cell-escaping`), fresh independent pass over current tip

Five-perspective panel run directly (correctness, consistency, scope, red-team, doc-propagation). Findings below are verified against the actual diff and a live run in the PR worktree, including a mutation check and rendered-output probing beyond the PR's own test suite.

### correctness — approve

- `markdown_text_cell` is complete for its stated scope: backslash, pipe, backtick escaped via a single `str.maketrans` pass (no re-escaping of its own substitutions — confirmed by reading `_GFM_LITERAL`), CR/LF folded to a space. Unit tests (`test_markdown_table.py`) pin each rule individually, including the ordering-sensitive case (`r"a \| b"` → `r"a \\\| b"`, i.e. the backslash is escaped *before* the pipe it would otherwise absorb).
- Mutation-verified live (not just re-reading the PR's claim): reverted `export_tab_venues.py`'s `markdown_text_cell(r['journal'])` back to bare `r['journal']`, reran the suite — both the rendered oracle (`test_rendered_venue_table_fidelity.py`) and the fast-tier source-inspection guard (`test_markdown_table.py::test_emitter_escapes_its_free_text_cell`) go red with the exact column-shift failure the ticket describes (`International Development Policy` lands in the Journal column, the split-off `Revue internationale` lands in Papers/Efficiency, the last column is dropped). File restored, diff clean afterward — this is not a vacuous test.
- `markdown_cell`'s extraction out of `_deposit_variables.py` preserves behavior for the codebook call site: the function moved to `_markdown_table.py` verbatim (byte-for-byte logic, only the docstring grew), `_deposit_variables.py` imports it at module level so `from _deposit_variables import markdown_cell` (used by `tests/test_variables_table.py`) still resolves via Python's normal name-binding — confirmed by running that file's `TestCodebookEscaping` class, all passing.
- Full gates in the PR worktree: `make lint` → **168 passed, 6 skipped**; `make check-fast` → **1168 passed, 11 skipped**. Ran the two new test files explicitly (pandoc is present on this machine, so the `integration`-tier rendered tests actually executed, not skipped): **15 passed** (`test_markdown_table.py` + `test_rendered_venue_table_fidelity.py`).

### consistency — approve

- `_markdown_table.py` follows the `_venue_naming.py` flat-module pattern verbatim: no dependencies, flat `_`-prefixed name, module docstring explaining the module-choice rationale, pure functions with `str`→`str` type hints. Added to `TestTypingCoreModules`'s mypy list alongside its sibling, matching precedent.
- Two functions instead of a `literal=` boolean flag is the right call for the stated reason (contract, not error-policy, differs) — checked this isn't just assertion: `markdown_cell` on a payload with balanced backticks treats the interior as a code span (only the pipe escaped inside); `markdown_text_cell` on the identical payload escapes the backticks too. Verified both branches pass (`test_balanced_backticks_are_a_code_span_only_for_authored_markdown`).

### scope — approve, with one nofollow note

- All 8 code/test files trace to ticket 0339's exit criteria (lift `markdown_cell`, route both emitters through the shared module, one rendered-oracle test, one unit-test file, Makefile prerequisite repair, mypy list entry).
- `build/build_analysis_archive.sh` — in scope: adding `_markdown_table.py` to the archived-scripts `SCRIPTS` array is required for the archive's `export_core_venues_markdown.py` / `export_tab_venues.py` copies to still import successfully post-refactor. The `_venue_naming.py` addition alongside it fixes a **pre-existing** omission on the same array the PR was already touching (that file imports `canonical_venue`/`venue_type` and was never archived) — legitimate same-line fix, not overflow, and stated as such in the PR body.
- `tickets/0349-archive-script-list-guard-truncates-sile.erg` — in scope as a *ticket filing*, not a code fix: it documents a real defect (`SCRIPTS=\((.*?)\)` non-greedy regex silently truncates at the first paren) surfaced by this PR's own workaround, correctly deferred to its own ticket per the "monster ticket" / tracking-ticket convention rather than fixed inline here. No code touches the guard in this diff.
- nofollow: the in-place comment in `build_analysis_archive.sh` ("keep this array free of parentheses") is a tripwire that depends on the next editor reading it — acknowledged by the author as exactly that, and ticket 0349's Action 3 already schedules its removal once the parser is fixed. Not this PR's problem to solve twice.

### red-team — comment (no blockers found; one residual gap already accepted by the author, confirmed by direct testing)

Probed beyond the PR's own test suite, against a live pandoc (`3.1.3`, `-f gfm`):

- `[text](url)`-shaped free text: confirmed by direct render that `Journal [of Economics](not-a-url) Studies` renders as an actual `<a href="not-a-url">` hyperlink inside the venue cell — not a crash, not a column split, but a silent typeset difference. This matches the PR's own documented scope limit, but the PR body's justification ("a bracket pair without a following `(` renders literally") is narrower than the real risk: a bracket pair *followed by* a paren, however unlikely in a journal name, does misrender. consider: narrow residual risk given actual journal-name shapes in the corpus (bilingual pairs joined by `|`, not `[x](y)`), already accepted by the author in round 1— no action requested, flagging for the record.
- NBSP (U+00A0), tab, Unicode line/paragraph separators (U+2028/U+2029): tested directly — none corrupt the row (GFM's line-ending definition is CR/LF/CRLF only, not U+2028/2029, and NBSP/tab are ordinary non-whitespace-collapsed characters to the pipe-table parser). No defect; `_GFM_LITERAL`'s scope (backslash, pipe, backtick, CR, LF) is complete for GFM's actual escaping needs. nofollow.
- HTML tags (`<b>...</b>`) in a venue name: not escaped, would pass through as literal HTML in the rendered manuscript (Quarto trusts embedded HTML in Markdown by default). consider: same class of gap as the bracket case, same-order likelihood in real journal names (near zero); not in the ticket's stated scope of backslash/pipe/backtick.
- `-f gfm` vs Quarto's actual `-f markdown` reader: already flagged in round 1, author explicitly left it unresolved ("noted, 0 exposure today; not changed"). Pandoc's inline-escape parsing (backslash escapes, code spans) is shared code across both readers, so the divergence risk is low, but this is not something I independently re-verified against `-f markdown` output — consider: worth a follow-up if `tbl-venues` is ever rendered through a materially different pandoc reader configuration.
- Two-function split / call-site contract check: grepped every importer of `markdown_cell` and `markdown_text_cell` repo-wide (excluding tests). Three call sites total, none on the wrong contract: `_deposit_variables.py` → `markdown_cell` (correct — codebook descriptions are authored Markdown), both venue emitters → `markdown_text_cell` (correct — raw corpus free text).
- CRLF-in-CSV: `pd.read_csv`'s C parser preserves an embedded `\r\n` inside a quoted field as literal characters, so the newline-folding path in `markdown_text_cell` is reachable, not dead code, for a `journal` value containing one.

### doc-propagation — approve

- Makefile: all four targets that transitively depended on `markdown_cell` (`tab_variables.md`, `codebook.md`) or now depend on `markdown_text_cell` (`tab_venues.md`, `tab_core_venues_top10.md`) declare `scripts/_markdown_table.py` as a prerequisite. Verified by reading the diff hunks directly, not by re-stating the PR body.
- `build/build_analysis_archive.sh`: `SCRIPTS` array updated (see scope section).
- Grepped repo-wide for stale references to `markdown_cell`'s old location (`_deposit_variables.markdown_cell`) in `docs/`, `deliverables/`, `tickets/` — only generated artifacts and ticket cross-references turned up, no doc text describing the old module layout that needs updating.

### Synthesis

No blockers. No dissent among perspectives — red-team's residual gaps (link/HTML syntax, gfm-vs-markdown reader) are the same ones the author already surfaced and consciously scoped out in the PR body's round-1 section; independent testing here confirms the gaps are real but narrow, and does not raise their severity above "consider."

---

```
review-pr: approve
blockers:
  - none
minors:
  - tag: consider
    file: scripts/_markdown_table.py:82 (markdown_text_cell)
    finding: a venue shaped like "X [label](note) Y" renders as a live hyperlink instead of literal text — confirmed by direct pandoc render; narrower in practice than the PR's own stated justification, but no corpus evidence of this shape today.
  - tag: consider
    file: scripts/_markdown_table.py:82 (markdown_text_cell)
    finding: embedded HTML tags in a venue name pass through unescaped and would render as live HTML in the manuscript; same likelihood order as the bracket case, out of the ticket's stated scope.
  - tag: nofollow
    file: tests/test_rendered_venue_table_fidelity.py (oracle uses `pandoc -f gfm`)
    finding: Quarto's manuscript render uses pandoc's default `-f markdown` reader, not `-f gfm`; already flagged and consciously deferred by the author in round 1, not independently re-verified here against `-f markdown` output.
  - tag: nofollow
    file: build/build_analysis_archive.sh:67
    finding: the "keep this array free of parentheses" in-place comment is a tripwire depending on the next editor reading it; ticket 0349 already schedules its removal once the parser is fixed.
gates: make lint 168 passed, 6 skipped | make check-fast 1168 passed, 11 skipped
dissent: none
```

PR #1160, opened 2026-07-27
fix(0343): move project secrets to the keystore, select them via KEYS=

MinhHaDuong:
**Ticket:** tickets/0343-move-project-secrets-to-the-keystore-and.erg

Removes the last three secret literals from `.env` (`AGENT_GH_TOKEN`, `OPENALEX_API_KEY`, `S2_API_KEY`), so no credential sits at rest in the repository directory.

## What the inventory found

All three values were **already** in `~/.config/keys/{github,openalex,semanticscholar}.env`, and hash comparison showed every keystore copy **identical** to its `.env` literal. No divergence to reconcile, nothing to copy — the migration is a deletion plus a `KEYS=` extension. Every keystore value is unique across the store; each credential lives in exactly one file, mode 0600 (dir 0700).

## Three entry points, one defect, found one at a time

Moving keys out of `.env` broke every path that fed them via `uv run --env-file .env`. Each was found only after fixing the previous one, which is the honest shape of this PR:

| Entry point | Symptom | Fix |
|---|---|---|
| `make …` | all four credentials absent in a plain terminal (`~/.bashrc` never loads the loader) | Makefile exports `BASH_ENV`, `SHELL` moved up beside it |
| `bash scripts/run_corpus_pipeline.sh` | documented usage got no `BASH_ENV` | script loads the keystore itself |
| `dvc repro` | picks its stage shell from `$SHELL` or `/bin/sh`; the Makefile's `SHELL` is make-internal and unexported | `scripts/pipeline_keystore.py`, applied at `pipeline_loaders` import |

The failure mode is why this mattered: a missing key reads as "use the free tier", so an unauthenticated multi-hour harvest looks like it is working. `docs/roadmap-datapaper.md` tells the author to run exactly that, two lines below the keystore instruction.

The Python resolver is the general fix — it sits at the one import every entry point already performs, never overwrites an already-set variable (so it composes with the bash loader), parses values literally (`K=$(echo pwned)` stays eleven characters), validates provider names before they index a path, and refuses protected destinations.

## Verification

- Every credential resolves and **matches its keystore source by hash**, from the repo root and a worktree, on the make, script, and dvc paths. No value or hash was ever printed — names and verdicts only.
- `gh api user` authenticates with the resolved `AGENT_GH_TOKEN`.
- `make lint`: 173 passed, 6 skipped. `/verify-adherence`: PASS.
- `make check` **pre-merge: 1807 passed, 39 skipped, exit 0** with the corpus reachable.
- `make check` post-merge (64 commits of `origin/main` integrated): 1847 passed, 3 failed — all three **data-state, not this branch**, which touches no corpus or analysis code:
  - two `test_corpus_acceptance` freshness failures — a Phase-1 rebuild is running in the shared primary checkout right now (`refined_works.csv` moved 15:29 → 16:09 during these runs) while `refined_embeddings.npz` / `refined_citations.csv` remain at 13:21;
  - `test_bimodality_dip` — a test that arrived with the merge, failing against `tab_bimodality.csv` dated 24 July, i.e. a stale derived table predating the diptest fix.

  Per AGENTS.md these belong to main and warrant their own ticket.

## Guards added

`tests/test_env_has_no_secret_literals.py` (7) and `tests/test_pipeline_keystore.py` (13). Both hardened after review broke the first versions:

- `.env` assignments are a known non-secret setting or `KEYS=`; no value carries a credential-shaped token; `KEYS=` selects every credential the code reads.
- Stale-claim scan over **every tracked text file** via `git ls-files` — the first version used a hardcoded list and missed the identical claim in `dvc.yaml:3`, on the branch adding the guard. Auto-discovery caught it immediately.
- Shape check now splits values into tokens and scans the `KEYS=` line too — previously a token hidden in a path or appended to `KEYS=` passed everything. Seven synthetic keys must be caught bare, in a path, appended to `KEYS=`, and space-surrounded; the live non-secret settings must not trip.
- The Makefile keeps both halves of the `BASH_ENV` wiring.

**These report variable NAMES only.** A guard that echoed the credential it found would reproduce the defect the ticket exists to prevent.

## For the author

- **Prose needs your arbitration**: `deliverables/_shared/_includes/agentic-workflow.md` and `deliverables/agentic/agentic-paper.qmd` claimed `.worktreeinclude` copies `.env` "so scripts find their API keys". False now; corrected minimally, wording is yours.
- **`HDMX-coding-agent` is not a GitHub account** (`gh api users/HDMX-coding-agent` → 404). It is a git author name; `AGENT_GH_TOKEN` authenticates as **`MinhHaDuong`** with full `repo`/`workflow`/`gist` scopes. Pre-existing — the token is byte-identical before and after. Corrected in this PR where the claim was false in prose (`.claude/rules/git.md`, the data-management plan, the agentic include). *An earlier version of this body said that was kept out of scope; commit `84322d60` landed it, and this line corrects that.* Not changed: `check-reviews.sh` still lists the login, now with a comment recording that it matches nothing, so the merge gate rests on self-review. Whether to provision a real machine account is your call.
- **Consequence documented, not fixed**: the loader re-applies `.env` per recipe, so `CLIMATE_FINANCE_DATA=/other make corpus` is now ignored in favour of `.env`'s value.
- Out of scope per the ticket: rotating the two exposed keys (yours) and the loader itself. `.claude/skills/update-publist/SKILL.md` still sources `HAL_ID`/`HAL_PASSWORD` from `.env` — same defect class, worth a follow-up.


MinhHaDuong (commented):

All five reviewers ran on `t343-keystore-migration` @ `origin/main...HEAD`. Tests re-run by the synthesizer in the PR worktree: `make lint` → **171 passed, 6 skipped**; `make check-fast` → **1147 passed, 11 skipped**. No reviewer printed a credential value; all findings below are names, line numbers, lengths, and verdicts only.

Panel verdicts: correctness `request-changes` · consistency `request-changes` · scope `comment` · red team `request-changes` · doc propagation `comment`.

The migration mechanics are sound — the parser/loader parity check holds, the `SHELL` relocation is behaviourally inert, the re-entry guard is real, and worktree path resolution is correct. What follows is what three or more reviewers converged on, plus the two bypasses the red team actually executed.

---

### 1. `verifiable:` — `dvc.yaml:3` still carries the stale claim, verbatim, and the new ratchet cannot see it — **high confidence, consistency**

`dvc.yaml:3-4` is byte-identical to the pre-PR `Makefile` comment this PR replaced:

```
# Env policy: secrets sourced from project .env via uv --env-file;
# never via export KEY := $(shell ...) or command-line KEY=value.
```

It contains `"secrets sourced from project .env"` — a string already pinned in `STALE_ENV_CLAIMS` (`tests/test_env_has_no_secret_literals.py:187`). The guard misses it only because `dvc.yaml` is absent from the hardcoded `MECHANISM_DOCS` list. Reproduced:

```
$ git grep -n "secrets sourced from project .env" -- . ':!tickets' ':!tests'
dvc.yaml:3:# Env policy: secrets sourced from project .env via uv --env-file;

$ uv run python -m pytest tests/test_env_has_no_secret_literals.py -q
5 passed
```

`dvc.yaml`'s own header says it "Mirrors Makefile Phase 1a–1f", so it was the file most likely to drift and it did. This is the exact failure the ratchet exists to prevent, sitting inside its blind spot on the very PR that introduces it.

### 2. `verifiable:` — the `MECHANISM_DOCS` hardcoded list, and the stated reason for keeping it hardcoded — **high confidence, consistency**

The docstring (`tests/test_env_has_no_secret_literals.py:193-195`) justifies the hardcoded list because a repo-wide walk "would sweep `.venv` and the ticket archive". Half of that is not true: `.venv` is gitignored and `git ls-files | grep -c '^\.venv/'` returns `0`, so a `git ls-files`-driven walk never sees it. Only `tickets/` needs excluding, and that is one line.

This repo has been bitten by exactly this before — memory `feedback_autodiscovery_class_guard`: an auto-discovering guard caught 5 offenders a manual sweep had missed. Concrete replacement:

```python
tracked = subprocess.run(["git", "ls-files"], cwd=PROJECT_ROOT,
                         capture_output=True, text=True).stdout.splitlines()
candidates = [f for f in tracked
              if not f.startswith("tickets/")
              and f != "tests/test_env_has_no_secret_literals.py"]
```

with a non-empty-candidates assertion so a discovery regression fails loudly. This would have caught finding 1 automatically.

### 3. `verifiable:` — two executed bypasses of the credential-shape guard — **high confidence, red team**

Both run against a copy of the parser on `/tmp` fixtures (invented strings, no real values):

**(a) One separator character exempts a whole value.** `PATHISH` (`:83`) exempts any value containing `/ \ @ space tab` *anywhere*, and `CREDENTIAL_PREFIXES` uses `startswith`, not substring. All three of these pass both guarding tests:

```
CLIMATE_FINANCE_DATA=/data/AKIAFAKEEXAMPLE1234567890ABCDEFabcdefghijklmnop   → PASS PASS
AGENT_GIT_NAME=HDMX Agent sk-liveFAKEexampleSECRETvalue1234567890            → PASS PASS
AGENT_GIT_EMAIL=agent+sk-liveFAKEexampleSECRETvalue1234567890@example.com    → PASS PASS
```

`CLIMATE_FINANCE_DATA` is the sharp case: it is *documented* to hold a path, so a `/` is expected rather than suspicious. Correctness independently reached the same conclusion from the other direction — a 40-char base64 secret access key routinely contains `/` and would sail through.

**(b) The `KEYS=` line's value is unconditionally exempt** (`:141`). A secret pasted as an extra comma-entry onto the one line every future `.env` edit touches is invisible to all five tests:

```
KEYS=openrouter:OPENROUTER_API_KEY,sk-livesecretexampleFAKEVALUE1234567890
→ test_env_assignments_are_allowlisted_or_keys: PASS
→ test_env_values_do_not_look_like_credentials: PASS
```

`test_keys_line_selects_every_consumed_credential` only checks required names are *present*; it never rejects a malformed extra entry.

Neither bypass needs a hostile actor — an ordinary editing slip triggers each. Suggested: flag opaque *substrings* rather than exempting whole values on one separator, and shape-check each `KEYS=` comma-entry against `provider(:VAR|:SRC=DST)?` instead of skipping the line wholesale.

**Dissent worth preserving.** Scope judged tests 1 and 2 genuinely complementary (the name allowlist cannot catch a secret under an allowlisted name; only the shape check can). The red team's (a) shows the converse hole: under an allowlisted name, the shape check is the *only* line of defence, and it is the one that fails. Both are right; together they say the pair has a real gap rather than redundancy.

### 4. `verifiable:` — `HDMX-coding-agent` is not a GitHub account, and two of the three claims sit in lines this PR edited — **high confidence, scope + doc propagation, converging**

Confirmed independently: `gh api users/HDMX-coding-agent` → `404 Not Found`.

| file:line | status |
|---|---|
| `.claude/rules/git.md:9` | **edited by this PR**; source half corrected, "machine user `HDMX-coding-agent`" left standing |
| `deliverables/_shared/_includes/agentic-workflow.md:61` | **edited by this PR**; "a dedicated GitHub account (`HDMX-coding-agent`)" left standing — and this is paper-facing prose |
| `docs/data-management-plan.md:88` | adjacent to an edited hunk; doubly false — also "scoped PAT for CI/automation only" when the repo has no CI at all (AGENTS.md, ticket 0321) |

The PR body discloses this and offers to file a ticket "say the word". Per the escalation table (*pre-existing but touched → comment + new ticket*) the ticket should be filed now rather than conditionally, and it should cover all **three** occurrences — the PR's own disclosure names only two and misses the paper-facing one.

### 5. `verifiable:` — `scripts/run_corpus_pipeline.sh` now documents a guarantee its own primary usage form does not have — **high confidence, correctness**

The new comment (`:4-7`) says credentials "reach the environment through the keystore loader that the Makefile wires in via BASH_ENV". But `:13-14` documents two usage forms, the *first* being direct:

```
# Usage: bash scripts/run_corpus_pipeline.sh
#   (or invoked via `make corpus`)
```

Nothing in the script sets or sources `BASH_ENV` (`grep -nF BASH_ENV scripts/run_corpus_pipeline.sh` → the comment only). This branch's own commit `55e0819b` establishes the gap is live, not theoretical: *"`~/.bashrc` does not load the loader, so in a plain terminal all four credentials came back absent."* That commit fixed the `make corpus` path and left the direct path as documented-but-unwired. A padme operator following line 13 gets `dvc repro` harvesting OpenAlex and S2 unauthenticated, silently.

Either narrow the Usage comment to `make corpus`, or have the script fail loudly when a required credential is empty. The second is the better shape given this project's standing pattern of silent pipeline degradation at credential joints — and it also closes the `$(wildcard …)` graceful-degradation concern below.

### 6. `consider:` — silent unauthenticated harvesting is the failure mode on every missing-key path — **medium confidence, red team + correctness**

`scripts/pipeline_io.py:54`, `scripts/harvest/catalog_semanticscholar.py:89`, and `scripts/filter_flags_llm.py` all resolve via `os.environ.get(NAME, "")` with no error on empty. So the loader being absent (`$(wildcard …)` → empty), a dropped `KEYS=` entry, or finding 5's direct-invocation path all converge on the same outcome: a rate-limited, subtly-degraded corpus build with a green suite. This predates the PR (an absent key already meant empty string), so it is not a regression — but the PR's "graceful degradation" framing reads as a feature where it is closer to an unguarded edge. A `make corpus-preflight` that fails when any `REQUIRED_KEYS_EXPORTS` name resolves empty would close findings 5 and 6 together. Follow-up ticket, not a blocker.

### 7. `consider:` — `tests/test_check_reviews.py` hermeticity is defeated by `BASH_ENV`, but this is **not** a regression from this PR — **high confidence on the mechanism, red team; framing corrected by the synthesizer**

`run_hook()` builds a hermetic env (`tests/test_check_reviews.py:62-63`, `AGENT_GH_TOKEN = "fake-token"`) and spawns `subprocess.run(["bash", HOOK_SCRIPT], env=env)`. Reproduced — and note the cwd dependency, which the first attempt got wrong:

```
$ env -i HOME=$HOME PATH=$PATH AGENT_GH_TOKEN=fake-token bash probe.sh    # cwd=/tmp
SURVIVED: test-injected value intact
$ env -i … BASH_ENV=$HOME/.claude/scripts/bash-env.sh bash probe.sh       # cwd=/tmp
SURVIVED: test-injected value intact
$ env -i … BASH_ENV=$HOME/.claude/scripts/bash-env.sh bash probe.sh       # cwd=repo root
CLOBBERED: value differs from injected (len=40)
```

The loader resolves `.env` from `$PWD`, so the clobber fires precisely where the hook runs. `env=` on `subprocess.run` cannot prevent it — the override happens through bash's own startup mechanism.

**But the red team's "new regression" framing does not survive checking.** Pre-0343, `.env` carried `AGENT_GH_TOKEN` as a literal and the hook itself does `set -a; source .env` (`.claude/hooks/check-reviews.sh:21-25`), which clobbered the test's fake token identically. The PR changes the *route*, not the outcome. `tests/test_check_reviews.py` currently passes (10 passed) with a live token flowing into the subprocess it tried to isolate — pre-existing, worth its own ticket, not a reason to bounce this PR. Fix: unset `BASH_ENV` in the constructed env, or assert on resolved identity rather than trusting `env=`.

### 8. `consider:` — two test/loader parity gaps in `test_keys_line_selects_every_consumed_credential` — **medium confidence, correctness + red team**

The file's docstring makes parity with the loader an explicit design goal (`:88-91`). Two places it does not hold:

- **Bare `provider` form.** `:165-166` `continue`s on any entry without a colon, so a bare `github` contributes nothing to `exported` — yet at runtime it sources the whole provider file and *does* export `AGENT_GH_TOKEN`. A `.env` simplified to bare form would fail the test while auth works fine. Not triggered today (all four current entries use the colon form).
- **Repeated `KEYS=` lines.** The test *unions* selectors across every `KEYS=` line; bash assignment is last-wins. So a `.env` where a conflict-merge appended a second `KEYS=` line passes the test while the first line's exports are silently dropped at runtime — a false PASS, feeding straight into finding 6's silent degradation.

### 9. `consider:` — terminology split between the code and the papers — **medium confidence, consistency**

The two prose files use "key store" (two words): `deliverables/_shared/_includes/agentic-workflow.md:61`, `deliverables/agentic/agentic-paper.qmd:81`. Everything else in the repo uses "keystore" (one word): `Makefile:77`, `scripts/pipeline_loaders.py:53`, `scripts/run_corpus_pipeline.sh:6`, `.agent/runbooks/on-start.md:16`, `.claude/hooks/check-reviews.sh:17`, the test file (7×), and the ticket (5×). Since both prose files already need author arbitration, fold this in — "keystore" matches the mechanism's name and every other doc.

### 10. `consider:` — `docs/data-management-plan.md:89` discloses the credential-store path in a deposit-bound document — **low confidence, doc propagation**

The new line names `~/.config/keys/<provider>.env` (mode 0600) explicitly. The DMP is externally facing and slated for OPIDoR deposit. Saying "a per-provider key store outside the repository, mode 0600" conveys the same reproducibility-relevant fact without publishing the path. Author's call.

### 11. `consider:` — measured cost of the `BASH_ENV` wiring — **low confidence, correctness**

Timed 20 hermetic bash startups: **~51 ms/call with the loader vs ~4 ms without (≈12×)**, dominated by the four `env -i bash -c` subshells (one per `KEYS=` entry). The Makefile exports `BASH_ENV` globally and does not use `.ONESHELL`, so all ~195 recipe command lines pay it. Not a defect — the re-entry guard is verified sound (`bash -c` nested three deep: 0.129 s, no fork bomb) and the design is deliberate. Recording the number so whoever next investigates build-time regressions does not rediscover it.

### 12. `nofollow:` — PR body says the doc-claim guard covers 9 files; `MECHANISM_DOCS` lists 10. Cosmetic.

### 13. `nofollow:` — reviewed and cleared

- **Makefile change in scope** (scope, high): it satisfies the ticket's own exit criterion ("credentials resolve for both the repo root and a worktree"), and Action 5's verification step is exactly what surfaced the regression. Leaving `make` broken would be worse. The `SHELL := /bin/bash` relocation is inert — Make resolves `SHELL` before any recipe runs, and `grep -nF '$(shell' Makefile` finds zero real calls.
- **Ratchet polarity** (scope, high): both ratchets pin negative guards, never positive phrasings — compliant with `.claude/rules/writing.md`. `STALE_ENV_CLAIMS`'s `"not exported into every process"` entry is not speculative: `git log -S` shows it was a real overclaim fixed in `815afbea` on this branch.
- **Parser parity** (correctness, high): `_parse_env` traced line-by-line against `bash-env.sh` — whitespace, `export` prefix, identifier validation, quote stripping, multi-`=`, CRLF. No divergence found. This was the highest-risk area and it holds.
- **Re-entry, worktree resolution, `KEYSTORE_LOADER ?=` override** (red team): all cleared. The override is the same trust level `make SHELL=…` already grants, and there is no CI to exploit it through.
- **Unticked exit criteria** in `tickets/0343-*.erg`: matches project convention — `erg-pr-merge` closes at merge.
- **`uv run --env-file` precedence** (doc propagation, empirically confirmed): leaf-process CLI overrides do win (`CLIMATE_FINANCE_DATA=fromcli uv run --env-file .env …` → `fromcli`), and recipe-shell overrides are swallowed by the loader — so the Makefile's comment is accurate on both halves. `NJOBS` is a Make variable, unaffected.
- **Reproducibility archives** (`build/templates/*`): never referenced API keys before or after. No propagation gap. `README.md` setup likewise.
- **Test-file conventions**: `pytestmark = pytest.mark.adherence`, built-in generics, naming — all match `test_script_hygiene.py` / `test_phase_layout.py` precedent. `ruff check` clean.

---

### Suggested disposition

Blocking, and both are cheap: **finding 1** (`dvc.yaml` comment) and **finding 2** (replace `MECHANISM_DOCS` with `git ls-files` discovery, which subsumes 1 permanently). **Finding 3**'s two bypasses are the substantive ones — the guard is materially weaker than its docstring claims, and tightening it is a contained edit to the same new file. **Finding 4** needs the follow-up ticket filed rather than offered, covering all three occurrences. **Finding 5** needs one line changed either way.

Findings 6, 7, and 8 are genuine but pre-existing or latent; they belong in follow-up tickets, not this branch. 9–11 are the author's call.

This review posts comments only — no fixes applied, no gate. Merge remains the author's decision.

MinhHaDuong (commented):
## `/review-pr` panel — 5 perspectives (correctness, red team, consistency, scope, doc propagation)

**Verdict: REQUEST-CHANGES** on one finding. Everything else is comment-tier.

Local gates re-run in the PR worktree: `make lint` **173 passed, 6 skipped**; `make check-fast` **1147 passed, 11 skipped**. `tests/test_env_has_no_secret_literals.py` **7/7**. No panel agent printed a credential value; neither does this review.

The migration itself is correct and well-guarded. The `KEYS=` line selects all four consumed credentials, the keystore files are 0600, the `_parse_env` helper faithfully mirrors `bash-env.sh`'s strict parser, and `bash-env.sh`'s `env -i` extraction subshell means the new `export BASH_ENV` cannot re-source itself. The security prose is honest — `tests/test_env_has_no_secret_literals.py:13-20` states plainly that a selected credential still reaches every bash subprocess, and the ratchet guards the reverse overclaim too.

---

## Blocker — the third documented entry point still loses its credentials

**`dvc.yaml` + `docs/roadmap-datapaper.md:26` — a direct `dvc repro` now runs silently unauthenticated.**

This is the exact regression the PR diagnosed and fixed twice (Makefile `BASH_ENV`, and `scripts/run_corpus_pipeline.sh:27-31`), left unfixed on the third path. Three of five reviewers found it independently.

Every stage in `dvc.yaml` sources its environment the same way (`dvc.yaml:20,37,52,106,138,238,290,350,367`):

```
cmd: uv run --env-file .env python scripts/harvest/catalog_openalex.py --resume
```

Before this PR that worked standalone, because `.env` held the literals. It no longer does, and nothing else on that path applies the `KEYS=` selection. Reproduced hermetically from the PR worktree:

```
$ env -i HOME=$HOME PATH=$PATH bash -lc 'cd <worktree> && uv run --env-file .env python -c "..."'
BASH_ENV set? no
S2_API_KEY present: False
OPENALEX_API_KEY present: False
```

`make corpus` is fine — `run_corpus_pipeline.sh` pre-resolves the credentials before calling `dvc repro`, and they are inherited. But `dvc repro <single-stage>` has no wrapped equivalent, and DVC picks its stage shell from `os.getenv("SHELL") or "/bin/sh"` (`dvc/stage/run.py:get_executable`) — the Makefile's `SHELL := /bin/bash` is a make-internal variable and is **not** exported into the recipe environment, so nothing about the Makefile fix reaches a standalone `dvc repro`.

Two things make this worse than a latent gap:

1. **This PR edits the doc that gives the broken instruction.** `docs/roadmap-datapaper.md:24-26` now reads "Put `S2_API_KEY` in `~/.config/keys/semanticscholar.env` and select it from the `KEYS=` line" — then, unchanged, "Run: `dvc repro catalog_semanticscholar`". Following those two lines in a plain terminal produces an unauthenticated harvest.
2. **`dvc.yaml:3-6`'s new comment asserts the opposite of what happens.** "Credentials … reach the environment through the keystore loader" is true only via `make corpus`, and the file's own header (`dvc.yaml:8-13`) documents bare `dvc repro` as a first-class entry point on hours-long stages.

Fix, either way: source the loader in the stage `cmd:`s (or a thin wrapper), **or** state the prerequisite plainly in `dvc.yaml`'s header and at `roadmap-datapaper.md:26`, and add a fail-fast credential check before `dvc repro` burns hours. Related — `scripts/run_corpus_pipeline.sh:74` still has no assertion that the credentials actually resolved; `bash-env.sh` warns and continues on a missing keystore file, so a misconfigured store lets the full pipeline complete unauthenticated.

---

## Comment tier

**`verifiable:` `.claude/hooks/check-reviews.sh:70,73` — the corrected claim recurs in the one file it wasn't fixed in.** Commit `84322d60` proved `HDMX-coding-agent` is not a GitHub account (`gh api users/HDMX-coding-agent` → 404, independently confirmed) and corrected the "machine user" wording in three files. `check-reviews.sh` — edited by this PR at lines 17-19, three lines above — still carries `# HDMX-coding-agent: local machine user identity.` and `AGENT_LOGINS="HDMX-coding-agent MinhHaDuong"`. That is not just stale prose: `AGENT_LOGINS` is matched against the API's `user.login` at line 79 to gate merges, so that half of the allowlist can never match and the merge gate rests on `MinhHaDuong` alone (i.e. on self-review). Per the escalation table this is pre-existing-but-touched → comment + follow-up ticket. `STALE_ENV_CLAIMS` does not cover this claim class, so it has no regression protection.

**`verifiable:` No in-repo onboarding path for the keystore.** `README.md`, `AGENTS.md`, and `.env.example` have zero hits for `KEYS=`, `~/.config/keys/`, or `bash-env.sh`. `README.md:18-29`'s Setup covers `make setup` / `uv sync` / `dvc pull` and says nothing about credentials. Before this PR a reader could at least discover the (insecure) working path; now that path is correctly blocked by the guard and nothing points at the replacement — and the loader lives outside the repo. One sentence in Setup or the `.env.example` header would close it.

**`verifiable:` PR body contradicts its own diff on the machine-user correction.** The "For the author" section says the finding was "kept out to preserve PR scope", but `84322d60` landed those corrections in `.claude/rules/git.md:9`, `docs/data-management-plan.md:91`, and `deliverables/_shared/_includes/agentic-workflow.md:61`. The corrections are accurate and welcome — the body just needs to say so, since it is what a later reader will trust.

**`verifiable:` Guard shape-heuristic has reachable holes** (`tests/test_env_has_no_secret_literals.py:93-95,148-157`). `TOKEN_SPLIT` does not split on `-`, `_`, `.` or `+`, while `OPAQUE_TOKEN` requires the whole token to be `[A-Za-z0-9]+` **and** to contain both a letter and a digit. So a value evades the shape check if it is hyphen/underscore/dot-separated throughout, hex-only or letters-only, base64 with `+`/`/`/`=`, or under `OPAQUE_MIN_LENGTH = 24`; and `sk_live_…`-style prefixes miss `CREDENTIAL_PREFIXES` (which lists the hyphenated `sk-`). None of this bites the four credentials actually migrated — this is coverage for the next vendor's format, not a live gap.

**`consider:` `scripts/run_corpus_pipeline.sh:28` uses a single-credential sentinel.** The loader is sourced only when `OPENALEX_API_KEY` is empty; a partial ambient environment (OpenAlex set from elsewhere, S2 not) skips it and leaves the rest unresolved.

**`consider:` `test_keys_line_selects_every_consumed_credential` (line ~239) unions every `KEYS=` line**, but `bash-env.sh` keeps only the last assignment. A duplicated `KEYS=` line passes the test while a credential silently stops resolving.

**`consider:` `export BASH_ENV` (Makefile:99) widens exposure to recipes that had none.** `clean`, `setup`, `venv-canonicalize`, `manuscript` and several archive targets never touch `$(UV_RUN)`; before this PR they received no credential at all, now they get all four in the recipe shell. Low practical severity, but a real move away from least-privilege scoping. Relatedly, `make check` / `check-fast` run pytest with all four ambient both before and after — so the specific leak vector the ticket narrates (a test subprocess dumping its environment) is not closed by this change, only relocated at rest. The test file says this; the ticket's framing reads broader.

**`consider:` The Makefile `SHELL`/`BASH_ENV` change is a second architectural decision on a secrets ticket** (`Makefile:79-96`). It is justified by Action 5 and it does fix a real regression, but it silently changes `VAR=x make target` semantics repo-wide, not just for credentials. No current usage breaks (the one command-line `CLIMATE_FINANCE_DATA=` is a leaf `uv run`). Disclosed in the PR and in the Makefile comment — noting it so the decision is on the record rather than inside a migration.

**`consider:` Terminology drift.** Every code and comment site says "keystore"; the two prose passages (`deliverables/_shared/_includes/agentic-workflow.md:61`, `deliverables/agentic/agentic-paper.qmd:81`) say "key store". Consistent with each other, divergent from the project's term. Author's call — these are the two manuscript edits already flagged for arbitration.

**`consider:` `.claude/rules/architecture.md`** is cited by `Makefile:71` and correctly documents `CLIMATE_FINANCE_DATA` relocation, but says nothing about `BASH_ENV`/`KEYS=`. A one-line cross-reference would close the loop in the canonical pipeline doc.

**`nofollow:` `.claude/skills/update-publist/SKILL.md:77`** (`Credentials from .env: HAL_ID, HAL_PASSWORD`) — confirmed stale (neither name is in `.env` at all). Already flagged by the author as out of scope; same defect class, worth the follow-up.

**`nofollow:` `tickets/0316-phase-secrets-out-of-project-env.erg`** describes the superseded flat-`~/.claude/.env` design and has no `Closed:` header. Housekeeping, not this PR's job.

**`nofollow:` Currency.** Branch is 50 commits behind `origin/main` (7 ahead). Per the rebase-at-every-gate rule this wants a `git merge origin/main` before the merge gate — non-blocking, and the merge-base diff is unaffected.

---

## Dissent preserved

The **consistency** reviewer opened with a "critical" claim that the diff artifact under review was stale and included 38 unrelated files (a `diptest` revert, deleted tickets, a `.gitignore` reversal). **This is a false positive** — I verified the artifact contains exactly the 12 PR files (`grep -c '^diff --git' → 12`, matching `gh pr diff 1160 --name-only`); it was generated with three-dot `origin/main...HEAD`, which is merge-base-relative and immune to the staleness. Its underlying observation — the branch is far behind main — is true and is recorded above as a currency nit. All other findings from that reviewer were scoped to the true diff and stand.

The **scope** and **doc-propagation** reviewers both returned *comment*; **red team** returned *request-changes* on the DVC finding; **correctness** and **consistency** returned *approve*. The DVC finding is the one I could reproduce mechanically, which is why it carries the verdict.

---

*Panel: correctness / red team / consistency / scope / doc propagation, all `model: sonnet`. No fixes applied — `/review-pr` posts findings only; the merge decision is the author's.*

PR #1164, opened 2026-07-27
fix(0327): reconcile the data paper's corpus counts from a generated ledger

MinhHaDuong:
Closes the two arithmetic gaps, the two column-sum mismatches and the stale "six sources" that four external reviewers found in the RDJ-26561 data paper.

## The one cause behind both gaps

`corpus_audit.csv` carries **three** action buckets — `keep` (33,344), `remove` (9,436), `deduped` (399) — and every consumer counted two.

- `filter_net_removals` read `action == "remove"`, so the 399 duplicate-DOI records dropped after enrichment never entered the arithmetic. **Gap 2.**
- `filter_protected` read `action == "keep"` while `filter_flagged` was measured over all rows, mixing pre- and post-deduplication stages. 19 flagged works were protected by the filter and deduplicated afterwards, so `flagged − protected` came out 19 above the removal count it should equal. **Gap 1.**

The prose also mis-stated the mechanism: "21 content-duplicates discovered during enrichment" matched neither the count (399) nor the operation (a DOI dedup pass keeping the highest `cited_by_count`, `corpus_filter.py:326`).

## What lands

**A generated CONSORT-style ledger** (`scripts/analysis/compute_corpus_flow.py` → `tab_corpus_flow.csv`/`.md`, rendered as Table 3):

```
44,174 pooled source records
  − 833 DOI duplicates at merge          = 43,341
  −   3 records without a title          = 43,338
  − 159 title+year duplicates at merge   = 43,179
  − 9,436 quality filtering (protection applied) = 33,743
  −   399 duplicate-DOI after enrichment = 33,344 refined works
```

It enumerates the three buckets explicitly and refuses to write a ledger that does not close — an unmodelled fourth action is a hard error, not a silent exclusion. `filter_protected` is now stage-consistent (1,808), so 11,244 − 1,808 = 9,436 matches the table exactly.

**Table 2's TOTAL row** is labelled the deduplicated union, with both columns' overlap measured (748/773 raw, 738/763 refined, 25 and 25 carrying three provenances) — which is what reconciles the 763 extra memberships against §2.2's 738 multi-source works.

**"Six sources" → eight** in the paper, the archive README, the build script's catalog loop, and the ED-04 Zenodo upload runbook. Scoped deliberately: `manuscript.qmd`, `manuscript-Gide.qmd` and `slides-gide.qmd` say "six" correctly (pinned to the v1.0 corpus), so a repo-wide guard would have forced three wrong edits.

**R1-12's dedup accounting** is answered by the ledger in `response-letter.md`.

## Two build bugs the adherence gate surfaced

Under a grouped Make target (`a.csv a.md &:`), `$@` binds to the **requested** member. Both `export_corpus_table.py` and `compute_corpus_flow.py` derive the `.md` from `--output`, and `paths.mk` lists only the `.md` — so a data-paper render wrote the CSV to the `.md` path, overwrote it with markdown, and left the tracked CSV stale while make marked the group updated. Reproduced before fixing. Both recipes now name the `.csv`; both scripts reject a non-`.csv` `--output`; an adherence test fails any grouped rule with mixed-extension members using bare `$@`. The pre-existing instance in `tab_corpus_sources` is fixed too.

Also from the gate: the flow rule now declares the `catalog_merge` run report it reads; `tab_corpus_flow.csv` gets a Pandera schema; the deposit's table staging derives from the qmd's own `{{< include >}}` directives instead of a hand-kept list that had already gone stale; and the deposit's five drifted hand-typed counts are corrected and pinned to the vars.

## Verification

- `make check`: **1841 passed, 37 skipped, 3 failed.** All three failures are pre-existing and belong to open ticket **0344** (Phase-2 derived tables stale against the corpus): two `TestCorpusAlign` mtime assertions in the shared checkout, and `test_bimodality_dip` on a `tab_bimodality.csv` predating 0345's column. This branch never touches bimodality and reads corpus data read-only.
- `make check-fast` and `make lint` green. `/verify-adherence` run twice; every blocking finding fixed.
- Ledger verified in the rendered PDF, not only in the source.
- Both generated tables are byte-stable when regenerated on the rebased tree.

## Word budget — handed to 0332, not paid for by cutting §2

The document renders at **3,921 words** against the 2,500 RDJ4HSS target, of which this change adds ~110 (the flow table and the longer Table 2 caption; the arithmetic sentence it replaces was removed). Per this ticket's own exit criterion the overshoot is recorded in the ticket log and handed to 0332, which owns the budget and the word-count guard. §2 was not condensed.

## Author action outstanding

The Zenodo record itself must be retitled *Six* → *Eight Sources* at the pending New-version upload. The runbook now instructs it; an agent cannot change the live record.

Ticket-ref: tickets/0327-datapaper-counts-do-not-reconcile.erg

> The ticket is already closed and archived on this branch (`tickets/closed/0327-…`), so this is a reference, not a close claim — `erg-pr-merge` has nothing left to close and its extractor needs the digits immediately after `tickets/`.


MinhHaDuong (commented):

Substantial PR: 21 files, 827 additions, changing published numbers in a submitted data paper. All five perspectives ran. Findings are deduplicated and attributed; dissent between reviewers is preserved verbatim rather than resolved.

**Tests run on this branch** (detached at `790cff29`, in a review worktree — not the shared checkout, which sits on `main`):
- `make lint` — **185 passed, 6 skipped**
- `make check-fast` — **1232 passed, 8 skipped**
- `make check` — **1914 passed, 37 skipped, 1 failed**

The single failure is `tests/test_bimodality_dip.py::test_bimodality_dip_column_is_not_silently_empty` (`dip_pvalue is present but empty in every row — diptest is not installed`). Pre-existing, belongs to the open `diptest` optional-dependency ticket; this branch touches nothing in the bimodality path. The two `TestCorpusAlign` mtime failures the PR body reports did not reproduce in a fresh worktree — they were shared-checkout artifacts.

---

## Must resolve before merge (merge mechanics, not code)

**`erg-pr-merge` cannot process this PR's close claim.** — *scope*, confidence high, reproduced.

The branch pre-closes and archives the ticket (commit `492ec50f`), so at the branch tip the file is `tickets/closed/0327-datapaper-counts-do-not-reconcile.erg`, and the PR body's claim line points there. The helper's extractor requires digits immediately after `tickets/`:

```
$ printf '%s\n' '**Ticket:** tickets/closed/0327-datapaper-counts-do-not-reconcile.erg' \
    | grep -oiP '^\*{0,2}ticket:?\*{0,2}:?\s*tickets/\K\d+'
NO MATCH -> TICKET_IDS empty
$ printf '%s\n' '**Ticket:** tickets/closed/0327-x.erg' \
    | grep -qiP '^\*{0,2}ticket-ref:?\*{0,2}:?\s*tickets/\d+'
NOCLOSE unset -> die
```

`TICKET_IDS` comes out empty and no `Ticket-ref:`/`Ticket: none` marker is present, so the helper dies at `no close-claim in PR body` before reaching the merge. Fixing the path alone does not help: the existence check globs `tickets/${tid}-*.erg` and would then report 0327 absent from `tickets/` at the branch tip (confirmed — only `tickets/closed/0327-*.erg` exists there).

Since the close commit is already on the branch and the ticket is genuinely done, the cheapest correct fix is to replace the body's claim line with a **non-closing reference the extractor recognises**:

```
Ticket-ref: tickets/0327-datapaper-counts-do-not-reconcile.erg
```

That sets `NOCLOSE`, skips the redundant close, and lets the merge proceed. Alternatively `Ticket: none` with a prose pointer. No code change needed.

---

## Findings

### The published numbers are correct — independently confirmed

`verifiable:` — *correctness*, *red team*, *doc propagation*, confidence high. Three reviewers independently reconciled the full chain against live data (`corpus_audit.csv`, `catalog_merge__20260727T132143Z.json`, `unified_works.csv`, `refined_works.csv`):

```
44,174 −833 → 43,341 −3 → 43,338 −159 → 43,179 [= unified_works rowcount]
       −9,436 → 33,743 −399 → 33,344 [= refined_works rowcount]
```

`value_counts()` on the audit gives keep 33,344 / remove 9,436 / deduped 399, summing to 43,179. `filter_flagged (11,244) − filter_protected (1,808) = 9,436` — the identity the ticket demanded now holds exactly; the old `action == "keep"` definition gave 9,455, the reported 19-work gap. Table 2's naive column sums (43,952 / 34,107) less their TOTAL rows give 773 / 763 extras, matching the new caption. Re-running `compute_corpus_flow.py` against live data reproduces the committed `tab_corpus_flow.csv`/`.md` byte-identically.

### `tab_corpus_sources.csv`'s two moved cells — provenance question closed

`verifiable:` — raised by *doc propagation* (confidence medium: two `%non-EN` cells moved, 7%→8% and 7%→9%, with unchanged denominators and no generator change in the diff — flagged against the project's "verify vars file provenance" lesson). I resolved it directly:

```
$ PYTHONPATH=scripts:libs/openalex-corpus/src uv run --env-file .env \
    python scripts/figures/export_corpus_table.py --output /tmp/regen_sources.csv
$ diff deliverables/_shared/tables/tab_corpus_sources.csv /tmp/regen_sources.csv
Files are identical
```

The committed table regenerates byte-for-byte from the current corpus. Genuine regeneration output, not a hand edit. **No action.**

### The new grouped-target adherence guard is narrower than its docstring

`verifiable:` — *correctness* and *red team* converged on this independently, confidence high. `tests/test_corpus_flow.py:156-173` promises "No `&:` rule whose members differ in extension may use bare `$@`". Four gaps in the parser:

1. **Multi-line target lists are silently skipped.** `members = line.split("&:")[0].split()` reads only the physical line holding `&:`. `Makefile:423-425`:
   ```make
   $(DERIVED)/tab_alluvial.csv $(DERIVED)/cluster_labels.json \
   $(DERIVED)/tab_core_shares.csv &: \
   ```
   genuinely mixes `.csv`/`.json`, but the guard sees only `tab_core_shares.csv`, computes a single-extension set, and `continue`s — the recipe is never inspected. This rule is correct today (it names its output explicitly), so nothing ships wrong; the guard simply cannot see it, nor any future grouped target declared across lines.
2. **Same-extension groups are exempt** (`< 2` extensions → skip). `a.csv b.csv &: … --output $@` has the identical write-one-member-leave-the-other-stale failure mode this ticket exists to fix, and is not flagged. Latent, no such rule exists now.
3. **Fixed 5-line lookahead** (`lines[i+1:i+6]`) does not distinguish backslash-continued prerequisite lines (also tab-indented) from recipe lines, so enough continuation lines push a genuinely broken recipe out of the window.
4. **False positive** on a harmless `@echo "building $@"` line, which never passes `$@` to the script.

None of these produce a wrong number today — both live rules are correctly written and the fix itself is right. But the guard is the standing regression gate for exactly the defect class this PR discovered, and it would not catch a repeat.

### `compute_corpus_flow.py` sits on the wrong side of the compute/export split

`verifiable:` — *consistency*, confidence high. `.claude/rules/architecture.md` (Phase 2, rule 4): "Compute / Plot / Include are separate… Never mix." `docs/repo-layout.md` classifies `export_*` → `scripts/figures/`, `compute_*` → `scripts/analysis/`. The new script writes both `tab_corpus_flow.csv` (fine for `scripts/analysis/`) **and** `tab_corpus_flow.md` — a captioned, `{#tbl-flow}`-labelled Quarto include. All seven other producers of such a file (`export_corpus_table.py`, `export_tab_venues.py`, `export_language_table.py`, `export_codebook.py`, `export_citation_coverage.py`, `export_core_venues_markdown.py`, `export_variables_table.py`) live in `scripts/figures/`. This is the only `scripts/analysis/` script emitting a Quarto-includable `.md`.

### Contract-file read bypasses the mandated loader

`verifiable:` — *consistency*, confidence medium. `compute_corpus_flow.py:189` does `pd.read_csv(REFINED_WORKS_PATH, usecols=["source"])`. `architecture.md` Phase-2 rule 9: "Never call `pd.read_csv()` on contract files (`refined_works`…) directly. Use `pipeline_loaders`." The module is already imported for the path constant. The in-line comment frames the raw read as a deliberate cheap-count choice; legacy `analyze_*.py` offenders are grandfathered under migrate-as-touched, but this is new code.

### Run-report selection is lexicographic, not chronological

`consider:` — *red team*, confidence medium. `latest_merge_report()` does `sorted(glob.glob(pattern))[-1]`. `catalog_merge.py --run-id` accepts any free-form string (sanitised only by `re.sub(r"[^\w.-]", "_", …)`), while the default `make_run_id()` is `%Y%m%dT%H%M%SZ` — digit-leading. Any letter-leading manual run-id therefore sorts after every real timestamp and permanently masks the true latest report:

```python
sorted(['catalog_merge__20260701T090000Z.json',
        'catalog_merge__20260727T120000Z.json',
        'catalog_merge__manual-fix.json'])
# -> [..., 'catalog_merge__manual-fix.json']
```

`dvc.yaml`'s `catalog_merge` stage never passes `--run-id`, so `dvc repro` is safe; this bites a manual/debug invocation only. **Pre-existing pattern** (identical code in `compute_vars.py:366`), but this PR adds a second and higher-stakes consumer — the published reconciliation ledger — without hardening it. On the current data it happens to select the correct report.

### Cross-stage skew that closes silently

`consider:` — *red team*, confidence medium. `corpus_audit.csv` and `refined_works.csv` are written from the same in-memory frame inside one `apply_filter()` call, so the new "anchor outside the audit" check (commit `1e0ccbe6`) genuinely closes the all-`keep` dry-run hole it was written for — not tautological. The residual: `catalog_merge` and `corpus_filter` are independent stages, and check 2 (`sum(buckets) == records_unified`) only fires when the two runs' record *counts* differ. A same-count, different-composition pair would produce a wrong ledger that closes. Structurally possible, vanishingly unlikely with real harvests.

### Untested twin guard

`verifiable:` — *red team*, confidence high. `export_corpus_table.py:149-152` adds the same "reject non-`.csv` `--output`" check as `compute_corpus_flow.py:180-184`, but only the latter is pinned (`TestOutputPathContract::test_script_refuses_a_non_csv_output`). `tests/test_corpus_table_export.py` has no equivalent. Dead code on the live build path — both recipes now pass an explicit `.csv` — so this is a coverage asymmetry, not a live bug.

### Ticket log lines drop the author field

`verifiable:` — *consistency*, confidence high. `tickets/AGENTS.md` specifies `YYYY-MM-DDTHH:MMZ author verb detail`. Three of nine log lines in `tickets/closed/0327-*.erg` run straight from timestamp into prose (`13:06Z word budget: …`, `13:06Z author action carried by …`, `13:54Z post-rebase onto origin/main …`) where the other six correctly carry `HDMX-coding-agent`. `erg validate` passes — the grammar accepts any token as actor — so this is silent convention drift.

### Scope creep, individually defensible

`consider:` — *scope*, confidence high on the facts, judgment call on the verdict. Four things ride along beyond ticket 0327's five actions:

- The grouped-`$@` fix to the **pre-existing** `tab_corpus_sources` rule and the repo-wide adherence guard (the new rule's own fix is unavoidable — the script needs a grouped target).
- The archive table-staging refactor (`{{< include >}}`-derived instead of a hand-kept list), which also opportunistically fixes a pre-existing bug where `tab_variables.md` was never staged.
- The five drifted hand-typed deposit counts (`corpus_raw` 42,922→43,179, citations 968,871→1,375,310, embeddings 38,479→38,736 — all confirmed against `data-paper-vars.yml`). A real and valuable fix, but a different defect class from the "six sources" residue.
- A new test class `TestDepositCountsTrackTheCorpus` beyond the ticket's Test spec.

Each is well-explained in its commit message and low-risk. Commit `6d24d44f` bundles at least six distinct concerns in one commit, against one-change-per-commit.

### The "six sources" sweep is correctly scoped

`verifiable:` — *consistency* and *doc propagation* both swept independently and agree. No stale "six sources" survives in `data-paper.qmd`, `README-datapaper.md`, `build_datapaper_archive.sh`, or the ED-04 runbook; `config/*.yaml`, `dvc.yaml`, top-level `README.md` are clean. The three excluded files are genuinely v1.0-pinned (`manuscript.qmd:291`, `manuscript-Gide.qmd:202` with the six-source provenance list, `slides-gide.qmd:135` as its companion deck). The six-**flag** pipeline references are untouched and `scripts/filter_flags.py` still defines exactly six. `{{< meta filter_net_removals >}}` and `{{< meta corpus_removal_pct >}}` have zero surviving consumers branch-wide. Table renumbering checks out: the flow table renders as Table 3 and every in-body cross-reference uses a label.

Residue outside the PR's declared exclusions, both `consider:`, low severity:
- `docs/rdj-submission-checklist.md:4` still ticks `[x] Title: "…Six Sources…"` as done and `[ ] Update Zenodo deposit with v1.1 data files` — a version behind the v2.0 this PR prepares.
- `docs/eshet-hes-slides-outline.md:40`, `docs/research-note-field-structure.md:27` say "six sources" and are not covered by the v1.0-pin rationale. Planning docs, not deposit-facing.

Neither the manuscript exclusion nor these are mechanically pinned — `TestSourceCardinality` parametrises over `[QMD, README]` only, so the correct "six" in the three v1.0 files rests on a test docstring, not a guard. Given this PR *is* a six→eight sweep, that's worth noting.

### Dissent preserved: the prose sentence

Two reviewers reached different conclusions on the same line. Both are recorded; this is prose on a submitted revision, so it belongs to the author, not to a reviewer or an agent.

`deliverables/data-paper/data-paper.qmd:100`:
> …{{< meta filter_protected >}} of them meet a protection criterion and **survive the filter**.

**Correctness** (`verifiable:`, high) reads this as precise and clears it: "survive the filter" is scoped to the filtering *stage* (`action != "remove"`), not to the final corpus, and the very next line includes `@tbl-flow`, whose "Duplicate-DOI records dropped after enrichment" row shows some survivors are later removed. No factual error.

**Red team** (`consider:`, medium) accepts the literal accuracy but flags an information loss: under the new definition, 19 of the 1,808 counted works have `action == "deduped"` and are absent from the shipped 33,344-row corpus. The old number (1,789) directly answered "how many protected works are in the final corpus" — and that number is now stated nowhere, in the prose or the table. R1-12 asked for per-category accounting; the fix trades a wrong arithmetic identity for a lost figure a referee may specifically look for.

### Lower-severity notes

- `consider:` low — `CorpusFlowSchema` pins shape, dtype, sign domain and `Stage` uniqueness, but not `In − Removed == Out` nor the chaining. Deliberate per the adjacent comment (the arithmetic lives in `build_flow`); exposure is a hand-edited CSV that make won't rebuild. *(red team)*
- `consider:` low — `CorpusFlowSchema` requires `Out > 0`; a future stage legitimately reducing an intermediate to exactly 0 would fail validation. *(correctness)*
- `consider:` low — `main()`'s `pd.read_csv(audit_path, usecols=["action"])` raises a raw pandas `ValueError` if the column is missing, rather than one of the script's own guided messages. *(correctness, red team)*
- `consider:` low — the `.csv`-only `--output` check is duplicated verbatim in two scripts instead of extending `script_io_args.validate_io()` with an `expected_ext` parameter. *(consistency)*
- `consider:` low-medium — `format_md()` interpolates `Stage` labels directly into a pipe table without going through `scripts/_markdown_table.py`. Nothing misrenders today, and only 3 of ~9 emitters use the helper, so this continues an existing inconsistency rather than introducing one — but it is the ticket-0325 defect class. *(consistency)*
- `consider:` low — `membership_overlap` maps zero-provenance rows to the same bucket as single-provenance ones; unreachable by construction today, but a silently-absorbed integrity signal in a PR about silently-absorbed buckets. *(red team)*
- `nofollow:` — `tests/test_datapaper_archive_layout.py` sets `pytestmark = adherence` module-wide while one test additionally carries `@integration` and spawns a subprocess, blending tiers. Cheap subprocess, harmless. *(correctness)*
- `nofollow:` — the ticket's Verification checkbox "every §2.3 count is a `{{< meta >}}` macro" is not literally true: the flag-6 calibration paragraph still hand-types "100 works", "AUC = 0.818", "81% accuracy". Pre-existing, untouched, and calibration statistics the flow table could never supply. *(scope)*

---

### What held up under attack

Zero-match run reports raise a clear `FileNotFoundError` pointing at `dvc repro`. Dry-run audits (`would_remove`) are detected and rejected with an operator-facing message rather than silently misread. Case variants (`"Keep"`, `"remove "`), integer-coded actions and missing-action rows all raise `ValueError` naming the offending value. The three-way closing check is real defence in depth for the defect it targets. The archive's new `{{< include >}}`-derived staging list is a strict superset of the old hardcoded one — nothing dropped — and is backed by an integration test that runs the real discovery pipeline. The Zenodo retitle is a bolded item in the runbook's numbered author-steps list, not buried in prose. The word budget was handed to ticket 0332 as the exit criterion required, with §2 left uncondensed.

### Panel verdicts

| Perspective | Verdict |
|---|---|
| Correctness | comment |
| Consistency | comment |
| Scope | request-changes (on the merge-mechanics blocker) |
| Red team | comment |
| Doc propagation | comment |

**Synthesis: comment.** The science is sound and reproduces byte-for-byte against live data; no finding touches a published number. The one item that must be handled before merge is the PR body's close-claim line, which is an edit to this description rather than to the code. `/review-pr` posts comments only — the gate is `/verify-gate`'s call and the merge is the author's.

PR #1172, opened 2026-07-27
test(0340): standing guard — every generated Markdown table row has its declared cell count

MinhHaDuong:
**Ticket:** tickets/0340-standing-guard-every-generated-markdown.erg
Ticket-ref: tickets/0358-migrate-the-hand-rolled-makefile-parsers.erg
Ticket-ref: tickets/0368-cop-topic-structure-table-broken-by-comments.erg
Ticket-ref: tickets/0370-export-corpus-table-interpolates-raw-values.erg
Ticket-ref: tickets/0378-a-trailing-comment-adds-a-fifth-cell-to.erg

Three emitters have now shipped the same defect — a free-text value carrying a raw `|` interpolated into a Markdown pipe table, where the renderer does not error but drops the overflowing cell and shifts every later value one column left. Ticket 0325 fixed the deposited codebook; 0339 fixed both venue tables. Each was caught by a person reading output, never by a test. This is the standing guard for the fourth.

## Discovery reads the build, not a list

A hardcoded list of the three known artifacts is exactly what lets the next emitter escape. `tests/_mk_discovery.py` — the shared build-discovery surface that already exists so no guard can silently narrow its coverage (ticket 0248) — gains `makefile_constants` / `expand_vars` / `rule_targets` / `generated_markdown_targets`. Every Make rule target ending in `.md` is a generated Markdown artifact.

Eight resolve today, four present on this checkout:

| artifact | rule | on disk |
|:---|:---|:---|
| `codebook.md` | `Makefile:288` | yes |
| `tab_variables.md` | `Makefile:283` | yes |
| `tab_venues.md` | `Makefile:272` | yes |
| `tab_venues_fr.md` | `manuscript.mk:62` | yes |
| `tab_citation_coverage.md` | `Makefile:260` | no |
| `tab_core_venues_top10.md` | `Makefile:330` | no |
| `tab_corpus_sources.md` | `Makefile:275` | no |
| `tab_languages.md` | `Makefile:278` | no |

Absent artifacts are skipped, not failed — several need corpus data. `tab_venues_fr.md` is the case a rule-line grep would miss: its rule names it only as `$(GIDE_VENUES_FR)`, in a per-deliverable fragment.

## The parser

A hand scan, not `re.split(r"(?<!\\)\|", …)`. The lookbehind cannot count the backslash run, so it misreads `\\|` — an escaped backslash followed by a live delimiter — as an escaped pipe. CommonMark's rule is that a backslash escapes exactly the next character, which is what the loop does. Fenced blocks are skipped: `tab_variables.md` ships a raw-LaTeX `longtable` whose column specifications are not Markdown. The delimiter row is width-checked alongside the body rows, since GFM requires it to declare the same count.

## Evidence the guard is not vacuous

The ticket names this as the main risk, so three independent vectors were mutated and each went red:

1. **Real artifact, real defect.** Replaced `| Energy Economics |` with `| Energy | Economics |` in `deliverables/_shared/tables/tab_venues.md` (an unescaped pipe, exactly the shipped defect class). Red on the discovery path, not a fixture:
   > `FAILED tests/test_markdown_table_shape.py::test_generated_markdown_rows_match_their_header[deliverables/_shared/tables/tab_venues.md]` — `line 3: 6 cells, header at line 1 declares 5`

   Reverted with `git checkout --`; the artifact is untouched in this diff.
2. **Broken discovery.** Changed the suffix filter to `.markdown` so discovery returned zero targets. `test_discovery_reads_the_build_not_a_list` and `test_the_sweep_actually_parsed_rows` both failed — the sweep cannot silently shrink to zero parametrized cases.
3. **Blinded detector.** Suppressed the `tables.append(...)` so `pipe_tables` found nothing while every file was still read. Three tests failed, including `the sweep parsed only 0 table rows across 4 artifacts`.

Plus a red/green fixture pair on the codebook recipe verbatim, a real-world positive (`is_flagged` legitimately carries an escaped `\|` and must read as exactly 5 cells), and five parametrized payloads asserting that whatever `markdown_text_cell` emits passes — which ties the guard to the 0339 fix so the two cannot drift.

## Round-2 fix from the adherence self-gate

The first `/verify-adherence` pass returned one blocker, fixed in `3387ce51`: `rule_targets()` matched physical lines while `makefile_constants()` collapsed continuations, so a rule whose target list runs across a `\` lost every target declared before the line the colon lands on (reproduced at `Makefile:380`, two of three grouped targets vanished). The loss is silent — an undiscovered target is indistinguishable from an artifact that was never built, which the sweep skips. Both readers now go through one `_logical_lines()`. Also now accepts `export NAME := value` (five in this build); without it those names resolve to a literal `$(NAME)/…` path that is never on disk. Two regression tests pin both.

`/verify-adherence` re-run: **PASS**, no mechanical failures. Its two residual nits are pre-existing and converge on one change, filed as **ticket 0358** (two hand-rolled Makefile parsers survive alongside the shared one; migrating them is a pure refactor with its own mutation-verified regression surface, not something to fold into a new-guard PR).

## Two review rounds

**Round 1** found one defect with teeth, and it defeated this module's own anti-vacuity thesis: a raw newline inside a value ends the table at that line, so every row below belonged to no table and was never width-checked. `markdown_text_cell` folds a newline to a space; `markdown_cell` — the codebook path — does not, so the hole was reachable from a shipped emitter. `scan()` now reports three defects rather than one: a width mismatch, a row belonging to no table, and an unclosed fence. The last two are not extra features but the two ways this parser could call a file clean while never reading the rows that matter. Round 1 also caught the sweep's headcount floor counting artifacts *present* rather than artifacts that *parse* — `tab_variables.md` is a raw-LaTeX longtable with zero pipe tables by design and was propping up one of three units.

**Round 2** found the sharpest one. `expand_vars` resolves a bare `$(NAME)` and nothing else, so a `.md` target built with `$(foreach m,…,$(eval …))` reaches `GENERATED` carrying a literal `$(m)`, fails `is_file()`, and is skipped as "not built on this machine" — silently, forever. That idiom appears thirty-odd times in `divergence.mk` and `zoo-figures.mk` for `.csv` and `.png`: it is this repo's standard shape for a per-method artifact, and so the likely shape of the fourth emitter this guard exists to catch. Refusing the input beats teaching the resolver Make's function library — any target mentioning `.md` that did not flatten to a plain path now fails by name. Round 2 also pinned `_is_separator`, which gates the whole detector and survived `return True` with every test green, and split the fence toggle so a ``` inside a `~~~` block no longer reopens the document.

**Round 3** found the last two silent misses, both cases where `malformed_rows()` returned empty on a document carrying the very defect the guard hunts: a 4-tick fence closed by a 3-tick line (the parser resumes inside the code block and reports nothing at all), and a row that drops GFM's optional outer pipes falling between the two checks — table detection needs a `|` anywhere, orphan detection needed a *leading* one. The orphan predicate now reads whole lines. That broadening was measured before adoption: zero orphans across all four present artifacts, but not a contract hand-authored prose satisfies, which is the evidence ticket 0368 needs before extending the sweep to `_includes/`. Round 3 also made `+=` append rather than resolve first-wins, pinned through `makefile_constants(files=…)` — giving that parameter its first real caller.

One round-3 reviewer claim was disproven rather than fixed: `_logical_lines()` does not lose a rule declared after a `\`-continued recipe line. GNU Make 4.3 swallows that line into the recipe identically, so the parser is Make-faithful and the proposed fix would have introduced a divergence.

Every assertion added across the three rounds is mutation-verified. Dropping the delimiter row from the checked rows, the orphan report, the fence report, `_is_separator`'s test, or the fence-marker match each fails exactly one test; a length-blind fence match, a leading-pipe orphan predicate and a first-wins `+=` each fail exactly one; blinding the escape scanner fails eight. Two predicates are pinned on synthetic inputs rather than against the tree, because the constructs they guard do not exist in the build yet — the unresolved `.md` target and the `+=` operator. A forward-looking ratchet that finds nothing today would otherwise be this module's one untested line.

## Two live defects found, filed not fixed

The review surfaced two real defects outside this ticket's scope:

- **Ticket 0370** — `scripts/figures/export_corpus_table.py::_write_md_table` still interpolates `str(v)` raw. That is the fourth instance of the 0325/0339 class, and the first found by looking rather than by a reader noticing. This sweep cannot catch it today: `tab_corpus_sources.md` is untracked and absent here, so it is skipped.
- **Ticket 0368** — `deliverables/_shared/_includes/cop-topic-structure.md` has a stray fifth cell at line 38 (a trailing HTML comment past the closing pipe) and an orphaned row at line 42, found by the new orphan detector. A prose file, so the author arbitrates rather than an agent editing it.

That the orphan detector's first act was to find a real rendering bug in a file it does not even cover is the answer to the charge that it is scope creep.

## Round 4 — two defects the gate found by execution

Both were found by running the guard, not by reading it, and both are fixed
here (`532d11eb` red, `7f60a89a` green).

**The orphan check missed a raw newline in the last column.** Round 3's
broadening closed the hole only when the truncation remainder splits into two
or more cells. When the newline lands in the *last* column, the remainder
carries no delimiter at all — one cell, in no `claimed` set and failing the
`len(split_row(...)) > 1` test for `row_shaped` — so the row was accepted and
every row below it went unchecked. `malformed_rows("| A | B |\n|:--|:--|\n| ok |
first\nsecond\n")` returned `[]`.

Reachable from a shipped emitter, not hypothetical: `_deposit_variables.py:402`
routes each codebook Description through `markdown_cell`, which folds nothing
(`markdown_cell("a\n\nb")` → `'a\n\nb'`), and Description is the codebook's last
column. Torn on a real artifact, the pre-fix parser reads it clean:

```
$ # 3-line Description injected into codebook.md, last column
$ old parser (HEAD~1):  CLEAN — defect invisible
$ fixed parser:
    line 12: text abutting the table above with no delimiter in it — the last
             cell of the row above was torn by a raw newline, …
    line 13: a table row belonging to no table — the table above it ended early …
```

Two changes. `row_shaped` now keys on `has_delimiter()` — one unescaped pipe —
which is the predicate `scan()`'s docstring always described; the two-cell test
silently excluded exactly the remainder a torn last cell leaves. And because a
newline in the last column can leave a remainder with *no* pipe, `scan()` also
reports the line a table body stops on when that line is non-blank, non-fence
and delimiter-free. In generated output a table is followed by a blank line,
EOF or a fence, and nothing else. No block-start exemption: a torn Markdown
description can begin with `#`, so exempting headings would reopen the hole.

`scan()` returns a fourth value for torn lines rather than folding them into the
orphan list, because the two have different causes and deserve different
diagnostics — an orphan says the rows below went unchecked, a torn line says the
value above it was cut in half.

`test_parser_flags_rows_orphaned_by_a_raw_newline` was passing for a reason
unrelated to its name: its fixture appended a synthetic `| x | y | z |` row, and
*that* row was what failed. Strip it and the same broken input read as clean. It
now asserts the break line in both the stripped and the appended document.

**`unresolved_markdown` missed the multi-path-variable shape.** `rule_targets()`
split a rule's target list *before* expanding it, so a `$(VAR)` holding several
space-separated paths collapsed into one space-joined key. `Makefile:328` is the
live instance (three `.yml` behind `$(COMPUTED_STATS)`). Silent for `.yml`; with
`.md` it is worse than silent, it is actively reassuring — the joined key still
ends in `.md`, so the unresolved-target ratchet passes, `is_file()` then fails on
an impossible path, and every artifact behind that variable is skipped as "not
built on this machine". `unresolved_markdown("a.md b.md c.md")` returned `False`.

Make expands first and splits after; so does `rule_targets()` now. Discovery goes
241 → 261 targets with all eight `.md` targets unchanged, and no target carries
whitespace. `unresolved_markdown` keeps a whitespace check anyway — a path is one
whitespace-free token, and the guarantee belongs at the point of use so any
future resolver handing a list to a caller expecting a path fails loudly.

**Vacuity re-checked after the fix**, both mutations from the gate:

```
$ generated_markdown_targets: .md → .never
FAILED test_discovery_reads_the_build_not_a_list  — found only 0 generated Markdown targets
FAILED test_the_sweep_actually_parsed_rows        — assert 0 >= 3
2 failed, 30 deselected in 0.31s

$ stray unescaped | past the closing pipe, codebook.md:11
FAILED test_generated_markdown_rows_match_their_header[…/codebook.md]
    line 11: 7 cells, header at line 7 declares 5
1 failed, 3 passed, 4 skipped, 31 deselected in 0.19s
```

Both restored; the tree is clean in this diff.

## A third instance, filed not fixed

**Ticket 0378** — `deliverables/_shared/_includes/clustering-comparison.md:51`
carries a provenance HTML comment past its closing pipe, a stray fifth cell
against a four-column header. Same defect as 0368, different file, so it is a
habit rather than a slip. Filed with `needs-human` for the same reason 0368 is:
the note claims the table's 27,315 predates a ~27,509 corpus, so the right fix
may be a data refresh rather than a move — the author's call.

Its exit criteria carry the measurement 0368's third action asked for and did
not have: extending the sweep to `_includes/` costs 9 findings across 28 files,
2 real cell-count mismatches and 7 orphans, most of the latter being ordinary
prose pipes inside HTML comments that would need an exemption, not a fix.

## Local gates

There is no CI in this repo (ticket 0321), so both gates were run locally in this worktree:

```
make lint        215 passed, 10 skipped, 1 warning   in 17.42s
make check-fast  1217 passed, 11 skipped, 26 warnings in 11.83s
```

The guard is `adherence`-tier — it reads files and runs nothing — so `make check-fast` carries none of it, per the ticket's invariant. The module runs 35 tests plus 4 skips in 0.45 s.

Nothing in the *covered* tree was flagged: all four present generated artifacts are well-formed, which is expected after 0325 and 0339 landed. The two defects above sit outside the sweep — one behind an absent artifact, one in a hand-authored include the sweep does not claim.


MinhHaDuong (commented):

Correctness · Consistency · Scope · Red team · Doc propagation, each in a fresh context. Local gates re-run in the PR worktree:

```
make lint        187 passed, 10 skipped     ✅  (matches the PR body)
make check-fast  1168 passed, 11 skipped    ✅  (matches the PR body)
```

**Panel verdict: comment** — 4 of 5 reviewers `comment`, red team `request-changes`. Dissent preserved below; I did not merge the split. The guard is a real net improvement over zero coverage and its two headline mutation claims reproduce, so I do not think the residual gaps justify blocking. Two of them are cheap enough to fix on this branch.

### Mutation claims verified independently

The PR body's evidence holds. Re-run in this worktree, each reverted with `git checkout --`:

- `verifiable:` Unescaped `|` injected into `deliverables/_shared/tables/tab_venues.md` → `test_generated_markdown_rows_match_their_header[…tab_venues.md]` red, on the discovery path, not a fixture.
- `verifiable:` Suffix filter `.md` → `.markdown` at `tests/_mk_discovery.py:165` → `test_discovery_reads_the_build_not_a_list` and `test_the_sweep_actually_parsed_rows` both red.
- `verifiable:` A **fourth** mutation the PR did not claim — blinding `split_row` to a fixed cell list — is also caught (3 tests red). The anti-vacuity design works.

Also confirmed: the new `_mk_discovery` code is purely additive; all 5 pre-existing callers use only `mk_fragments`/`all_makefiles` and are unaffected.

---

## Findings

### 1. `verifiable:` A newline in an interpolated value is invisible — and blinds the sweep to every row below it

**Red team, high confidence — reproduced.** `pipe_tables` (`tests/test_markdown_table_shape.py:122`) ends a table at the first non-blank line lacking a `|`. A free-text value carrying a raw newline produces exactly that line, so the table terminates there and every subsequent row is never examined:

```
| Key | Value |
|:--|:--|
| a | multi
line
| b | fine |
| c | THIS ROW HAS | TOO MANY | CELLS |
```

```
malformed_rows() -> []
pipe_tables()    -> [(header line 1, 2 rows)]   # rows 5-6 never parsed
```

The genuinely malformed 4-cell row on the last line is silently dropped. `markdown_text_cell` closes this for its own callers (`'a\nb'` → `'a b'`), but `markdown_cell` does not (`'a\nb'` → `'a\nb'`) — and `markdown_cell` is the codebook path (`scripts/_deposit_variables.py`). This is the "passes by finding nothing" hazard the module docstring at lines 16-20 says it closes, reachable through a shipped emitter.

Not a blocker in my read — the ticket's stated class is the `|` character, and the guard honours that — but it is the one finding I would want closed before this is relied on. A narrow fix: inside a table region, treat a non-blank pipe-less line as an offender rather than a terminator, terminating only on a blank line or a fence. Every current emitter puts a blank line after its table, so the false-positive risk looks low. Ticket is fine too, if you prefer not to touch the parser here.

### 2. `verifiable:` `tab_variables.md` counts toward the anti-vacuity floors while parsing zero tables

Measured across all 8 discovered targets on this checkout:

| artifact | tables | rows |
|:---|---:|---:|
| `codebook.md` | 4 | 37 |
| `tab_venues.md` | 1 | 17 |
| `tab_venues_fr.md` | 1 | 17 |
| **`tab_variables.md`** | **0** | **0** |
| (4 others) | absent | — |

`tab_variables.md` is one of only four present artifacts and satisfies `len(present) >= 3` (`tests/test_markdown_table_shape.py:299`) while contributing nothing the parser can check — its content is a fenced raw-LaTeX `longtable`, correctly skipped. The `>= 30` row floor has real headroom (71 actual), but the `present >= 3` floor has one unit of margin and one of its three units is inert. Consider flooring on *artifacts that yielded at least one table* instead of artifacts on disk.

### 3. `verifiable:` "six" `export`/`override` assignments is five — the miscount is in four places

Exhaustive grep across `Makefile` + every `.mk`:

```
Makefile:60   export PYTHONHASHSEED := 0
Makefile:61   export SOURCE_DATE_EPOCH := 0
Makefile:72   export PYTHONPATH := …
Makefile:86   export PATH := …
deliverables/manuscript/manuscript.mk:25   export SOURCE_DATE_EPOCH ?= 0
```

Five. The sixth `export`-prefixed line, `Makefile:96 export UV_PROJECT_ENVIRONMENT`, is a bare re-export with no `=` — not an assignment, and `_ASSIGN` correctly does not match it. The claim appears at `tests/_mk_discovery.py:24`, `tests/test_markdown_table_shape.py:260`, in the PR body, and in ticket 0358. Found independently by two reviewers. Cheap to correct on this branch; the code is right, only the prose is off by one.

### 4. `verifiable:` Ticket 0358 cites `tests/test_arch_compliance.py:380` as a hand-rolled Makefile parser — it is not one

That line is `_ASSIGN = re.compile(r"^\s*([A-Za-z_]\w*)\s*=(?!=)")` inside `TestCorpusThroughLoaders` — a **Python-source** assignment regex for architecture rule 9 (corpus access through loaders). The file touches `MAKEFILE` exactly once, at line 120, for a `".feather" in content` substring check. No Makefile parsing exists anywhere in it.

Found independently by Consistency and Doc-propagation, confirmed on my own read. As written, ticket 0358's Action 2 ("migrate the other two copies") and its exit criterion are unsatisfiable — whoever picks it up will hunt for a parser that isn't there. The other two citations (`test_phase_layout.py:46`, `test_evict_analysis_intermediates.py:180`) do check out, as does the first-wins/last-wins divergence claim. Worth correcting the ticket while it is still on this branch.

### 5. `verifiable:` An unclosed code fence blinds the parser to the rest of the file

A file opening a fence that is never closed leaves `fenced = True` for the remainder, so a real table below it — including one with a 3-cell row under a 2-cell header — is never scanned: `malformed_rows()` returns `[]`, no error, no crash. No artifact has an unbalanced fence today, so this is latent — but combined with finding 2, an artifact can pass its own parametrized case by yielding zero rows while the tree-wide `>= 30` floor stays satisfied on the strength of its siblings.

### 6. `verifiable:` The delimiter-row guarantee has zero mutation coverage

`tests/test_markdown_table_shape.py:99-101` states delimiter-row inclusion is deliberate ("GFM requires it to declare the same column count"). Deleting the delimiter row from `rows` at line 120 leaves the full module green (17 passed, 4 skipped, 0 failed). No fixture or shipped artifact has a malformed delimiter row, so the stated guarantee could be removed without any test noticing.

### 7. `verifiable:` Two of the four new parser capabilities serve zero `.md` targets today

Measured by disabling each feature and re-running discovery:

| capability | needed by any of the 8 `.md` targets? |
|:---|:---|
| variable expansion (`expand_vars`) | yes — `tab_venues_fr.md` via `$(GIDE_VENUES_FR)` |
| grouped-target `&:` handling | yes — `tab_corpus_sources.md` |
| `export`/`override` assignments | **no — 0/8** |
| `\`-continuation collapsing | **no — 0/8** |

The PR is candid about this (`test_rule_targets_reads_a_continued_target_list` guards a `.csv`, and says outright "No `.md` target is written that way today"), and ticket 0358 explains the real motive: `_mk_discovery.py` is being grown into the single parser for the other hand-rolled readers. Legitimate shared-infrastructure instinct, but it ships ahead of its justification — 0358 itself records that `makefile_constants(files=...)` has no caller and a docstring naming one that does not exist. Not something to unwind now; noting it so the follow-up actually lands.

### 8. `consider:` `adherence` tier for a content-shape sweep

The module runs in ~0.15 s of pure-Python file reads with no subprocess — the project's own definition of the *fast* tier by cost. `test_phase_layout.py` is a precedent for `adherence` on a Makefile reader, but it enforces architecture hygiene; this checks generated *content*. The consequence: a regression in this defect class surfaces only at `make lint`/`make check`, never in the fast inner loop, at no cost saving. Judgment call, no rule violated.

### 9. `consider:` Discovery runs at import time

`GENERATED = generated_markdown_targets()` (`tests/test_markdown_table_shape.py:41`) reads the Makefile at collection. On a partial checkout with the Makefile absent, this raises `FileNotFoundError` as a pytest **collection ERROR** rather than a test failure — reproduced by renaming the file away. Diagnosable but misleading.

### 10. `consider:` Back-to-back tables with no blank line merge into one

Two adjacent tables with no separating blank line are parsed as a single table, with the second table's header and delimiter read as body rows of the first. Both false-positive and false-negative directions are reachable depending on the column counts. Latent: `codebook.md`'s four tables are blank-line separated, and the venue tables have one table each.

### 11. `nofollow:` Scoping notes, checked and dismissed

- Hand-authored prose tables under `deliverables/_shared/_includes/*.md` are out of scope — not Make targets, and not emitter-interpolated. Consistent with the project's documented "net for wired producers, not a complete phase audit" convention.
- `tab_venues_fr_caption.md` is correctly not discovered: it is a prerequisite, not a target.
- `split_row` not protecting a pipe inside a backtick code span **matches real GFM** (raw pipe scan precedes inline parsing); `markdown_cell`/`markdown_text_cell` escape inside code spans, so shipped artifacts never hit it. Verified, not a defect.
- No count-preserving single-pipe corruption found: under the `"| " … " |"` wrap convention every current emitter uses, a stray pipe at any leading/trailing/middle position is always caught.
- `_SEPARATOR_CELL` accepts the GFM minimum `-`, plus `---:` and `:---:`. Cyclic `A := $(B); B := $(A)` resolves to a literal rather than recursing. Tab-indented recipe continuations are correctly excluded from `_RULE`/`_ASSIGN`. `$(patsubst …)`-style function calls are correctly left unexpanded.
- The outer scan loop cannot infinite-loop; `i` strictly advances. The fence-toggle-inside-a-broken-row case does not desync.
- No duplicate pipe-table parser exists elsewhere in `tests/` or `scripts/` — `scripts/_markdown_table.py` only escapes, never parses. So this is not, ironically, a fourth parser.
- Doc-propagation verified every other factual claim in the new docstrings and **all eight** `Makefile:NNN` / `manuscript.mk:62` line references in the PR body: exact matches, including the on-disk yes/no column. `ruff check` clean on both files. `make lint` auto-discovers the module via its `pytestmark`; no marker or inventory file needs editing.
- One transient empty-`GENERATED` double-failure was observed once when running `test_mk_discovery_unified.py` and this module together; 8 retries including `-n4` all passed. Not reproduced, recorded rather than diagnosed.

---

## Recommendation

Merge-worthy. Before merge, the two cheap corrections — finding 3 (five, not six) and finding 4 (the wrong `test_arch_compliance.py` citation in ticket 0358, which will misdirect its executor) — are worth landing on this branch since they are pure text.

Finding 1 is the one with real teeth: the guard's own anti-vacuity thesis is defeated by a newline in a `markdown_cell` value, which also drops every row below it. Either the narrow parser change or a filed follow-up ticket, author's call.

<sub>Panel: 5 agents, `sonnet`, fresh contexts. All experiment mutations reverted; `git status --short` clean.</sub>

MinhHaDuong (commented):

Risk assessed **substantial**: 615 added lines, a new shared Make-parsing surface that other guards are slated to migrate onto, and a discovery-driven guard whose whole value is catching a defect nobody has thought of yet. All five perspectives ran.

**Panel verdict: comment.** Ticket 0340's exit criteria and its verification checklist are met on the tree as it stands. Gates re-run in the PR worktree: `make lint` → **190 passed, 10 skipped**; `make check-fast` → exit 0, unchanged; the module itself runs in 0.16 s. The findings below are about the guard's *future* reach, not about what it ships.

**Dissent preserved.** The red team returned **request-changes** — "the guard is net-positive and should merge, but two holes deserve a fast follow-up before the next emitter lands." The other four returned comment/approve. The disagreement is entirely over whether findings 1 and 2 block or follow up. The author decides.

---

### 1. `verifiable:` Templated and list-function `.md` targets escape discovery silently — the guard's central promise does not hold for this codebase's dominant idiom

Two reviewers found this independently; reproduced at synthesis. `expand_vars` substitutes bare `$(NAME)` only, so a target built with a Make list function does not resolve:

```
$ expand_vars("$(addprefix $(TABLES)/tab_,$(addsuffix .md,$(METHODS)))", consts)
'$(addprefix deliverables/_shared/tables/tab_,$(addsuffix .md,a b))'

$ [k for k in rule_targets() if "$" in k][:4]
['$(foreach', 'm,S1_MMD S2_energy …,$(eval', 'data/derived/tables/tab_div_$(m).csv', …]
```

Two sub-cases, and they fail differently:

- **`$(addprefix …)` form** — the key never ends in `.md`, so it is dropped from `GENERATED` outright. Invisible.
- **`$(foreach m,…,$(eval …))` form** — the key *does* end in `.md` (`…/tab_summary_$(m).md`), so it enters `GENERATED`, then `path.is_file()` is False and the parametrized test **skips it, silently, forever**.

This is not hypothetical: `scripts/analysis/divergence.mk` and `zoo-figures.mk` already use the `$(foreach…$(eval…))` idiom 30+ times for `.csv` and `.png` per-method outputs. It is *the* established way this repo adds a new templated artifact — so it is the likely shape of the fourth emitter the ticket exists to catch.

`rule_targets`'s docstring anticipates the imprecision and calls it "harmless to any suffix-filtering caller." That holds for the `$(addprefix)` form; it does **not** hold for the `$(m)` form, where the unresolved path is indistinguishable from "artifact not built" — precisely the conflation the same docstring flags as the silent failure mode two paragraphs earlier.

Cheap close: fail (or loudly `xfail`) on any discovered target whose resolved path still contains `$` or `%`, instead of letting it fall through to `pytest.skip`. One condition in `test_generated_markdown_rows_match_their_header`, and it converts the whole class from silent to loud.

### 2. `verifiable:` `_is_separator` has zero negative-space coverage — mutation survives with a byte-identical result

Reproduced at synthesis. Baseline and mutant, same worktree:

```
$ pytest tests/test_markdown_table_shape.py -q          # baseline
20 passed, 4 skipped in 0.16s

# _is_separator body replaced with `return True`
$ pytest tests/test_markdown_table_shape.py -q
20 passed, 4 skipped in 0.17s
```

`_is_separator` is the gate on the entire table detector — it decides what `scan` even treats as a table — and nothing constrains it. With `return True`, any two adjacent pipe-bearing lines become a table (probe: `"| A | B |\n| C | D |\n| 1 | 2 | 3 |"` now reports a width mismatch that is not one). The existing anti-vacuity tests all pass through it and none pin it. One test with a non-separator second line closes this.

### 3. `verifiable:` PR body carries two numbers that no longer match the tree

- **"six in this build"** (export assignments) contradicts the code comment at `tests/_mk_discovery.py:22-24`, which says five. **Five is correct** — `Makefile:60,61,72,86` plus `manuscript.mk:25`; `Makefile:96` is a bare re-export with no value and is correctly not an assignment. The code is right, the body prose is wrong.
- **"make lint 187 passed, 10 skipped"** → actual **190 passed, 10 skipped** (measured twice independently). The round-1 commit `40a70062` added tests after the body was written. `check-fast` (1168/11) and the 0.16 s module runtime both still hold.

### 4. `verifiable:` `makefile_constants(files=…)`'s docstring names a caller that does not exist

`tests/_mk_discovery.py:98-100` states the parameter exists because "`tests/test_phase_layout.py` deliberately reads the main Makefile alone." It does not: `test_phase_layout.py` never imports `_mk_discovery`'s constants reader, it has its own hand-rolled `_makefile_constants()` at line 46. Confirmed by grep — no caller anywhere passes `files=`.

Ticket 0358 owns the real fix and states this defect explicitly, which is the right call. But a two-line reword now ("reserved for the 0358 migration") removes a currently-false statement from committed code at near-zero cost, and the same claim is repeated in commit `a1b32033`'s message.

### 5. `verifiable:` Mismatched fence markers share one toggle — a miss vector and a false-positive vector

`scan` (line 124) treats any of ` ``` ` / `~~~` as toggling the same `fence_opened_at` state, so a `~~~` block closed by ` ``` ` is accepted, and the wrongly-extended region's tables go unchecked. The mirror case is worse for adoption: a document that *shows* fence syntax by example (an unmatched ` ``` ` inside a `~~~` block — valid CommonMark) reports a bogus "unclosed code fence," and a guard that false-positives on valid input gets disabled. No current artifact triggers either direction.

### 6. `verifiable:` Anti-vacuity floors are thinner than they read, and half the coverage is dormant on a normal checkout

- `len(with_tables) >= 3` has **zero margin** — exactly 3 present today.
- `sum >= 30` against **71** rows actually parsed tolerates a **58% silent collapse**.
- Without a `dvc pull`, **4 of 8** discovered targets are absent and skipped, so `make lint` in the ordinary dev state exercises half the claimed coverage. Worth one sentence in the module docstring so a future reader does not mistake a green run for full coverage.

### 7. `verifiable:` A live emitter still bypasses the escaper

`scripts/figures/export_corpus_table.py::_write_md_table` interpolates `str(v)` raw into its pipe table for `tab_corpus_sources.md` — the same anti-pattern 0325 and 0339 fixed, sitting next to escaped siblings. Not caught today only because that artifact is absent on this checkout and `Source` is a small fixed vocabulary. Discovery *does* find the target (via the `&:` grouped-target rule), so this is an unescaped emitter, not a coverage gap. Ticket-worthy on its own.

### 8. `consider:` A real, pre-existing rendering defect outside 0340's boundary

`deliverables/_shared/_includes/cop-topic-structure.md:42` — a row split off by an interspersed blank line plus comment block renders (verified against `pandoc -f gfm`) as raw paragraph text, not a table row. The orphan detector added by this PR is what surfaced it, which is a point in the orphan check's favour. Not a Make target, so out of the guard's declared scope; worth its own ticket. (Two `_includes/*.md` files also trip a *false* positive on a trailing HTML comment with no closing pipe, which pandoc silently tolerates — relevant only if the guard is ever pointed at hand-authored includes.)

### 9. `consider:` Table detection accepts a bare prose pipe followed by a dash-only line

`scan` line 133 requires only that the header candidate *contain* `|`, and does not cross-check the delimiter's width against the header's. So `"See Revista de Economía | Economics Review for details.\n---\n"` reports `line 2: 1 cells, header at line 1 declares 2` — a false positive on prose. Swept all 153 tracked `.md`: zero hits today. Inert, but it is the same class as finding 5 — the guard's false-positive surface is what determines whether it survives contact with a future artifact.

### 10. `consider:` Ticket 0358's line citation for the second parser is off

0358 cites `test_evict_analysis_intermediates.py:174-181`. The continuation joiner is at **159-177**; the last-wins logic it describes (`_variable_map`, "Later assignments win") is at **184-202**. The cited span catches neither mechanism cleanly. A handoff ticket's line numbers are its navigation.

---

### `nofollow:` — examined and closed

- **Orphan-row and unclosed-fence detection are not scope creep.** `markdown_cell` (the codebook path) does not fold newlines, so the *same* underlying bug — an emitter interpolating unescaped free text — manifests as a raw newline that silently ends the table and leaves every row below unchecked. Same defect class, different character. Fence-skipping is independently required (`tab_variables.md` embeds a raw-LaTeX `longtable` in a fence); asserting the skip terminates is a one-line addition on top.
- **`adherence` is the right tier.** Surveyed all 21 `@pytest.mark.adherence` modules: `test_phase_layout`, `test_layering`, `test_arch_compliance`, `test_mk_discovery_unified` and the rest are all the same shape — read the tree, assert a structural contract, run nothing, sub-second. The new module matches local practice exactly, even though that practice has drifted from the generic tier table's "ruff/mypy/hygiene" wording.
- **No dependency substitute exists.** `markdown-it-py` is present only transitively via `rich`, and its GFM pipe-table support needs `mdit-py-plugins`, which is not in the lock. Adding a dependency for a 40-line cell-count check would be worse. The hand-rolled parser is the right call, and `split_row` genuinely implements CommonMark's "backslash escapes exactly the next character" correctly — verified on the doubled-backslash-then-live-pipe case the docstring calls out as the lookbehind's failure mode.
- **Filing 0358 in this PR matches precedent** (`1c7a9388`, `d06bbe94`), it sits in its own isolated commit, and the body correctly names only 0340 in a closing `**Ticket:**` line. `erg check tickets/` passes (343 tickets). Ticket seat 0358 is free on `origin/main` and unclaimed by all 9 other open PRs.
- **Commit trail is a genuine red → green → fix → ticket → fix sequence** with per-commit rationale, not a dump. Ruff clean. No stale doc references, no duplicate guard to retire (`test_rendered_table_fidelity` and `test_variables_table` operate at different layers). Every one of the PR body's 8 rule locations and its "four present on this checkout" claim verified exact, as did the `Makefile:380` continuation claim and the `.gitignore` un-ignore that makes `codebook.md` always tracked. `mk_fragments()` returns all 14 on-disk `.mk` files, 1:1 with `find`.
- Adversarial sweep (10 MB lines, deep backslash runs, CRLF, non-UTF-8, fence-only, zero-body-row) found no crash and no pathological runtime. A BOM-prefixed first row gains a phantom leading cell, but nothing here writes `utf-8-sig`.

---

**Recommendation.** Merge is defensible as-is — the guard is a real improvement over nothing, and every exit criterion holds. The two findings worth closing *before* this guard is relied on as the sole backstop are **#1** (unresolved `$(…)` targets must fail, not skip — one condition) and **#2** (one test for `_is_separator`). Both are small enough to land here; **#3** and **#4** are text fixes. **#7** and **#8** are separate tickets.

*Panel: correctness, consistency, scope, red team, doc propagation — all `sonnet`, fresh contexts. No fixes applied; `/review-pr` posts comments only.*

MinhHaDuong (commented):

Correctness · Consistency · Scope · Red team · Doc propagation. All findings below were re-derived by the synthesizer before posting; two agent claims changed status in triage.

**Panel verdicts:** consistency `approve` · scope `approve` · doc-propagation `approve` · correctness `comment` · red team `request-changes`.

**Synthesis: `comment`.** Two reproducible blind spots in the new parser, neither reachable by any emitter shipping today. The guard is a clear net gain over nothing; the merge call is the author's. Dissent preserved below.

---

### Gate

```
make check   1813 passed, 7 failed, 82 skipped   (241 s)
make lint     199 passed, 10 skipped
make check-fast  1168 passed, 11 skipped
```

All 7 failures are `tests/test_corpus_acceptance.py::TestFileExistence` — DVC-gated corpus absent in the review worktree (`data/catalogs/refined_works.csv` is present at 63 MB in the primary checkout). Environmental, unrelated to this diff, which touches only `tests/` and three `.erg` files.

The module itself: **29 passed, 4 skipped, 0.12 s**; the 4 skips are the corpus-gated artifacts. Confirmed collected as `adherence` only — `-m "not slow and not integration and not adherence"` deselects all 33, so the ticket's `make check-fast` invariant holds.

---

### Two verifiable blind spots — recommended before merge

Both are silent misses: `malformed_rows()` returns **empty** on a document carrying the exact defect class this guard exists to catch. Flagged because this module's own stated thesis is that a discovery-driven guard must not pass by finding nothing.

**1. `verifiable:` — a 4+ backtick fence closes early, and the table below is never read.** HIGH confidence.

`scan()` stores `fence_marker = stripped[:3]` regardless of the opener's real run length, and closes on `startswith(fence_marker)`. CommonMark requires the closer to be the same character *and at least as long*. A 3-backtick line inside a 4-backtick fence closes it; the true closer then re-opens.

```python
doc = "````\n```\n````\n| A | B |\n|:--|:--|\n| 1 | 2 | 3 |\n```\n"
scan(doc)     # tables: []  orphans: []  unclosed: None
malformed_rows(doc)  # []   <- 3-cell row against a 2-cell header, reported clean
```

The red team found the loud variant (unclosed-fence reported instead of the real defect). The balanced variant above is worse: no diagnostic at all. Fix is small — capture the full marker run, require `len(closer_run) >= len(opener_run)`.

No shipped artifact uses a 4+ backtick fence today (`tab_variables.md` uses a plain 3-tick fence), so nothing is currently missed.

**2. `verifiable:` — an orphaned row without outer pipes is invisible to both checks.** HIGH confidence.

`row_shaped` appends only when `stripped.startswith("|")` (line 156), but header recognition and the body-continuation loop both accept `"|" in line` — valid GFM. A row that loses its table and lacks a leading pipe therefore belongs to no table *and* never enters the orphan set:

```python
malformed_rows("A | B\n:--|:--\nok | first\nsecond\nx | y | z\n")   # []      <- missed
malformed_rows("| A | B |\n|:--|:--|\n| ok | first\nsecond\n| x | y | z |\n")
# ['line 5: a table row belonging to no table …']                            <- control fires
```

Both current emitters always write outer pipes (`export_tab_venues.py:125-127`, `_deposit_variables.py:397`), so this is unreachable today. Note the fix is a design call, not a one-liner: broadening `row_shaped` to any `|`-bearing line would manufacture false orphans from prose. A window bounded to lines shortly after a table ended is the narrower option.

---

### Investigated and dismissed — preserved as dissent

**`nofollow:` — the correctness reviewer's top finding does not hold.** It reported that `_logical_lines()` folds a rule line into a preceding `\`-continued *recipe* line, silently losing the target, and proposed skipping the join for TAB-indented lines. Checked against GNU Make 4.3 itself:

```
foo.md:
	echo a \
bar.md:
	echo b
```
```
$ make -n bar.md
make: *** No rule to make target 'bar.md'.  Stop.
```

Make swallows the line into `foo.md`'s recipe exactly as `_logical_lines()` does. The parser is Make-faithful here; the proposed fix would introduce the divergence, not remove it. No action.

---

### Minor findings

- `consider:` (correctness, medium) — `split_row` splits on a **raw** `|` inside a code span; pandoc 3.1.3 treats the span as opaque and does not (`split_row("| \`x | y\` | z |")` → 3 cells; pandoc renders 2). A false-positive direction, unreachable because both escapers escape `|` inside spans too.
- `verifiable:` (scope) — `makefile_constants(files=...)` has zero call sites anywhere in the tree; every caller uses the `None` default. Self-disclosed in the docstring, with its first caller ticketed as 0358. Inert surface, not a behaviour risk.
- `consider:` (correctness, low) — `+=` resolves through `setdefault`, so `NAME := x` followed by `NAME += y` silently drops `y` rather than appending. No `+=` exists in the tree today. Same class: `::=` is misparsed by `_RULE` as a rule declaring target `NAME`; unused here.
- `consider:` (consistency) — `tests/_mk_discovery.py:21-26`: two em-dash interruptions in one comment paragraph, against the one-per-paragraph prose guard.
- `nofollow:` (red team) — probed and confirmed absent from this tree: `.PHONY:` targets ending `.md`, `define`/`endef` bodies parsed as rules, Windows drive-letter colons, absolute-path targets, and non-UTF-8 artifacts (which fail loudly with a raw `UnicodeDecodeError` rather than a clean message).
- `nofollow:` (scope) — `40a70062` and `44fd505d` each bundle a code fix with ticket filings, where `0b4b2c8a` correctly filed 0358 alone. The tickets are discovery artifacts of those review rounds, so the coupling reads as deliberate.

---

### What the panel verified rather than assumed

Unusually for a comment-dense diff, **every factual claim in the new code and the PR body was independently re-derived and holds**: the five `export`-prefixed assignments at `Makefile:60,61,72,86` + `manuscript.mk:25`; `Makefile:96` correctly excluded as a bare re-export; `Makefile:380-382` dropping two of three grouped targets under a physical-line reader; 37 `$(foreach)`/`$(addprefix)` occurrences across `divergence.mk` and `zoo-figures.mk`; eight discovered targets with four on disk, matching the body's table row for row including rule locations; 71 parsed rows against the floor of 30; and both gate figures.

Ticket 0361's claim verified at `export_corpus_table.py:103` (`str(v)` interpolated raw, `_write_md_table` spanning 89-107). Ticket 0362's verified at both cited lines — and `malformed_rows()` run against the real `cop-topic-structure.md` reports both defects, so the orphan detector does catch its motivating case. Every cited ticket ID (0219, 0222, 0227, 0239, 0244, 0248, 0325, 0339) resolves, subjects match, no ID collision with `origin/main` or any open PR, and `erg check tickets/` passes across 345 tickets.

Scope: one `**Ticket:**` line (0340), three findings filed-not-fixed-not-closed, ticket 0340 itself left untouched for `erg-pr-merge`. Checked against `feedback_scope_discipline` — no violation; that precedent concerned landing an architecture rule plus unrelated fixes, neither of which happens here.

One consistency finding worth the author's eye independent of this PR: `tests/test_evict_analysis_intermediates.py:159-246` hand-rolls a **second** ~88-line Makefile parser with **last-wins** resolution, opposite to this module's first-wins. Ticket 0358 covers it; the semantic divergence is worth naming in that ticket if it is not already.

PR #1179, opened 2026-07-27
fix(0360): share the primary DVC cache with every worktree

MinhHaDuong:
**Ticket:** none — 0360 was closed and archived earlier in this branch (commit `1a32ada7`), so `erg-pr-merge` has nothing left to close. Its file is at `tickets/closed/0360-worktree-dvc-cache-is-private-so-make-da.erg`; it supersedes the workaround that cost `tickets/closed/0347-*` its DVC bookkeeping.

> Corrected after review: the body previously claimed `**Ticket:** tickets/closed/0360-…`. `erg-pr-merge:147` extracts with `tickets/\K\d+`, so `closed/` breaks the match, and the `Ticket-ref:` NOCLOSE guard has the same shape — both came back empty and the merge would have died at Step 1 with "no close-claim in PR body".

## The bug

`make data` could not work in a worktree. `.dvc/` is not tracked wholesale: a worktree gets `config` from git and `config.local` from `.worktreeinclude`, but the cache is gitignored — so a fresh worktree started with an **empty private cache**, and `config.local` carries only the remote URL, never `cache.dir`. `dvc checkout` searched that empty cache and failed.

The consequence was not cosmetic: with the documented on-demand path unusable, the standing workaround became "do corpus work in the primary checkout" — which is exactly how **0347** lost worktree isolation and skipped its `dvc commit` / `dvc push`, leaving `dvc.lock` pointing at a superseded corpus.

## The fix

The post-checkout hook symlinks `.dvc/cache` at the primary's, the same idiom it already uses for `.venv`, with the path derived from `git rev-parse --git-common-dir` so a clone elsewhere works. Only an absent or already-symlinked `.dvc/cache` is replaced — a standalone checkout owns a real cache directory and clobbering it would orphan its blobs.

**Verified end to end**, not just by assertion: in a probe worktree the link resolves to the primary cache and `make data` checks the corpus out, delivering `refined_works.csv` at 66,400,079 bytes — the repopulated corpus from 0347.

## Two corrections to my own ticket, recorded rather than quietly swapped

**Action 1 was unusable.** The ticket proposed writing `cache.dir` into `.dvc/config.local`. That file is gitignored (`.dvc/.gitignore:1`), so the fix could not ship in a PR at all — it would be a per-machine setup step. Action 2's computed path is the only viable form.

**The disk-cost claim was wrong twice, and the method is why.** Draft 1 asserted "costs no extra disk" without measuring. Draft 2 measured badly: comparing inodes and running `du` **cannot distinguish a reflink from a copy** — a reflink produces different inodes, and `du` deduplicates by inode, not by shared extent. Correct: DVC's default `cache.type` chain is reflink-then-copy, and this repo is on btrfs, so `dvc checkout` reflinks. Verified with `filefrag` after a full `make data` in a probe worktree — workspace file and cache blob report identical physical extents flagged `shared`, and free space was unchanged across the 2.2 GB checkout. The 2.2 GB is the *apparent* size, and the real cost only where the filesystem has no reflink.

That reverses the reasoning, not just the number: copy-on-write is not a price paid for isolation, it is what provides the isolation, for free. `cache.type` still stays unset, and `hardlink` is still refused — one shared inode, and Phase-1 rewrites `data/catalogs/*.csv` in place.

## Tests

Three added to `tests/test_post_checkout_hook.py`, mirroring the file's existing split between source-inspection pins and one behavioural guard. All confirmed **red before the fix**.

The behavioural test drives the hook directly rather than relying on `git worktree add` to fire it, for two reasons: whether git runs hooks depends on `core.hooksPath`, machine-local config a test has no business asserting; and a fresh worktree is checked out at HEAD, so its own copy of the hook would grade the last commit instead of the change under test. My first draft made both mistakes.

## Heads-up, outside this PR's scope

On this machine `core.hooksPath` is set in `.git/config` to `/home/haduong/Climate_finance/hooks` — **a directory that does not exist**. So no git hook currently fires here: not pre-commit, not pre-push, not post-checkout. `make setup` sets it correctly to `.githooks`. I have not changed your local git config; the cause is not established (`on-start.sh` only sets it when `hooks/pre-commit` exists, so it is not the current culprit).

## Review rounds

Two `/review-pr` rounds, both request-changes, both acted on in full.

**Round 1** caught a dropped ticket action: the note in `.claude/rules/architecture.md` § Data location telling the next session that corpus work belongs in a worktree. The ticket had closed `fixed` with every box ticked and no box covered that action, which is how it stayed invisible. Done now, along with the one hazard the sharing widens (`dvc gc` prunes by reachability from whichever checkout invokes it). Round 1 also found a test docstring still asserting the retracted cost claim — and a behavioural test I added in response to it was itself vacuous: `ln -sfn` onto a real directory nests the link inside it rather than replacing it, so `is_symlink()` and a sentinel file both survive. It now asserts the directory's contents.

**Round 2** found the cost claim wrong (above) and a worse problem: that same never-clobber test could write into the real 19 GB cache. It called plain `git worktree add`, and where `core.hooksPath` is configured git fires post-checkout itself, so the link exists before the fixture runs. Reproduced by injecting `core.hooksPath` via `GIT_CONFIG_*` — the test failed and wrote a stray blob into the primary cache, which I removed. Fixed with `-c core.hooksPath=/dev/null` plus a precondition assert.

Round 2 also let me verify the fix through the documented path rather than a hand-driven probe: with hooks pointed at this branch's `.githooks`, `git worktree add` created the link and `make data` checked the corpus out of the shared cache.

## Two follow-ups filed, not folded in

- **0380** — `make data` still exits 255 on `data/catalogs/run_reports/catalog_merge.json`, which has no hash info in `dvc.lock`. The whole corpus lands; the target does not. Separate from this bug, and previously masked by it. (Allocated as 0378 and renumbered: `origin/main` and PR #1172 both took that seat while this branch was in review.)
- **0379** — a guard for `dvc gc`. This PR documents the hazard, which is not the same as preventing it.

## Gate

`make lint` 278 passed · `make check-fast` 1303 passed · hook tests 16 passed (3 integration) · `erg check` PASS (369 tickets). No CI here, so this is a local gate on the merged result.

`origin/main` integrated twice during review (18 then 50 commits). One conflict, in ticket 0366 — closed upstream while this branch logged a note on it — resolved as a chronological union; `architecture.md` auto-merged and the union was verified by grep rather than trusted to the clean exit.

One red gate after the second integration was **not** this branch: `test_keys_line_selects_every_consumed_credential` failed on the worktree's stale `.env`, which `.worktreeinclude` copies once at creation and which had since gained the `hal:` entries from ticket 0364. It passes on the primary checkout at the same commit. Refreshing the local copy cleared it. Worth knowing: a long-lived worktree drifts from the machine's credential selection, and the symptom reads like a code failure.

## Round 3

No blockers. Two findings fixed anyway, both in the argument rather than the code.

The documented write mechanism was wrong, and it was the half carrying the `hardlink` refusal. Phase 1 does *not* rewrite `data/catalogs/*.csv` in place — `pipeline_io.save_csv` writes a temp file and `os.replace()`s it, an atomic rename onto a fresh inode that would survive even a hardlinked cache. The hazard is real but lives elsewhere: `np.savez_compressed()` in `enrich_embeddings` and `corpus_align` opens the embeddings `.npz` directly and truncates it. One in-place writer is enough, so the conclusion stands and the reasoning is now accurate.

"`cache.type` stays unset" was asserted in three prose locations and enforced in none, while `dvc config --local cache.type hardlink` lands in the gitignored `.dvc/config.local` where no diff would show it. `test_dvc_cache_type_is_not_an_aliasing_type` now parses both config files and rejects `hardlink`/`symlink`; confirmed red by injection.


MinhHaDuong (commented):
## `/review-pr` — five-perspective panel on #1179

Panel: correctness · consistency · scope · red team · doc propagation (all `sonnet`, read-only, fresh contexts).

**Local gate re-run in the PR worktree** (no CI in this repo): `make lint` 180 passed / 6 skipped · `make check-fast` exit 0 (1228 + 25 collected, all passed) · `pytest tests/test_post_checkout_hook.py -m integration` 2 passed. The slow tier was not run.

**Panel verdicts: 4 × request-changes, 1 × comment.** The dissent is preserved below — it matters.

The mechanism itself survived review well. The red team could not break the symlink logic: the `[ ! -e .dvc/cache ] || [ -L .dvc/cache ]` guard genuinely never clobbers a real cache directory, `rm -rf`/`find -delete`/`git clean -xfd` all refuse to descend a symlinked directory (verified in a `/tmp` sandbox, trailing slash included), a dangling link produces a clean DVC error rather than a crash, and the copy-not-hardlink rejection is correct on its premise (~25 Phase-1 scripts do rewrite `CATALOGS_DIR` CSVs in place). Both blockers below are about text and process, not about the four lines of shell.

---

## Blocker 1 — `erg-pr-merge` will abort on this PR body, before any merge happens

*(scope; mechanism corrected by my own reproduction — see the dissent note)*

The close claim is `**Ticket:** tickets/closed/0360-...`. `erg-pr-merge`'s extractor requires digits **immediately** after `tickets/`:

```
^\*{0,2}ticket:?\*{0,2}:?\s*tickets/\K\d+
```

`tickets/closed/0360` does not match. Neither does the `**Ticket-ref:** tickets/closed/0347-...` line match the no-close suppressor (`...tickets/\d+`). Reproduced against the live PR body:

```
$ BODY=$(gh pr view 1179 --json body -q .body)
TICKET_IDS: []
NOCLOSE=(empty)
NONE=(empty)
```

With all three empty and `tickets/` present, `erg-pr-merge:155` reaches
`die "no close-claim in PR body"` — it bounces at step 1, before the merge.

Correcting the line to `tickets/0360-...` does **not** fix it either: the existence-at-tip guard (`erg-pr-merge:211`) resolves the slug from `tickets/`, and the ticket has already been moved to `tickets/closed/` on this branch:

```
$ compgen -G "tickets/0360-*.erg"
(no match in tickets/)
```

→ `die "close-claimed ticket(s) 0360 are absent from tickets/ at the branch tip"`.

Root cause: the ticket was created, closed (`Closed: fixed`), and archived to `tickets/closed/` all inside the single commit `1a32ada7` — it never exists in an open state anywhere in git history. `erg-pr-merge` is built to perform the close itself at merge time.

**Fix, pick one:** (a) leave the ticket at `tickets/0360-...` on the branch without the `Closed:` header or the archive move, restore the `**Ticket:** tickets/0360-...` line, and let `erg-pr-merge` close + archive it — this is the convention; or (b) keep the pre-close and state in the PR body that `erg-pr-merge` must be skipped and the merge done directly with `gh pr merge 1179 --merge`. As filed, the documented merge path fails.

## Blocker 2 — a test docstring added by this PR asserts the exact claim this PR retracts

*(consistency, doc propagation, and independently red-flagged — 3 of 5 reviewers converged)*

`tests/test_post_checkout_hook.py:114-115`:

> Symlinking is the same idiom the hook already uses for .venv, **and it costs nothing: dvc hardlinks out of a cache on the same filesystem.**

That is the "costs no extra disk" claim the PR's headline correction demolishes. Eleven lines away, `.githooks/post-checkout:50-58` — added in the same commit — says the opposite, measured:

> DVC's cache.type is unset here, so its default is `copy` and each worktree that asks for data spends ~2.2 GB of real disk.

The correction landed in the hook comment and the ticket log; it did not land in the new test's docstring. The PR contradicts itself in one diff, and a reader who trusts the test walks away with the retracted belief. One-line fix in a file this PR already touches.

---

## Minor findings

`verifiable:` **Module docstring is now false.** `tests/test_post_checkout_hook.py:4-5` — "The post-checkout hook **only** co-locates the uv environment". The hook's own top comment was correctly updated to "two shared, off-tree resources" (`.githooks/post-checkout:7-12`); the test module docstring sitting directly above the two tests that exercise the second resource was not.

`verifiable:` **`test_hook_replaces_stale_dvc_cache_symlink` is anchored on prose, not on code.** It slices around the *first* occurrence of `".dvc/cache"`, which is inside a comment sentence at `.githooks/post-checkout:61`, not at the conditional. Measured on the current file: the first occurrence is at offset 3526/3949, leaving `len(after) = 413`; `ln -sfn` sits at offset 355 within that tail, so there are **238 characters of headroom** before it falls outside the `after[:600]` window. Adding ~240 chars of pure rationale prose to that comment block — no logic change — makes both assertions fail. Given the block is already ~60% prose and invites more, split on `SHARED_CACHE=` instead.

`verifiable:` **Ticket action 3 never shipped, and closed silently anyway.** The ticket's action 3 asks for a note in `.claude/rules/architecture.md` § Data location that a worktree needs `make data` once and that corpus work belongs in a worktree. Actions 1 and 2 got explicit dispositions (`~~struck~~ Unusable`, `**Done.**`); action 3 got none. `git diff origin/main...HEAD -- .claude/rules/architecture.md` is empty, and § Data location (lines 128-152) contains zero occurrences of "worktree" or "make data". Yet every Verification and Exit-criteria box is `[x]` and the ticket is `Closed: fixed`. None of the checked criteria covers action 3, which is how the omission passed closure. Either add the note or give action 3 the same explicit disposition the other two got.

`consider:` **`dvc gc` from a worktree is now a repo-wide destructive operation.** Before this PR each worktree's cache was private and empty, so `dvc gc` there was a no-op. Now every worktree points at the one shared cache, and `dvc gc` computes its reachable set from *that worktree's* `dvc.lock`. The red team reproduced blob loss in a `/tmp` sandbox: a worktree pinned to an old commit ran `dvc gc --workspace` and removed the primary's newer, not-yet-pushed blob ("Removed 1 objects from repo cache"); the primary's next `dvc checkout` then failed. Compounding it, `.dvc/tmp` (holding DVC's `lock`/`rwlock`) stays *per-worktree* — only `cache` is symlinked — so two worktrees each hold "their own" exclusive lock while mutating one cache. `dvc gc` appears nowhere in the Makefile, scripts, or docs (grepped), so this is latent rather than active, and content-addressed writes make ordinary concurrent checkouts safe. Worth one line in the hook comment: run `dvc gc` only from the primary, only after `dvc push`.

`consider:` **The behavioural test has an undeclared environmental precondition.** The hook only links when `[ -d "$SHARED_CACHE" ]`, i.e. the primary must already hold a populated `.dvc/cache`. On a fresh clone, or any machine where dvc was never run in the primary, `test_fresh_worktree_shares_the_primary_dvc_cache` fails with a message that actively misleads ("the worktree has a private cache" — it has none at all). The repo's own convention for this class of precondition is `pytest.mark.skipif`, used elsewhere in `tests/` for tool availability.

`consider:` **The ~2.2 GB cost is documented where nobody reads it before paying it.** `Makefile:198-201`, the comment directly above the `data:` target, still cites only the historical "~1.7 GB" eager-copy and says nothing about the newly measured per-worktree cost. That comment is the one place a reader looks immediately before typing `make data`; the cost currently lives only in a git hook and a closed ticket. Neither README, AGENTS.md, nor architecture.md mentions it.

`consider:` **The primary checkout becomes an undocumented single point of failure.** The symlink target is an absolute path baked in at hook-run time. Clearing the primary's local cache to reclaim disk (trusting the padme remote) used to affect one checkout; it now breaks `make data` in every live worktree simultaneously, and moving or renaming the primary orphans every existing worktree's link with no self-repair. It fails cleanly rather than destructively, so this is a documentation gap.

`nofollow:` Comment-to-code ratio in the new hook block is 23:5, against roughly 9:15 in the sibling `.venv` block it mirrors. The copy-vs-hardlink rejection rationale is genuinely load-bearing — it stops a future "optimisation" from corrupting the shared cache — so this is a stylistic preference, recorded not pressed.

`nofollow:` The self-link guard `[ "$SHARED_CACHE" != "$(pwd)/.dvc/cache" ]` compares raw strings without canonicalisation, so symlink aliasing could in principle defeat it. No reviewer could construct a failure: the independent `-e`/`-L` guard is a complete backstop, since a real cache directory fails it regardless.

`nofollow:` Invoked from a subdirectory the hook would create a stray, useless `<subdir>/.dvc/cache` link. Git always runs hooks with cwd at the worktree toplevel, and the new integration test drives it the same way, so this is unreachable in practice.

`nofollow:` The ticket's own § Test section still specifies the retracted assertion ("compare inode … hardlink or reflink rather than a copy"), which its later Verification section explicitly corrects. Residual inconsistency inside the ticket body only; the shipped test correctly asserts symlink structure instead.

---

## Dissent preserved

**Correctness returned `comment`, not `request-changes`, and its reasoning is the most load-bearing thing in this review.** It traced and *empirically verified* every branch of the shell: the four states of the `-e`/`-L` guard, both `--git-common-dir` invocation contexts on git 2.43 (`.git` relative from the primary toplevel, absolute from a worktree), the self-link guard from the primary checkout (block correctly skipped, real directory untouched, `islink` → False after an in-place hook run), and the `cd`-failure path outside any git repo (`SHARED_CACHE` becomes `//.dvc/cache`, `[ -d ]` is false, safe no-op at exit 0 — with a cosmetic stderr leak, since the outer `2>/dev/null` scopes only the `cd`, not the inner `git rev-parse`). Its verdict: nothing here blocks shipping the *behaviour*; the problems are in the tests and prose the PR adds around it. I agree, and neither blocker above contradicts it.

**One reviewer's mechanism was wrong and I am flagging it rather than laundering it.** Scope reported that `erg-pr-merge` recovers `0360` via a title fallback (`fix\(\K\d+`) and then dies inside `erg close`. That is not what the script does: `TITLE_TID` (`erg-pr-merge:227`) feeds only a *non-blocking warning*, never `TICKET_IDS`. My reproduction above shows the abort happens earlier and more cleanly, at the step-1 `die "no close-claim in PR body"`. Same conclusion, different failure point — and the difference changes the fix, so it is worth stating.

---

No fixes applied and no gate cast — this is `/review-pr`. Recommend a follow-up commit on this branch for the two blockers plus the three `verifiable:` items (all small, all caused directly by this diff), and a fast-follow ticket for the `dvc gc` guard.

MinhHaDuong (commented):
## `/review-pr` panel — round 3 (five perspectives)

**Verdict: comment — no blockers.** Panel split 3 approve (correctness, consistency, scope) / 2 comment (red team, doc propagation). No agent requested changes. The dissent is narrow and preserved at the bottom.

Local gate re-run on this branch: `make lint` **180 passed**, fast tier **1244 passed / 11 skipped**, `tests/test_post_checkout_hook.py` **15 passed**.

Two things the panel verified rather than took on trust, worth recording because they were the loudest fears going in:

- **The behavioural tests are not vacuous.** Correctness reverted the whole DVC block and confirmed three tests go red; then mutation-checked the never-clobber guard in a scratch copy of the hook (replacing the `{ [ ! -e … ] || [ -L … ]; }` condition with `true`) and confirmed it is specifically the `sorted(...) == ["sentinel-blob"]` assertion that bites — `is_symlink()` alone still passes, exactly as the docstring claims.
- **Nothing deletes through the cache symlink.** Red team reproduced twice in `/tmp`: `git worktree remove --force` and `shutil.rmtree` both unlink the symlink node rather than descending. The `finally:` blocks in the new integration tests cannot reach the real 19 GB cache, even with a live link present mid-test. The one historical write incident (the stray sentinel blob) is correctly closed by `-c core.hooksPath=/dev/null` plus the precondition assert.

---

### Findings

**`verifiable:` The `cache.type`-stays-unset invariant is stated in three prose locations and enforced in none.** *(red team, high)*
The entire safety argument — reflink COW is what keeps a Phase-1 write off the cache blob — rests on `cache.type` being unset. `grep -rn "cache.type\|cache_type" tests/ .claude/ Makefile` returns only prose: `tests/test_post_checkout_hook.py:117-120` and `.claude/rules/architecture.md:156,165`. No assertion anywhere. Worse, `cache.type = hardlink` can land in `.dvc/config.local`, which is gitignored — so it would never appear in a diff. Before this PR a bad `cache.type` damaged one private empty cache; after it, one shared 19 GB cache behind every checkout. Suggest an adherence test asserting `cache.type` is absent or `reflink` in both `.dvc/config` and `.dvc/config.local`.

**`verifiable:` The documented write mechanism is wrong, and it is the one carrying the `hardlink` argument.** *(red team, high — confirmed independently)*
Hook comment and `architecture.md` both say "Phase-1's in-place rewrites of `data/catalogs/*.csv`". The actual path for every DVC-tracked catalog is `scripts/pipeline_io.py:save_csv()` — `tempfile.mkstemp` in the same dir, then `os.replace()` at `scripts/pipeline_io.py:128`. That is an atomic rename onto a **new inode**, not an in-place rewrite; `grep` finds no direct `to_csv()` on any catalog path. The reality is safer than documented (a rename never touches the old extents, and would survive even a `hardlink` misconfiguration for these files), but the stated mechanism is the load-bearing half of "why `hardlink` would corrupt every checkout at once" — and it is not what the code does. Fix the wording, or the next reader reasons from a false model.

**`verifiable:` The hook does not fire on this machine, and only a closed ticket's log says so.** *(doc propagation, high — confirmed)*
`git config --get core.hooksPath` → `/home/haduong/Climate_finance/hooks`; `ls` on it fails. No git hook runs here: not pre-commit, not pre-push, not this one. The PR body flags it as out of scope, and `make setup` repairs it — but that fact now lives only in `tickets/closed/0360`'s log, not in any living doc nor in the `setup:` target. A line in `.claude/rules/git.md` next to "After cloning: `make setup`" would keep it findable.

**`consider:` The same facts are asserted in four places.** *(consistency + doc propagation, converged)*
Reflink/cost measurement, "leave `cache.type` unset", and the `dvc gc` hazard appear near-verbatim in the hook comment (~38 lines), `architecture.md` § Data location (31 lines), a test docstring (~16 lines), and ticket 0379's body. A fact stated in four places is a fact wrong in three places after the next correction — and this claim has already been wrong twice. `architecture.md` § Data location is the right single home; the hook comment can keep the operationally-necessary bits (never-clobber invariant, path derivation, why stderr is kept) plus a pointer. Ticket 0379's action 3 already plans to retire part of this, so the state is tracked rather than unnoticed.

**`consider:` The comment-to-code ratio is a style outlier in its own file.** *(consistency, medium)*
The `.venv` block runs ~9 comment lines to ~16 of shell (≈0.6:1); the new DVC block runs ~38 to 5 (≈7.6:1). Largely the same duplication as above. Also: the block opens "Same idiom as .venv above" and only later explains that it deliberately diverges on stderr — a forward pointer would remove the friction.

**`consider:` An existing `.dvc/cache` symlink pointing somewhere deliberate is silently replaced.** *(correctness + red team, converged, medium)*
The guard preserves a real directory but treats any symlink as replaceable, and a *dangling* symlink also satisfies `[ ! -e .dvc/cache ]`. The comment covers the `dvc cache dir <path>` case correctly (that lands in tracked config, so the link is inert), but not a hand-rolled link. Unsupported setup, no instance in this repo — noting it, not blocking.

**`consider:` `dvc gc` is now destructive across every checkout, and the guard is a separate open ticket.** *(red team, high on the hazard, non-blocking on the sequencing)*
Confirmed from `dvc/repo/gc.py`'s reachability model: no race needed — a worktree on a stale `dvc.lock` running `dvc gc` deletes blobs the primary still needs. Mitigating: `grep -rn "dvc gc"` across `Makefile`, `scripts/`, `.githooks/`, and the skills finds nothing, so exposure is a hand-typed command only. Red team's one actionable suggestion for the author: **fast-track 0379 before the next multi-worktree autonomous wave** — a raid/nightbeat session is precisely where a stale worktree and a "let me free some disk" `dvc gc` coexist. Also a noted limit of 0379's planned design: DVC's repo lock lives at `<repo>/.dvc/tmp/lock` and `.dvc/tmp` is *not* shared, so a Makefile refusal catches the sequential wrong-cwd case but nothing stops a live `dvc gc` racing a concurrent reader in a sibling worktree.

**`consider:` Two new tickets filed on a feature branch.** *(scope, medium)*
0379 and 0380 are pure `tickets/*.erg` additions discovered while fixing 0360 — correctly *not* folded into the fix, but the standing scope-discipline rule points them at a separate tickets-only fast-path PR, which this repo explicitly supports. Low severity: they touch no code. The `architecture.md` / `git.md` / `workflow.md` edits are **in scope** by contrast — ticket 0360's own Action 3 requires them.

**`nofollow:` Failed `cd` collapses `SHARED_CACHE` to `/.dvc/cache`, not empty.** *(correctness, medium)* — verified directly; caught today only because the `-d` test happens to fail for a root-level path.
**`nofollow:` The hook ignores `$3` (file- vs branch-checkout).** *(correctness)* — pre-existing, shared with the `.venv` block, idempotent no-op.
**`nofollow:` Missing cross-worktree cache lock is not itself a corruption vector.** *(red team)* — `LocalHashFileDB.add`/`move` are content-addressed with a rename onto the final oid path, so concurrent adds of the same oid race harmlessly onto identical bytes.
**`nofollow:` `deliverables/_shared/_includes/agentic-workflow.md:39`** still credits `.worktreeinclude` alone with worktree data setup. Pre-existing, manuscript prose — author's call, not an agent edit.

---

### Cleared on inspection

`.dvc/.gitignore:1,3`, `.dvc/config.local` contents, the `architecture.md § Data location` heading and both cross-references to it, and ticket 0252's 8-orphan-pointer rationale all check out exactly as cited. Tickets 0379/0380 conform to `%erg 0.1` and describe post-merge reality; `erg check tickets/` PASS (357). Retracted cost claims survive only inside 0360's append-only log, where they belong. No LLMism hits in the added rules prose. `Ticket: none` is correct: the close+archive really is in `1a32ada7`, and `~/.claude/skills/merge/erg-pr-merge:147`'s `tickets/\K\d+` confirms a `tickets/closed/…` path would not match, with the `NOCLOSE` branch at `:152` letting the merge proceed.

### Dissent preserved

> **Correctness:** "*Verdict: approve.* No blockers. The three `consider:`-tagged items are genuine latent fragilities, currently inert."
> **Red team:** "*Verdict: comment.* The one thing worth acting on before merge is optional: consider whether 0379 should land alongside this PR rather than after it, given the hazard is already reproduced, not hypothetical."
> **Scope:** "*Verdict: approve.* No blocking scope violations… a stricter session would have split [0379/0380] into a second fast-path PR."

The author decides. Nothing here blocks the merge; the two `verifiable:` doc/enforcement items (the unpinned `cache.type`, the wrong write mechanism) are the ones I would fix in this PR rather than defer, since both are small and both concern the exact invariant this change makes newly load-bearing.

PR #1162, opened 2026-07-27
feat(0329): state the retrieval protocol in the paper, generated from config

MinhHaDuong:
**Ticket:** tickets/0329-datapaper-retrieval-protocol-in-text.erg

Answers the one finding all four RDJ-26561 external reviewers raised independently: the data paper names its method without specifying it. openai/gpt-5.5 (grinchy) put it as *"the paper must describe the construction protocol at a level that allows methodological evaluation without reverse-engineering code."*

Every number this PR adds to the paper is read from the configuration the harvest used. None is typed.

## What the reviewers asked for, and where it now is

| Hole | Fix |
|---|---|
| Eight target languages never named | §2.1 names them, generated from a new `term_languages` tag on every Tier-1 term |
| Cross-encoder threshold not reported | §2.2, `{{< meta filter_reranker_threshold >}}` |
| Outlier rule: global or stratified? | §2.2 states **global** — one corpus-wide centroid, one global mean/SD (`filter_flags.py`) |
| Near-duplicate thresholds not given | §2.2, three macros; the constants moved from Python into `corpus_filter.yaml` |
| "Key documents" — key according to whom? | §2.1 states ticket 0288's author-validated inclusion rule, settled but never written down |
| Query fields, per-tier terms, grey enumeration | Deposited `tab_retrieval_protocol.{csv,md}`, rendered from config, staged into the Zenodo `data/products/` |

## Three judgement calls worth reviewing

**Action 4 dropped.** A flat CSV of the taxonomy would flatten the Tier 3/4 co-occurrence semantics and become exactly the drifting copy the ticket's own Invariants forbid. Replaced by a §2.1 pointer naming the deposited YAML, which already ships in the archive's `code/`.

**Two columns the ticket asked for are absent.** Harvest date and records returned/retained have no machine-readable source. Hand-typing them would recreate the defect this artifact exists to prevent. The caption says which cells are config-rendered and which describe a hand-exported harvest, so a referee cannot mistake one for the other.

**Config duplication, guarded rather than derived.** `term_languages` restates the Tier-1 term strings as keys. Deriving the term list from a language-keyed block instead would have changed the structure three harvesters read, for no gain the bijection guard does not already provide.

## Guarded, not asserted

`tests/test_retrieval_protocol.py` — 38 tests, fast tier, config and text only:

- `term_languages` against `tiers.1.terms` as a bijection, so the language list cannot drift from the harvest;
- §2.1 names every queried language **and no unqueried one**;
[...]

PR #1188, opened 2026-07-27
test(0363): a render oracle for Quarto meta macros

MinhHaDuong:
Closes the render-oracle gap: an undeclared `{{< meta >}}` key can no longer reach a rendered deliverable without a test failing.

## The defect

Quarto exits 0 on a macro naming a key its metadata does not declare. It writes the literal `?meta:key` into the document and warns only on stderr, so the render succeeds, the suite passes, and only a human reading the finished PDF sees it. The data paper resolves every number it reports through that mechanism.

## Two guards

They fail for different reasons, so both are here.

**Render oracle** (`integration`) asks Quarto itself and reads the output it produced. It renders `--to markdown`, not the document's own PDF format: shortcodes resolve in a Lua filter ahead of writer selection, so both formats resolve identically — review confirmed by rendering both ways and diffing placeholder, warning, and exit code — while only one needs a LaTeX run. 2–8 s per document, ~1 min for the tier. `--output -` writes to stdout, so no rendered document lands beside a tracked source.

**Static resolver** (fast tier, 0.17 s) answers from disk, against the generated `*-vars.yml` — not the `DOC_VARS` dict one layer above it, which is the gap that let this through. It needs no toolchain, so it covers machines where the render skips.

Documents are discovered by glob, not listed, and a separate test asserts the glob found something and found the data paper — an empty parametrize list is a pass with exit 0.

## Proof they can fail

The ticket's invariant is that a check which cannot go red certifies nothing. Nothing below is argued; each was run.

- Synthetic: a document with one declared and one undeclared key. Both guards see it. The render test asserts `returncode == 0` deliberately — that silence is *why* no build step catches this, and a Quarto that starts failing the render will say so here.
- Real: a bogus macro appended to `data-paper.qmd` turned both guards red. Review round 3 went further and planted keys in a shared include and in one nested two levels deep; all three placements were caught, and evasions via code fences, raw LaTeX, `..` traversal and odd whitespace all failed, because Quarto expands inside them too.
- `test_static_resolver_reads_the_generated_vars_file` deletes the vars file and asserts the same document then fails, pinning that resolution comes from the artifact Quarto loads.

## Found live: corpus-report

`corpus-report.qmd` reads `technical-report-vars.yml` but is absent from `compute_vars.DOC_VARS`, so it has no vars file of its own — **12 distinct keys, 22 occurrences**, across `citation-quality.md`, `corpus-construction.md` and others. `test_doc_vars_completeness` is keyed off `DOC_VARS`, so it never saw the document at all.

This upgrades ticket 0357 from "drift risk rather than a live error" to a live error; 0357 gets a log note recording it. The fix needs a Phase-2 regeneration against the corpus, so it is not bundled here.

It is recorded by **pinning the exact 12 keys**, not by an xfail. An xfail says only "this document fails", and review proved by mutation that a *new* unresolved key there still reported the identical `1 xfailed` — the one document with a deferred fix would also have been the one where a fresh defect could hide. Equality makes both directions red, verified by mutation: adding a bogus macro fails with `1 key(s) nothing declares … ['brand_new_missing_key']`, and declaring `corpus_sources` fails with `1 key(s) now resolve — drop them from KNOWN_UNRESOLVED`.

README is corrected too: it listed corpus-report among documents whose numbers "are regenerated by Phase 2", which this branch disproves.

## Coverage boundary, stated plainly

The render oracle skips a document whose gitignored generated includes are not built, naming the file and `make corpus-tables`. Today that is **data-paper**, which needs `tab_corpus_flow.md` — added by ticket 0327 and not yet built anywhere. The static resolver still covers it in full for every include that exists; the blind spot is only macro keys living *inside* an unbuilt include.

## Quarto's reserved keys

`declared_keys()` first counted every top-level front-matter key as resolvable. Quarto keeps some for itself — a header carrying `format:` does not make `{{< meta format >}}` resolve. Probing every front-matter key of all 11 deliverables against a real render: `format`, `metadata-files` and `number-sections` are refused in every document carrying them, the other twelve resolve, and none is ambiguous. `test_quarto_still_refuses_the_reserved_keys` re-renders that set so a Quarto upgrade cannot change the answer silently.

A metadata file that is not a YAML mapping now raises. `set()` over a list yields its elements and over a scalar its characters, so every macro would have looked declared — a false all-clear in the one tier meant to run without a toolchain.

## Also in this branch

- The shortcode scan moves to `tests/_qmd_meta.py`; `test_doc_vars_completeness` now calls it, so the two guards cannot disagree about what a document's keys are. Old and new key sets are identical.
- Nested includes resolve against the **root document's** directory only. The first draft also accepted parent-relative as a "safe superset"; `.claude/rules/architecture.md` (ticket 0359) settled that Quarto does not, so that branch could only admit a file Quarto never reads. Dead in fact too — no include across the 11 deliverables resolves parent-relative but not root-relative.

## Gates

- `make lint` — 191 passed, 6 skipped.
- `make check-fast` — 1327 passed, 11 skipped.
- `make check` — **1979 passed, 8 failed**. Every failure is the same class and none is from this diff: the worktree holds only `.dvc` pointers, so the corpus is absent (`CLIMATE_FINANCE_DATA=data` is relative). Seven are the pre-existing `test_corpus_acceptance` file-existence checks; the eighth, `test_corpus_flow.py::test_generated_flow_closes_to_the_refined_corpus`, arrived with the rebase.

  Worth a follow-up in 0327's code, not fixed here: that eighth test raises `FileNotFoundError` where its two siblings in the same file skip cleanly on "corpus data not present here". One `pytest.skip` guard would make the corpus-absent signal uniform. Happy to file it.
- `erg check` — PASS, 358 tickets.
- `/verify-adherence` — PASS.

Three review rounds, every finding fixed.

## Offered follow-up, not done here

A ratchet asserting that every `ticket NNNN` cited in `tests/` and `scripts/` names a real ticket — round 1 raised it, and round 1's own stale-0357 reference is the case for it. Happy to file it.

**Ticket:** tickets/0363-no-render-oracle-for-quarto-meta-macros.erg


MinhHaDuong (commented):

**Panel verdict: comment**, with dissent preserved. Correctness / Doc-propagation / Consistency returned *comment*; Scope returned *approve*; **Red team returned *request-changes*** on two findings. I reproduced both myself before posting. Nothing here contradicts the ticket's exit criterion — an undeclared key does make the suite fail today, on both guards, verified independently by three agents. The two red-team findings are about the guard's *durability*, which is the same defect class the ticket exists to close, so they get the top of the report rather than the footnotes.

Gates I ran in this worktree: `make check-fast` → **1240 passed, 11 skipped, 1 xfailed** (32.6 s). `make lint` → **180 passed, 6 skipped**. `pytest tests/test_meta_macro_resolution.py tests/test_doc_vars_completeness.py` → **40 passed, 2 xfailed**.

---

### The two the author should rule on

**1. `verifiable:` [red team, corroborated by correctness] Nothing pins that any document was discovered — all three parametrized guards can go vacuously green.**

`deliverable_qmds()` globs `deliverables/*/*.qmd` and nothing asserts the result is non-empty. pytest treats an empty parametrize list as a skip and exits 0. Reproduced:

```
@pytest.mark.parametrize("x", [])
def test_b(x): assert False
→ SKIPPED, exit 0
```

A directory rename or a glob edit silently removes `test_no_deliverable_uses_an_undeclared_meta_key`, `test_document_scan_is_not_silently_empty` and `test_rendered_deliverable_has_no_meta_placeholder` from the suite with no signal. This repo has already renamed a deliverable directory once (`companion-paper.qmd` → `multilayer/`). The module's own docstring makes auto-discovery the design argument — "a hardcoded list silently stops covering the document someone adds next" — which is right, and the missing half is the floor that says discovery still found something. One line: `assert len(deliverable_qmds()) >= 10`.

**2. `verifiable:` [red team, independently found by correctness] `pytest.skip()` inside `xfail(strict=True)` reports SKIPPED, not XFAIL — the corpus-report exemption stops self-destructing on the render side.**

Reproduced standalone:

```
@pytest.mark.xfail(strict=True, reason="known defect")
def test_a(): pytest.skip("include missing")
→ SKIPPED, exit 0   (not XFAIL, not a strict-xfail failure)
```

And on the real target: `deliverables/_shared/tables/tab_corpus_sources.md` is gitignored and untracked, so on a worktree that has not run the analysis targets, `test_rendered_deliverable_has_no_meta_placeholder[corpus-report]` skips before it can ever hold the strict-xfail contract. `make check` does not depend on the analysis targets, so this is the ordinary fresh-worktree path, not a contrived one.

**Mitigation the red team's write-up understates, and the author should weigh:** the *static* side, `test_no_deliverable_uses_an_undeclared_meta_key[corpus-report]`, has no skip path. It holds the strict xfail unconditionally and will go red the moment 0322 lands. So the "exemption cannot outlive the defect" property survives — through one guard of two, on machines without the generated includes. That is weaker than the docstring claims, not absent.

---

### Correctness

- `nofollow:` Root-relative include resolution was traced against the real three-level chain (`technical-report.qmd` → `techrep-zoo.md` → `_includes/zoo/*.md`) and matches Quarto. The old parent-relative code was *dead* on all 18 zoo files — Scope confirmed `os.path.isfile()` returned False for every one of them, so the refactor fixes a silent gap, it does not merely relocate one.
- `verifiable:` **`declared_keys()` misses dotted meta paths.** `META_RE` accepts `{{< meta corpus.total >}}` (`[^\s>]+?`), Quarto resolves it against nested YAML — verified with a live render — but `declared_keys()` collects only top-level keys, so a legitimately-resolvable dotted key false-positives as unresolved. No live instance: `grep` finds no dotted key in the corpus. Latent, and it fails loud rather than silent.
- `verifiable:` **`declared_keys()` ignores project-level `_quarto.yml` metadata**, which Quarto merges (verified live). Red team checked every deliverable: all `_quarto.yml` files carry only `project: {type: default}` and no `_metadata.yml` exists anywhere under `deliverables/`. Latent, same direction (over-flag, not miss).
- `verifiable:` **`test_document_scan_is_not_silently_empty`'s second clause is self-gated.** `if meta_keys_used(qmd): assert declared_keys(qmd)` conditions on the very function whose failure it is meant to detect. Blanking `META_RE` was reproduced: all 11 parametrized instances still passed. The regression is caught elsewhere (`test_static_resolver_*`, the corpus-report xfail flipping to XPASS), so the suite as a whole is not blind — but this test does not do the job its docstring assigns it. *Dissent, preserved:* Scope read the same test as a direct implementation of the ticket's "must be able to fail" invariant and called it requested, not gold-plating. Both are right about different clauses — the `qmd.resolve() in files` assertion is a real guard; the meta-key clause is the no-op.

### Scope

- `nofollow:` Diff is three files, all under `tests/`, no production code. Both ticket-requested guards present, correctly tiered (`make check` runs everything, `make check-fast` excludes `integration` — confirmed at `Makefile:739`/`745`).
- `nofollow:` The `test_doc_vars_completeness.py` refactor is necessary coupling, not creep: the two guards must agree on what a document's keys are, and the PR discloses it. The behavior change is real (regex widened, nesting now transitive and root-relative) but inert on the current corpus — the "six cases pass unchanged" claim was verified, though it holds somewhat fortuitously: the one key in a newly-reached zoo file (`emb_dimensions`) was already found via a different level-1 include.
- `nofollow:` Deferring the corpus-report fix to 0322 is right — `grep` confirms no `corpus-report` entry in `compute_vars.DOC_VARS`, 0322 action 2 is open and on main, and bundling the fix would itself be creep.
- `consider:` pytest XFAIL is not a red result, so for that one document the exit criterion "a test failing" is met only in the strict-xfail-guards-the-exemption sense. The ticket's own invariant is scoped to the data paper, so this is consistent — noted for the record.

### Consistency

- `nofollow:` `_qmd_meta.py` follows the `tests/_*.py` precedent (`_source_roots.py`, `_gfm_render.py`, both confirmed present) including the `require_<tool>()` skip-helper shape. Style-compliant (built-in generics, no `typing` imports, no `print`); `ruff check` clean on all three files. No heavy import, so the conftest auto-`slow` hook does not fire — the fast-tier claim holds.
- `nofollow:` Grepped the repo for `_all_vars_for_doc` / `_find_meta_vars`: zero hits outside git history. Refactor left no stale references.
- `verifiable:` **The branch is 16 commits behind `origin/main`, and two of the missing commits carry the text it cites.** `tests/_qmd_meta.py:66-68` cites `.claude/rules/architecture.md` (ticket 0359) for root-relative resolution and the "13 confident false positives"; that paragraph landed on main in `6c9dbe60` (PR #1185) and the widened 0359 in `e8599673`. Neither is in this branch's tree — the citation is true against main, unverifiable against the PR's own checkout. `mergeStateStatus` is CLEAN and the PR touches neither file, so a rebase resolves it with no conflict risk. Per `rules/git.md` § rebase-at-every-gate, do it before merge.
- `consider:` `KNOWN_UNRESOLVED` says "ticket 0322 action 2 owns registering corpus-report's vars"; that action is conditional in 0322 ("if live: register…"). Slightly overstated, not misleading.

### Doc propagation

- `verifiable:` **The skip message names a Make target that does not exist.** `tests/test_meta_macro_resolution.py:148` says `(run \`make analysis\`)`. `grep -nE "^analysis[a-z-]*:" Makefile` returns only `analysis-figures`, `analysis-tables`, `analysis-stats` — `make analysis` gives `No rule to make target`. A contributor hits this exact message on the fresh-worktree path in finding 2 above, so the two compound. Suggest `make papers` or the three explicit targets.
- `verifiable:` **`README.md:70-75` now makes a claim this PR disproves.** The "Live" bullet names corpus-report among documents whose "numbers are regenerated by Phase 2 (`compute_vars.py`, …) and should track the corpus." They do not — corpus-report is outside `DOC_VARS`, has no vars file, and renders 12 distinct keys as literal `?meta:`. Worth a one-line caveat pointing at 0322, in this PR or a follow-up.
- `nofollow:` **The 22/12 numbers check out exactly.** Two agents recomputed independently — one via `unresolved_meta_keys()`, one via a live `quarto render --to markdown --no-execute --output -` counting real `?meta:` tokens — both landing on 22 occurrences across 12 distinct keys, in `corpus-report.qmd`, `core-vs-full-definition.md`, `citation-quality.md`, `corpus-construction.md`. PR body and code comment both accurate.
- `nofollow:` `--output -` and `--no-execute` confirmed valid against installed Quarto 1.8.27.
- `consider:` `_qmd_meta.py:70` says "Generated tables under `_shared/tables/` are gitignored." Of the five table files actually `{{< include >}}`d, only two are; `.gitignore:48-52` deliberately un-ignores `tab_variables.md`, `tab_venues.md`, `tab_venues_fr.md` so a clean-room build works. The code is right (per-path `is_file()`, no bulk assumption); the comment overgeneralizes.

### Red team — remaining

- `consider:` **"rendering a tracked document leaves no artifact beside its source" is literally false**, though harmless. `cwd=qmd.parent` makes Quarto write `deliverables/*/.quarto/project-cache/` — gitignored at `.gitignore:120-121`, and `git status --porcelain` before/after the integration run is byte-identical. Suggest "no *tracked* artifact". Would fail on a read-only checkout.
- `verifiable:` **"under a second per document instead of a minute" is off by 3–9×.** Measured with `--durations=0`: corpus-report 9.29 s, data-paper 6.26 s, manuscript-Gide 5.51 s, agentic-paper 5.48 s, technical-report 5.42 s, manuscript 5.30 s; full tier 56.3 s first run, 32.6 s warm. The tiering decision is still correct — this rides `make check`, not the fast loop — but the docstring misstates the cost to whoever decides whether to add the 12th document.
- `consider:` **`INCLUDE_RE`/`META_RE` are not code-fence-aware.** A shortcode inside a fenced block is followed as a real include; a nonexistent example path lands in `missing`, which via finding 2 converts the containing document's render test into a skip. No deliverable trips this today (`agentic-paper.qmd`'s inline mention lacks a path argument, so it does not match).
- `consider:` A UTF-8 BOM breaks `front_matter()`'s `\A---\n` anchor, so `declared_keys()` returns empty and every key over-flags. Fail-loud direction; no deliverable carries a BOM. A `.lstrip("﻿")` closes it.
- `nofollow:` Checked and cleared: `metadata-files` globs (Quarto itself rejects them, so no false negative is reachable), nested-path `metadata-files` entries (resolve correctly), the whitespace-split placeholder scan (caught placeholders in table cells, bold text, and adjacent to `()[]`), `--to markdown` vs `--to latex` shortcode parity (both emit the literal `?meta:` prefix), include cycles (the `seen` set holds), and CRLF front matter (`read_text()` normalizes before the regex runs — a first hypothesis here was tested against the raw string, found wrong, and retracted).

---

### Suggested disposition

Findings 1 and 2 are one-line changes in the file they concern and are in the same defect class the ticket closes; the doc-propagation pair (`make analysis`, `README.md:70-75`) is the kind a contributor hits within a week. The three docstring accuracy nits (timing, "no artifact", `_shared/tables/`) are free to fix while the file is open. Everything else is latent with no live instance and belongs in a follow-up ticket if anywhere. Rebase onto `origin/main` before merging so the `architecture.md` citation resolves in the tree that carries it.

MinhHaDuong (commented):
## Multi-perspective review — 5 agents (correctness, consistency, scope, red team, doc propagation)

**Verdict: comment.** No blockers. The two guards do what ticket 0363 asks, are correctly tiered, and the branch already absorbed a review round that fixed the obvious things. Four verifiable minors below, all cheap, plus two resolved disputes recorded for the trail.

Gate run in the PR worktree: `make check` → **7 failed, 1903 passed, 78 skipped, 2 xfailed** (348 s). All 7 are `test_corpus_acceptance` file-existence checks against `data/catalogs/*` — pre-existing, this worktree holds only DVC pointers. Matches the PR body's disclosure. `make lint` → 180 passed, 6 skipped. `ruff check tests/` clean.

Exit criterion verified independently, not taken on trust: a scratch copy of `deliverables/data-paper` + `_shared` under `/tmp` with `{{< meta totally_not_a_real_key >}}` appended made the static resolver return `{'totally_not_a_real_key'}` and the render oracle emit `?meta:totally_not_a_real_key` on stdout with `Unknown meta key` on stderr at `returncode == 0`. Tracked sources stayed clean throughout.

---

### 1. `verifiable:` — two prose claims say ticket 0357 is "not yet on main". It is on main.

`README.md:80` and `tests/test_meta_macro_resolution.py:43` both state 0357 was "filed on the branch of PR #1162 and not yet on main". PR #1162 merged before this branch's own HEAD commit:

```
$ git cat-file -e origin/main:tickets/0357-corpus-report-is-outside-the-vars-registr.erg && echo PRESENT
PRESENT on origin/main
$ git rev-list --count HEAD..origin/main
38
```

The claim is false today and will still be false the moment this merges. Reword to drop the "not yet on main" qualifier in both places. (This branch is 38 commits behind `origin/main` and was never rebased — `rules/git.md` § rebase at every gate. `git merge-tree` reports no content conflict on the four touched files, so a rebase is currency, not conflict resolution.)

*Perspectives: doc propagation (high), consistency (high) — independently converged.*

### 2. `verifiable:` — `declared_keys()` counts front-matter keys Quarto refuses to resolve

`tests/_qmd_meta.py:108` unions **every** top-level front-matter key into the declared set. Quarto's `meta` shortcode does not resolve its own reserved keys, so the static guard would call such a macro declared while the page renders a placeholder. Probed against Quarto 1.8.27:

```
F: **?meta:format**. N: **?meta:number-sections**. T: P.
WARNING (main.lua:17906) Unknown meta key format specified in a metadata Shortcode.
WARNING (main.lua:17906) Unknown meta key number-sections specified in a metadata Shortcode.
```

`format` is in all 11 deliverables' front matter; `number-sections` in `manuscript.qmd` and `manuscript-Gide.qmd`. Failure scenario: someone writes `{{< meta number-sections >}}` in the manuscript → `test_no_deliverable_uses_an_undeclared_meta_key` passes, `?meta:number-sections` ships. The render oracle still catches it wherever Quarto is installed, so this is a **fast-tier-only** blind spot, and no deliverable exercises it today. Cheapest fix: subtract a small reserved-key set, or narrow `declared_keys` to `metadata-files` plus an explicit allowlist.

*Perspective: red team (high, verified). Confirmed independently by the synthesizer.*

### 3. `verifiable:` — on a fresh worktree the render oracle skips exactly the two documents that motivate the ticket

`source_files()` returns unresolvable includes; `test_rendered_deliverable_has_no_meta_placeholder` skips the whole document when any is missing, and `meta_keys_used()` discards them so the static guard scans a partial tree. Three of the 50 scanned files are gitignored generated tables:

```
deliverables/_shared/tables/tab_citation_coverage.md
deliverables/_shared/tables/tab_corpus_sources.md
deliverables/_shared/tables/tab_languages.md

corpus-report   render oracle SKIPS on a fresh worktree, needs all three
data-paper      render oracle SKIPS on a fresh worktree, needs tab_corpus_sources, tab_languages
```

So before `make corpus-tables`, the render oracle covers 9 of 11 documents and skips the data paper — the 77-key document the ticket names. Latent rather than live (those three files carry 0 meta macros today, verified), and the design is deliberate and documented. But the PR body says only "It skips where … a gitignored generated include is absent"; it does not say the skip lands on the data paper. Worth either naming that in the docstring or making the skip loud (an `xfail`/warning rather than a silent skip) so a fresh-worktree run doesn't read as coverage it isn't.

*Perspective: red team (high) — the finding it graded a blocker. Downgraded here: the static guard still covers the other 47 files of those documents, and an ungenerated include cannot reach a rendered deliverable either.*

### 4. `consider:` — `KNOWN_UNRESOLVED`'s reason string mis-attributes the fix

`tests/test_meta_macro_resolution.py:42` says "Ticket 0322 action 2 owns registering its vars". 0322 action 2 reads *"If live: register the `verify_*` / `complete_*` vars for corpus-report's vars file"* — a different variable family, and conditional on action 1's live-vs-v1 decision. The 12 keys actually rendering `?meta:` are 8 `cite_*` and 4 `corpus_*`; the registry wiring that surfaces them is 0357's subject ("Neither branch works without first wiring corpus-report into the registry"). Point the reason string at 0357 and drop or soften the 0322 action number; `README.md:79-81` carries the same imprecision more mildly. Three perspectives flagged this from three different angles.

Note the strict-xfail self-destruct property is **not** at risk: a partial fix leaves `unresolved` non-empty, so the mark stays XFAIL and never XPASSes early.

*Perspectives: scope (medium), doc propagation (medium), consistency (medium).*

### 5. `consider:` — the `--to markdown` equivalence is asserted in a docstring, now empirically confirmed, still unpinned

`tests/_qmd_meta.py:134-139` claims markdown and PDF resolve shortcodes identically because resolution happens in a Lua filter ahead of writer selection. Scope flagged it as unverified; red team then rendered the broken fixture through both `--to markdown` and the real `--to pdf` (xelatex) path and got the identical placeholder and warning. So the claim holds. Nothing in the branch pins it, though — a future Quarto that moved shortcode expansion behind writer selection would silently invalidate the whole cheap-oracle substitution. One `slow`-marked test rendering one document both ways would close it. Optional.

*Perspectives: scope (medium), red team (verified true).*

### 6. `consider:` — unguarded placeholder families

Quarto emits `?invalid meta type:key` with "Unsupported type 'List'" for a declared key whose value is a multi-element list — matching neither `PLACEHOLDER` nor `WARNING`. Not live (all vars files are flat scalars) but structurally outside both guards. A one-character change (`PLACEHOLDER = "?meta"`, or an alternation) would cover it. Same tier: `META_RE`'s `[^\s>]+?` cannot match a key containing `>`, which Quarto resolves anyway — contrived enough to be `nofollow:` on its own, but it stacks with finding 3.

*Perspective: red team.*

---

### Dissent and disputes, resolved

**The 22-vs-12 numbers are not a contradiction.** Consistency graded the `KNOWN_UNRESOLVED` comment's "renders 22 placeholders" as wrong. Correctness and doc propagation independently computed both figures and found them different, correct measures. Confirmed here:

```
distinct unresolved keys: 12
total occurrences:        22
```

(15 occurrences in `citation-quality.md`, 3 in `corpus-construction.md`, 3 in `core-vs-full-definition.md`, 1 in `corpus-report.qmd`.) PR body, README (12 keys) and the test comment (22 placeholders) are each accurate. No change required — though writing "22 occurrences of 12 distinct keys" in the comment would have saved a reviewer round.

**Red team returned `request-changes`; the other four returned `approve`/`comment`.** Its two grounds are findings 2 and 3 above, both confirmed and both retained — downgraded to minors because each has a second guard covering it in the configuration the author actually renders in. Recording the disagreement rather than flattening it: a reviewer with a stricter reading of the ticket's "a check that cannot go red certifies nothing" invariant would block on finding 3.

### Verified sound, no action

- **Nested-include resolution against the root document is correct**, and exercised by real structure: `technical-report.qmd` → `_shared/_includes/techrep-zoo.md` → `zoo/*.md` (18 files); resolving against `current.parent` would undercount by 18. Matches `.claude/rules/architecture.md` (0359) verbatim.
- **The `test_doc_vars_completeness.py` refactor is behavior-preserving**, proven not asserted: old vs new key sets recomputed per document — `technical-report` 2/2, `data-paper` 70/70, `multilayer-detection` 32/32, identical.
- **Tiering**: static tests fast (no `HEAVY_MODULES` import, no auto-slow-mark), render tests `integration` with the marker declared in `pyproject.toml`, `make check` runs unfiltered so they execute. Timings check out: 4.55 s for one document, 51.83 s for the 11-document tier, against the docstring's "2-8 s / ~1 min".
- **`--output -` leaves no artifact**: `.quarto/` is gitignored (`.gitignore:114,121,142`), `git status` clean after the full run.
- **Helper-module convention** matches all five `tests/_*.py` siblings, including `pytest.skip()` in a helper (`_gfm_render.py::require_pandoc`).
- **No second source of truth**: `META_RE`/`INCLUDE_RE` exist only in `_qmd_meta.py`; `_all_vars_for_doc`/`_find_meta_vars`/`_find_includes` have no remaining callers repo-wide.
- Front-matter edge cases (CRLF, BOM, no trailing newline) only ever shrink the declared set — they fail toward false positives, never false negatives. Cyclic includes terminate. xdist `-n 4` poses no same-document render collision (each deliverable parametrized once).

🤖 Review by 5 parallel agents via `/review-pr`

MinhHaDuong (commented):
## `/review-pr` — five-agent panel (correctness, consistency, scope, red team, doc propagation)

**Verdict: comment.** No blockers. Every panel member reached approve or comment; the guards were attacked directly and held. Three findings carry attached evidence and are worth fixing before merge; the rest are recorded for the trail.

**Gates re-run on the branch (this review's own run, not the PR body's):**
- `make lint` — 191 passed, 6 skipped
- `make check-fast` — 1324 passed, 11 skipped, 1 xfailed (26.7 s)
- `tests/test_meta_macro_resolution.py` + `tests/test_doc_vars_completeness.py`, all tiers — 41 passed, 1 skipped, 2 xfailed (21.6 s). Quarto 1.8.27 present, so the integration tier really ran; the one skip is `[data-paper]` on the unbuilt `tab_corpus_flow.md`, exactly as the PR body states.

### What the red team could not break

Worth stating first, because it is the ticket's whole point. **Mutation test, three placements** — a bogus `{{< meta definitely_not_a_key >}}` in `data-paper.qmd` (direct), in `_shared/_includes/techrep/overview.md` (shared include), and in `_shared/_includes/zoo/S1_mmd.md` (nested two levels through `techrep-zoo.md`): all three turned `test_no_deliverable_uses_an_undeclared_meta_key` red, then restored clean. Code fences, inline spans, HTML comments, LaTeX raw blocks, `..` traversal in an include path, and multiline/tab whitespace inside the shortcode were all tried as evasions — Quarto expands inside every one of them and the static regex tracks it identically, so static and render oracle agreed in every case. Rendering a real deliverable left only the already-gitignored `.quarto/`; no tracked file touched. The integration tier under `-n 4` (matching `make check`) showed no races across ~8 runs, including the two directories that hold two `.qmd` sharing one cache.

Doc propagation also verified the design's load-bearing Quarto claim empirically rather than taking it on trust: a synthetic broken `format: pdf` probe rendered `--to markdown` and `--to pdf` produced the identical `?meta:absent_key`, the identical stderr warning, and the identical exit 0. The cheap-render substitution is measured, not assumed.

---

### Findings

**1. `verifiable:` — the "77 macro keys" count is stale; actual is 75 distinct / 87 occurrences.**
`tests/test_meta_macro_resolution.py:123` — "the data paper is named outright because its 77 macro keys are why this suite exists." Two panel members flagged it independently and I re-verified in the worktree:

```
distinct keys (include tree): 75
occurrences (include tree):   87
```

`len(DOC_VARS["data-paper"])` is also 75. The number traces to ticket 0363's filing text; `data-paper.qmd` is untouched by this diff, so it was already stale against `origin/main` when the docstring was written — most likely 0327's corpus-flow macros landed after 0363 was filed. Same root cause makes the PR body's parity line wrong: "2/2, **70/70**, 32/32" should read **75/75**. The parity *claim* is true — two reviewers reimplemented the pre-diff algorithm and got identical key sets for all three `DOC_VARS` documents — only the cited number is off. A test whose subject is "numbers that rot silently" should not ship a rotted number in its own docstring.

**2. `verifiable:` — the `KNOWN_UNRESOLVED` xfail is content-blind, so a *new* corpus-report defect is invisible until 0357 lands.**
`tests/test_meta_macro_resolution.py:40`. `xfail(strict=True)` asserts only that the test fails, never *why*. Mutation-proven: adding an unrelated bogus key to corpus-report's reachable include still reports the identical `1 xfailed`. Since 0357's fix needs a Phase-2 regeneration and is explicitly deferred, that exemption will sit there a while, and anything else that breaks in corpus-report during the wait passes review silently. Asserting the unresolved set *equals* the known 12 keys — rather than merely being non-empty — restores the signal and keeps the self-destruct property.

**3. `verifiable:` — `declared_keys()` does not validate the vars-file shape, and the fast tier is the one that gives a false all-clear.**
`tests/_qmd_meta.py:133`, `set(yaml.safe_load(_read(path)) or {})`. A list-typed `*-vars.yml` silently becomes a set of its items treated as declared keys; a bare scalar string becomes a set of its *characters*; a list-of-dicts raises an unhandled `TypeError`. A real render hard-fails loudly (rc=1) on all of these — but the static resolver is precisely the guard documented to run where the toolchain is absent, so that is the tier where a malformed generated artifact reads as "all clear." An `isinstance(loaded, dict)` assertion with the path in the message costs one line.

**4. `consider:` — `declared_keys()` never reads `deliverables/*/_quarto.yml`.**
Quarto *does* merge project-level metadata (including `metadata-files:`) into a document's resolvable `{{< meta >}}` namespace — confirmed by building a throwaway project and resolving a project-level key. All nine `_quarto.yml` in this repo are today exactly `project: {type: default}` (verified), so nothing is live, and the render oracle is immune since it shells out inside the deliverable's own project dir. The direction is also safe: the static guard would over-flag, never under-flag. A one-line docstring caveat on `declared_keys()` is the whole fix if you want one.

**5. `consider:` — dotted keys are a latent false positive.**
`{{< meta a.b >}}` genuinely resolves against nested YAML in Quarto, but `declared_keys()` sees only top-level keys, so a future document using one would be wrongly flagged. Grep confirms no current usage. Same safe direction as #4.

**6. `consider:` — ticket 0357's body still carries the assessment this PR disproves.**
`tickets/0357-…erg:31` reads "The values agree with config today, so this is drift risk rather than a live error." This branch establishes 12 distinct keys / 22 occurrences rendering as live placeholders — I re-derived both numbers independently and they match the `KNOWN_UNRESOLVED` comment digit for digit. The diff touches no ticket file. Per this project's ticket discipline the upgraded assessment belongs as a log line on 0357, where the next reader of that ticket will find it, not only in a merged PR description.

**7. `consider:` — `QUARTO_REFUSES` is provably inert against the current corpus.**
Measured across all 11 deliverables: `meta_keys_used ∩ front_matter_keys − vars_file_keys` is empty for every document. No deliverable resolves a macro against a front-matter key at all. Ticket action 3 asks only for resolution "against the generated vars YAML directly"; reading front matter too is what creates the need for the frozenset, its comment block, and a dedicated integration test. Defensible as forward-looking correctness — Quarto really does expose `title` and friends — but it is surface the current corpus never exercises, and `test_quarto_still_refuses_the_reserved_keys` is itself integration-gated, so drift on a Quarto upgrade would not be caught in the fast loop either. Noted as a follow-up simplification candidate, not a change request.

**8. `nofollow:` — `_FRONT_MATTER_RE` returns `{}` on CRLF or a UTF-8 BOM.** All 11 `.qmd` are LF with no BOM (checked), and the failure mode is loud (empty declared set → every key flagged), not silent.

**9. `nofollow:` — `unresolved` in `source_files()` is not deduped.** No duplicates occur today and the single caller reads `missing[0]`.

**10. `nofollow:` — `test_document_scan_is_not_silently_empty` overlaps `test_discovery_finds_the_deliverables`.** Both guard "the guard cannot vacuously pass," at different granularity. Both are sub-second. Judged earned by two reviewers, marginal by one; recorded, not pursued.

---

### One observation, cause not established

During the panel, one reviewer saw a single non-reproducible failure of `test_no_deliverable_uses_an_undeclared_meta_key` across the three documents that read `../_shared/technical-report-vars.yml`, reporting keys present in no tracked file. Four isolated reruns were clean, the tree stayed clean, and my own runs above are green. Five review agents shared this one worktree concurrently, which is the likelier explanation than a defect in this code — but I am reporting the observation rather than a verdict. Worth a second look only if it recurs outside a shared-worktree session.

### Dissent, preserved

Scope and correctness read the front-matter-plus-`QUARTO_REFUSES` design as forward-looking correctness that is well documented and well tested; scope separately reads it as beyond the ticket's literal action 3 and a candidate for trimming. Both positions are in finding 7 as stated. The call is the author's.

🤖 five-agent panel via `/review-pr`

PR #680, opened 2026-04-16
ticket 0064: fill companion paper §5 Results with real-corpus outputs

MinhHaDuong:
## Summary
- Wave C.2 of epic 0026. Splits the numerical fill of §5 Results from the structural rewrite in ticket 0057.
- Blocked-by 0042 (pipeline outputs on real corpus) and 0057 (scaffolding + methods prose in place).
- Scope: six lead methods (S2 energy, L1 JS, G9 community + C2ST PCA-32, C2ST TF-IDF, G2 spectral). Report Z-scores, bootstrap CIs, permutation p-values, transition zones, validated zones, interpretation per zone.

## Test plan
- [ ] Read each `tab_summary_{method}.csv` once 0042 completes.
- [ ] Every `{{< meta X >}}` in §5 resolves via `companion-paper-vars.yml`.
- [ ] All four figures cross-referenced at correct subsections.
- [ ] `/review-pr-prose` AI-tells auditor passes.
- [ ] `quarto render content/companion-paper.qmd` clean.
- [ ] `make check-fast` passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1190, opened 2026-07-27
fix(0364): select HAL credentials, delete the shadowing skill fork

MinhHaDuong:
Ticket-ref: tickets/0364-hal-credentials-are-not-selected-by-keys.erg

*Deliberately `Ticket-ref:`, not `Ticket:`.* Merging this deletes the project fork, which makes the harness `SKILL.md` the only `update-publist` document — and it still carries the stale credential-source line until harness PR [#680](https://github.com/MinhHaDuong/ImperialDragonHarness/pull/680) merges. Closing 0364 here would claim a fix that has a regression window in it. 0364 closes when #680 lands.

## What was wrong

`.claude/skills/update-publist/SKILL.md` said its credentials come "from `.env`".
Neither `HAL_ID` nor `HAL_PASSWORD` was ever there, and since 0343 `.env` holds no
credential value at all — it carries a default-deny `KEYS=` line naming which of
`~/.config/keys/*` this project may load. `hal` was not named, so nothing loaded
them and the HAL SWORD deposit could not authenticate. The failure reads as an
ordinary auth error, so the cause is invisible from the message.

## What changed

**Selection** (machine-local, not in the diff — `.env` is gitignored): the `KEYS=`
line gained `hal:HAL_ID,hal:HAL_PASSWORD`, one entry per name rather than a bare
`hal`, matching the least-privilege form the other four use.

**Deletion.** `update-publist` is a *harness* skill; this copy was a stale fork
shadowing it, and the two had drifted into disagreeing on fact — this one puts
`Ha-Duong.bib` at `~/CNRS/html/`, the harness copy at `src/`, and only the harness
copy has the pre-upload payload-review gate and the credential-safety rules for
passing secrets to curl. Repairing the fork would have kept both alive to drift
further, so it is deleted and the better document survives. Its own stale line is
fixed in harness PR MinhHaDuong/ImperialDragonHarness#680.

**Ratchet.** `STALE_ENV_CLAIMS` gains both phrasings — the fork's and the harness
copy's — so neither can be pasted back in. The second is anchored on the credential
name rather than cut at "from the project `.env`", because `.env` legitimately
supplies non-secret settings and the shorter form would eventually fire on a true
sentence about `CLIMATE_FINANCE_DATA`.

**What is deliberately absent:** `REQUIRED_KEYS_EXPORTS` does *not* gain the HAL
names. With the fork gone, no code in this repo reads them; that list means "this
project breaks without it", not "this machine can load it".

## Verification

Credentials were only ever compared **by hash** — no value was printed at any point.

- `HAL_ID` / `HAL_PASSWORD` resolve through the bash `BASH_ENV` loader from three
  startup cwds (`$HOME`, the repo root, this worktree), each matching
  `~/.config/keys/hal.env` by SHA-256 prefix. The `$HOME` case exercises the
  harness-level selection added to `~/.claude/.env`, so a `/update-publist` run
  from any directory authenticates.
- The narrowed ratchet pattern matches the pre-fix harness `SKILL.md` (1 hit) and
  not the fixed one (0 hits); the fork's phrasing matched the pre-fix file at
  `HEAD` and matches nothing tracked now.
- `make lint`: 180 passed, 6 skipped.
- `make check`: 1843 passed, **7 failed**, 78 skipped. All 7 are
  `tests/test_corpus_acceptance.py` file-existence checks — the corpus lives in
  the primary checkout and the worktree holds only `.dvc` pointers. Same 7 test
  ids as the pre-change baseline; this diff touches no corpus code.
- `erg check tickets/`: PASS (351 tickets).
- `/verify-adherence`: PASS. Its one nit (a ticket citing a test id that does not
  exist) and its advisory on the over-broad pattern are both fixed in this commit.

Not rebased onto current `origin/main`: force-push is guarded on this branch and
the bases do not overlap (no file in this diff was touched upstream), so per the
git rule it merges as-is rather than being rewritten after push.


MinhHaDuong (commented):

Panel: correctness, consistency, scope, red team, doc propagation (all `sonnet`, decorrelated from the coding tier). Verdicts split **3 approve/comment · 1 request-changes**, so dissent is preserved below rather than averaged away.

Tests re-run in the PR worktree: `make lint` → **180 passed, 6 skipped**; `make check-fast` → **exit 0**. Matches the PR body. The 7 `test_corpus_acceptance.py` failures the body reports under `make check` are the known worktree/DVC-pointer artifact, not this diff.

The credential work itself is correct and well-evidenced. `hal:HAL_ID,hal:HAL_PASSWORD` parses cleanly through the `SRC=DST` branch of `test_keys_line_selects_every_consumed_credential`; both new `STALE_ENV_CLAIMS` strings were confirmed to match their intended source character-for-character and nothing else; the loader-precedence claim that justifies the whole design was independently verified against `~/.claude/scripts/bash-env.sh:145-148,216-217,229` — a project `KEYS=` line does **replace** the harness one. No credential value was printed at any point in this review.

---

### Contested blocker — ticket close outruns the harness PR it depends on

**Scope (request-changes):** the body carries `**Ticket:** tickets/0364-…`, which `erg-pr-merge` treats as an unconditional close. But the ticket's own exit criterion — "one surviving skill document names that source correctly" — is false today, and its last verification box is `[ ]` for exactly that reason.

**Correctness (`consider:`):** workflow policy permits unconditional close regardless of checkbox state; the tracking-ticket convention merely *suggests* keeping a parent open until its child merges.

Verified directly:
- `gh pr view 680 --repo MinhHaDuong/ImperialDragonHarness` → **`state: OPEN`**.
- `~/.claude/skills/update-publist/SKILL.md:56` still reads ``\`HAL_ID\` and \`HAL_PASSWORD\` from the project \`.env\``` — the identical defect class this ticket exists to kill.

Merging this PR deletes the fork, which makes the harness copy the *only* `update-publist` document — and it is the one still carrying the stale claim. Between the two merges there is a real regression window, not just documentation lag. Two cheap remedies: merge #680 first, or swap to `Ticket-ref:` and close 0364 when #680 lands. Author's call.

---

### Knowledge lost by the deletion — belongs to #680, flagged before it is gone

`verifiable:` **CIRED struct-id contradiction, unresolved.** The deleted fork said `struct-1002424`; the surviving harness copy says `struct-1380080` (`~/.claude/skills/update-publist/SKILL.md:73,85`). #680's diff does not touch that line. Deleting the fork removes the only record that a discrepancy ever existed, and a future TEI deposit will use whichever ID happens to survive with nothing to check it against. Highest-value item in this section — a live correctness risk for the next HAL deposit. (doc propagation, high confidence)

`verifiable:` **Four operational details present in the fork, absent from both the harness copy and #680's diff:** the `-H "X-test: 1"` dry-run step (fork:65), the `PUT .../sword/{hal-id}` update-an-existing-deposit recipe (fork:75), the ORCID `0000-0001-9988-2100` (fork:57), and the HAL domain codes `shs.eco` / `shs.hisphilso` (fork:60). The four sections the ticket log *does* claim were ported (entry-type table, PDF review gate, `halJournalId` lookup, STATE.md tie-back) are genuinely in #680 — that claim checks out. (doc propagation)

`consider:` The ticket's "Scope change" note characterizes the fork/harness disagreement as the bib-path and the payload-review gate only. It omits the struct-id conflict and the four items above, so a reader trusting the ticket would believe #680 already covers everything. (doc propagation)

---

### The sharpest question — blast radius, and why it is not a blocker here

`consider:` **`HAL_PASSWORD` now reaches every bash subprocess spawned from this repo**, for a skill no code in this tree calls. The module docstring already concedes the mechanism ("a selected credential still reaches every bash subprocess") — but it wrote that clause about *this project's own* credentials. A science-paper repo now permanently carries a foreign SWORD password in every subprocess environment, for a feature invoked rarely and by a human.

Red team reached a definite position, and I agree with it: this is a genuine broadening relative to 0343's default-deny intent, but **not a regression this PR introduces**. The `.env` edit is machine-local and already applied outside the diff; blocking the PR would not undo the exposure, it would only make `REQUIRED_KEYS_EXPORTS` lie about what the local `.env` does — defeating the exact silent-drift guard 0343 built. The sound fix is architectural and out of scope: either make `bash-env.sh` compose harness and project `KEYS=` additively instead of overriding, or have `update-publist` source `hal.env` inside its own SWORD step, scoped to that call. **Worth a follow-up ticket** before the next project needs a harness-only credential and hits the same friction.

---

### Guard hardening — real gaps, none introduced by this diff

`consider:` **The ratchet is weaker than the body discloses.** `test_no_doc_claims_env_supplies_credentials` does `if claim in line` — per physical line, case-sensitive, exact substring. The body admits only that "a partial paste slips through". Full, freshly-worded restatements evade cleanly: lowercase `credentials from \`.env\``; backtick-free `Credentials from .env`; any line-wrapped rendering (no single line holds the substring); and ``HAL_ID and HAL_PASSWORD are read from the project's \`.env\``` misses the second pattern entirely on apostrophe-s alone. The file's own header calls these claims "lexically stable" — case and rewording variants falsify that. A normalized whole-document scan would close most of it. (red team, high confidence)

`consider:` **`SCAN_EXTENSIONS` allow-list has holes.** It misses tracked `.txt` (9 files), `.tex` (1), and `.json` (6). Grepped all three — no live stale claim hiding today, so this is structural rather than active. But the constant's own rationale invokes auto-discovery "for exactly this reason"; an allow-list of text extensions is the same shape of bet that already failed once on `dvc.yaml`. (red team, medium confidence)

`consider:` **`test_shape_heuristic_accepts_the_real_non_secret_settings` hand-copies a literal of the real `KEYS=` line and never reads `.env`.** When the real line changes it keeps passing against a stale synthetic sample — never fails loudly — and its docstring ("The live `.env` non-secret values must not trip the heuristic") becomes false advertising. Pre-existing pattern; this diff extends it rather than introducing it. Parsing `.env` when present, with the synthetic sample as fallback, would remove the false-confidence framing. (red team; correctness confirmed the literal matches the live line *today*, so no drift yet)

---

### Documentation and style

`consider:` **`AGENTS.md` § Credentials is narrower than the mechanism.** It describes default-deny per-provider selection but says nothing about override-vs-merge precedence between the project and harness `.env` — the exact subtlety this ticket exists to capture, currently documented only in a test-file comment (`tests/test_env_has_no_secret_literals.py:61-64`). Nothing in AGENTS.md is *false*; it is just the more discoverable home for the fact. Raised independently by consistency and doc propagation. (medium confidence)

`consider:` `# idem` (`tests/test_env_has_no_secret_literals.py:77`) is the only occurrence of the word in the tracked tree; every other multi-name comment in this file spells the reason out per entry. (consistency, low)

`nofollow:` Prose in the two expanded comment blocks matches the file's established rationale-heavy voice — no antithesis reversal, no tricolon, one em dash per paragraph. Length is consistent with the existing `TOKEN_SPLIT` and auto-discovery blocks. Not a defect.

`nofollow:` `AGENTS.md`'s two table edits are complete — no dangling row, no orphaned sentence. Repo-wide grep for `update-publist` leaves only the test-file comments and the ticket body, both correct to retain. `.claude/settings.json`, `.worktreeinclude`, `STATE.md`, `ROADMAP.md`, `.agent/runbooks/` reference nothing deleted.

`nofollow:` Ticket conforms to `%erg 0.1` — no `Status:` header, both log entries well-formed and appended rather than edited. `erg check tickets/` PASS (351 tickets). Body-section rewrites are compliant: the append-only rule binds log entries only, and superseded content stays recoverable via `git log -p`.

`nofollow:` **Scope-change authorization is self-certified.** The ticket note and the PR thread are the only trace that a human made the delete-over-repair call, and `AGENT_GH_TOKEN` posts as the repo owner, so agent- and human-authored content are indistinguishable on the forge. A standing harness gap, not misconduct in this PR. (scope)

---

### Recommendation

The code is sound and the verification in the body holds up under independent re-run. One decision is owed before merge: **the ticket-close claim versus the still-open #680**. Everything else is a follow-up — of which the CIRED struct-id contradiction is the one that should not wait, because the fork's copy disappears with this merge.

🤖 `/review-pr` · five agents · no credential value displayed

PR #1191, opened 2026-07-27
fix(0370): escape every cell the Markdown table emitters interpolate

MinhHaDuong:
**Ticket:** tickets/0370-export-corpus-table-interpolates-raw-values.erg

> Renumbered 0367 → 0370 on the base branch (its commit `4a943dc4`): PR #1176
> took the 0367 seat. Second renumber on this ticket — it moved 0364 → 0367
> against #1140 earlier. In-code references follow in `3d2bfdd8`.

Stacked on #1172 — ticket 0367 exists only on that branch, so this PR
targets `t0340-markdown-table-cell-count-guard` and GitHub will retarget it
to `main` when #1172 merges. The diff below is this change alone.

## The defect

`export_corpus_table.py::_write_md_table` joined `str(v)` values with ` | `.
A value carrying a pipe ends its cell, and GFM does not error — it drops the
overflow and shifts every later value one column left. Fourth instance of the
class: 0325 (deposit codebook), 0339 (both venue tables), now this (0370). First one
found by reading an emitter rather than by a reader noticing a broken table.

The 0340 sweep cannot see it: its target `tab_corpus_sources.md` is untracked
and absent without corpus data, so the check skips it.

## The fix

Every column routed through `markdown_text_cell`, not only `Source`. A
per-column branch would make each emitter carry a claim about which of its
columns can hold a pipe *today*; the helper exists so no emitter has to know.
The TOTAL row escapes first and wraps in `**…**` second — the reverse hands
the emphasis markers to the escaper.

## Action 3 sweep

Driven by the 0340 discovery surface (`generated_markdown_targets()`), not a
hand list. It returns **10** targets on current `main` — the sweep first ran
against 8, and `tab_retrieval_protocol.md` and `tab_corpus_flow.md` arrived
with ticket 0327 while this PR was open. All 10 accounted for:

| Target | Emitter | Status |
|:--|:--|:--|
| `codebook.md` | `export_codebook.py` → `_deposit_variables` | `markdown_cell`, covered by 0325 |
| `tab_variables.md` | `export_variables_table.py` → same path | covered by 0325 |
| `tab_venues.md`, `tab_venues_fr.md` | `export_tab_venues.py` | `markdown_text_cell`, covered by 0339 |
| `tab_core_venues_top10.md` | `export_core_venues_markdown.py` | `markdown_text_cell`, covered by 0339 |
| `tab_retrieval_protocol.md` | `export_retrieval_protocol.py` | `markdown_text_cell`, already escaped when it landed |
| `tab_citation_coverage.md` | `export_citation_coverage.py` | every value coerced through `int` or a numeric format — no free text; recorded in the docstring rather than escaped, matching how the venue emitters already record the same judgement |
| `tab_corpus_sources.md` | `export_corpus_table.py` | **fixed** — this ticket |
| `tab_languages.md` | `export_language_table.py` | **fixed** — second live instance found by the sweep |
| `tab_corpus_flow.md` | `export_corpus_flow.py` | **fixed** — arrived with 0327 after the sweep; `Stage` is hand-authored prose interpolated raw |

The exit criterion is checked against `main` at merge, not against the tree the
sweep first saw, which is why the third fix is in this diff rather than in a
follow-up.

The language table is a genuine second instance, not a precaution:
`normalize_lang` returns any two-character value unchanged, so a malformed
corpus `language` field reaches the `Code` column raw and the `Language`
column upper-cased. Neither is curated in-repo.

## Tests

Red first, five tests, all asserting on the **rendered** page via
`tests/_gfm_render.py` and driving the real emitter (not the helper — a test
that calls the helper passes even after the emitter stops calling it):

- `test_pipe_bearing_source_keeps_its_eight_cells` — the row survives intact
- `test_total_row_stays_bold_around_an_escaped_value` — escape-then-wrap order
- `test_shipped_labels_are_untouched_by_the_escaper` — no-churn invariant: the
  real `SOURCE_META` labels must emit no backslash at all, so regenerating
  `tab_corpus_sources.md` on a corpus machine is a no-op
- `test_pipe_bearing_language_code_keeps_its_four_cells` — the sweep finding
- `test_total_row_keeps_its_emphasis` — `**Total**` is authored markup; the
  escaper touches `\`, `|` and `` ` ``, never the asterisk

Red evidence before the fix: `<td>**TOTAL</td> <td>all sources**</td>` for the
corpus table, and `<td>A</td> <td></td> <td>a</td> <td></td>` for the language
table — both counts dropped entirely.

## Gates

- `make check-fast` — 1218 passed, 11 skipped
- `make lint` (adherence) — 212 passed, 10 skipped (one pre-existing
  `8 scripts exceed 500 lines` warning, unrelated)
- integration tier — 361 passed, 7 skipped (15 min)
- `/verify-adherence` — PASS, no mechanical failures

No CI in this repo; every gate above is local.

## One finding the gate raised, deliberately not fixed here

`deliverables/_shared/tables/tab_corpus_sources.md` and `tab_languages.md` are
gitignored, while their siblings `tab_variables.md`, `codebook.md` and
`tab_venues.md` are force-tracked by explicit negations — and the `.gitignore`
comment even claims the `.csv` is "grouped with tab_corpus_sources.md" before
negating only the `.csv`. Both files are `{{< include >}}`d into `data-paper.qmd`
and `corpus-report.qmd`, so by the handoff-artifact rule they should be tracked.
The consequence is exactly the blind spot this ticket describes: the 0340 shape
guard skips these two files on any machine without corpus data.

Tracking two generated artifacts is a different change from escaping their
emitters, so it gets its own ticket rather than riding this diff:
**filed as 0371 in #1197**, together with 0375 below.

## Review round 1 (#1191 panel: 3 approve, 2 comment, no dissent)

The panel reproduced the pre-fix defect by driving the old `_write_md_table`
through real `pandoc -f gfm` (a piped source label drops `%Refs`), then failed
to break the post-fix emitter with embedded newlines, `---`, leading `#`, HTML,
a lone trailing backslash, fullwidth-pipe U+FF5C, pre-escaped `a\|b`, and
`a||b`. It verified the 8-target sweep claim by running
`generated_markdown_targets()` itself.

Three actionable findings, all addressed in `decd0386`:

1. **The promised tracking ticket did not exist** — now filed, #1197 (0371).
2. **`_markdown_table.py`'s docstring still named only the venue-table
   emitters.** It is the one documented home of the escaping convention and
   this PR adds two callers; the docstring now names them and the input
   contract each brings. Fixed in this diff, as the panel asked.
3. **A zero-refined source ships a literal `nan`** — `row.get(c, "")` never
   fires because `pd.DataFrame` puts NaN under a key that *is* present.
   Pre-existing, latent, and a different concern from escaping: **filed as 0375
   in #1197**.

Two `consider:` items were also taken rather than waved through:

- The two emitters expressed "keep the total row bold" differently. The
  language table left `**Total**` in the row data and escaped it whole — correct
  only because `*` is outside the escaper's character set, which makes the
  emphasis hostage to a later widening of that set in a file nobody would
  re-read for it. It now stores the label plain and wraps after escaping, the
  same order as `export_corpus_table.py`. Only the label cell is wrapped:
  bolding the counts would churn the shipped table. **Verified byte-identical**
  against the previous emitter on a 15-language input.
- The language table had no no-churn pin while its sibling did. Added on the
  same shape — drive the real script over every `LANGUAGE_NAMES` value, assert
  the emitted file carries no backslash at all and exactly the four asterisks
  of the Total row.

## Review round 2 — one blocker, fixed in `537b9f4c`

**Makefile prerequisite drift, introduced by this diff.** Both emitters now
import `scripts/_markdown_table.py`, but their targets did not list it, making
them the only two of six such targets missing it. The PR's own comments argue a
widening of the escaper's character set is foreseeable; without the
prerequisite, that widening would leave these two tables stale. Also tightened
the `export_citation_coverage` docstring, whose "coerced through int" claim
glossed the Period label composing two ints around an en-dash.

The panel also proved a wider gap and I filed it rather than folding it in:
**the escaper is calibrated to GFM, but Quarto renders through pandoc's
`markdown` reader**, where `@`, `~` and `^` are live syntax. Reproduced here —
`@smith2020` renders literally under `-f gfm` and becomes
`<span class="citation">` under `-f markdown`, and both consuming `.qmd` files
set `bibliography:`. The sharper half: every escaping test since 0325 asserts
through `-f gfm`, so the suite would stay green while the manuscript carried
the defect. It predates this diff and reaches all six emitters through the
shared helper — the 0339 venue emitters interpolate corpus journal names, a far
likelier `@` carrier than a hand-edited source label. **Filed as 0376 in
#1197**, ordered oracle-first.

## Review round 3 — no blockers

3 approve, 2 comment, no dissent. The panel independently reproduced red-before-fix
by reverting each emitter, ran `generated_markdown_targets()` itself to check the
8-target claim, grepped all of `scripts/` for pipe-table interpolation and found
nothing the sweep missed, and failed to break the escaper with non-string types,
U+2028/2029/NEL/VT/FF, a delimiter-look-alike data row, cross-cell backtick
pairing and RTL/ZWJ marks.

Two one-line findings taken in `dfd3210a`, both about the tests:

- `tests/_gfm_render.py` — `row_with` returned `matching[0]`, so an ambiguous
  needle silently substitutes a decoy row and the caller asserts on cells it
  never meant to read. It now fails loudly. Verified against all three consuming
  suites, including the merged venue-fidelity one.
- The language table's no-churn pin called `require_pandoc()` while never
  rendering — it reads the emitted bytes, so the gate only made it skip where
  its sibling runs.

## Merge with `main`, and a ninth emitter

#1172 merged, so this PR retargeted to `main` and picked up ticket 0327. Two
conflicts, both resolved toward main's version with this branch's contribution
re-applied on top: the Makefile rule now passes an explicit `.csv` path (0327's
grouped-target `$@` fix) *and* declares `scripts/_markdown_table.py`; the
`CAPTION` constant became 0327's `build_caption()`, carrying this branch's note
on why the caption is deliberately unescaped, extended to cover its six new
interpolations (all `int`-formatted). Git auto-merged `_write_md_table`, so the
escaping and 0327's new `caption` parameter compose.

0327 also added `export_corpus_flow.py`, a **ninth emitter with the same
shape**: `Stage` is prose authored by hand in `compute_corpus_flow` and
interpolated raw. It is fixed here, with the Make prerequisite and both pins
its siblings have. Landing it in a follow-up would have merged this PR with its
own exit criterion false.

## Review round 4 — no blockers

2 approve, 3 comment, no dissent. The panel re-derived rather than trusted:
red-before-fix reproduced from `git show origin/main:` copies, the no-churn
invariant proved **byte-identical** across the `**Total**`→`Total` change on a
synthetic 11-language corpus, the exit criterion swept across all of `scripts/`
rather than `scripts/figures/`, and Makefile prerequisites derived mechanically.
It also checked the 0340 guard for the false-positive risk that mattered — its
`split_row()` consumes `\`+next-char as one unit, so escaped cells pass on a
corpus machine.

Three findings, all taken:

- **The docstring omitted a caller this PR adds** — the same file's round-2 fix
  had already corrected this once, and the merge introduced
  `export_corpus_flow.py` without naming it. Fixed in `a8c95cf0`, which also
  adds `tab_retrieval_protocol.md`, a caller that predates this branch and was
  never listed.
- **This body claimed 8 sweep targets; the function returns 10.** Corrected
  above. Round 3 reported verifying the 8 by running the function — that
  verification did not hold, and the count had moved under it.
- **Ticket 0371 undercounted its own blind spot.** `git check-ignore` over the
  discovery list gives **five** ignored tables, not two: the ticket was filed
  from the diff that surfaced the problem rather than from the problem. Widened
  in #1197 — five of ten generated targets are invisible to the 0340 guard on a
  corpus-free checkout, so it is blind to half its surface.

The panel also produced severity evidence for **0376**: the newly-escaped
`Code` column carries the raw corpus `language` value whenever it is exactly
two characters, and `@x` fits. Reproduced through the real emitter — the row
renders `<td>@X</td>` under `-f gfm`, which is what every escaping test since
0325 asserts through, and `<td><span class="citation">` under `-f markdown`,
which is what Quarto uses. This diff widens 0376's blast radius from
hand-edited labels to a malformed OpenAlex field.

Two `consider:` items noted rather than taken: the Total row's position-based
detection is already pinned by `test_total_row_keeps_its_emphasis` (bolding the
wrong row leaves `Total` without its `<strong>`), and the empty-cell `****`
shape is unreachable from both call sites.

### On the `make check` failures

The panel's full `make check` exited 1 with 7 failures, all
`test_corpus_acceptance.py` file-presence assertions. `.env` sets
`CLIMATE_FINANCE_DATA=data` (relative), so it resolves into the worktree, whose
`data/catalogs/` holds only `.dvc` pointers — the corpus lives in the primary
checkout. Confirmed environmental and unrelated to this diff; it is the
documented worktree/corpus trap, not a regression. The tiers this branch can
honestly gate — fast, adherence, integration — are all green on `dfd3210a`.


MinhHaDuong (commented):

**Verdict: comment.** No blockers. The fix is correct, the tests are genuine red→green against the *rendered* page, and the action-3 sweep claim checks out mechanically. Five findings below, none of which need to hold the merge; two are worth a small follow-up commit and one needs a ticket that the PR body itself promises.

Panel verdicts: correctness **approve**, consistency **approve**, scope **comment**, red team **approve**, doc propagation **comment**. No dissent between perspectives — every disagreement was about tag level, not about substance.

### Gates re-run on the branch head (bc82c26b)

- `make lint` — 212 passed, 10 skipped (the `8 scripts exceed 500 lines` warning is pre-existing, unrelated)
- `make check-fast` — 1218 passed, 11 skipped
- `tests/test_corpus_table_export.py tests/test_language_table.py` — 24 passed, 0 skipped (pandoc present)

### What the panel independently confirmed

- **The escape is genuinely load-bearing.** Red team reproduced the pre-fix behaviour by driving the old `_write_md_table` body through real `pandoc -f gfm`: a pipe in the source label splits the row and silently drops `%Refs`. Correctness reproduced the same thing separately. This is not a theoretical fix.
- **`markdown_text_cell` survives an adversarial payload set.** Embedded `\n`/`\r\n`, `---`, `:--`, leading `#`, `<td>`/`<br>`, trailing lone backslash, fullwidth-pipe U+FF5C, `a\|b`, `a||b` — every one rendered as a single well-formed row with the correct cell count, and every `\`/`|`/`` ` ``-bearing value round-tripped to the original text. `str.translate` is single-pass, so multi-character payloads in one cell do not cascade.
- **The 8-target sweep table is accurate.** `tests/_mk_discovery.py::generated_markdown_targets()` was run directly and returns exactly the 8 targets the PR body enumerates, item for item. No target is missing.
- **`export_citation_coverage.py`'s docstring claim is true.** Every table-row value in `render_table()` is coerced through `int(...)` or a numeric format specifier; so is the prose paragraph beneath it. The claim that "the venue emitters already record the same judgement" is also true — `scripts/figures/export_tab_venues.py:122` carries the identical shape of comment.
- **The `.gitignore` claim in the PR body is accurate, not rhetorical.** `git check-ignore -v` confirms `tab_corpus_sources.md` and `tab_languages.md` fall under the blanket `deliverables/_shared/tables/*` ignore (`.gitignore:47`) with no negation, while `tab_variables.md`, `codebook.md`, `tab_venues.md`, `tab_venues_fr.md` are force-tracked — and `.gitignore:57-60` does say the `.csv` is "grouped with `tab_corpus_sources.md`" while negating only the `.csv`.
- **Marker discipline is right.** `test_shipped_labels_are_untouched_by_the_escaper` is correctly unmarked: `tests/_tier_autoscan.py:26` lists `HEAVY_MODULES = ("dcor", "torch", "ot", "sentence_transformers", "matplotlib")`, and pandas is not among them.
- **The CSV sibling is untouched.** `save_csv` runs on `summary` before `_write_md_table` builds its locally-scoped escaped strings.

### Findings

**`verifiable:` The tracking ticket the PR body defers to does not exist.** (scope + doc propagation, both high confidence.) The body states the gitignored-handoff-artifact finding "gets its own ticket rather than riding this diff." `grep -rl 'tab_corpus_sources\|tab_languages' tickets/` returns only ticket 0367 itself and four closed tickets; the highest live ID is 0368. `git diff --stat main...HEAD -- tickets/` shows this branch adds 0358, 0367 and 0368 — none of them that ticket. Merging as-is drops the finding on the floor. The PR body's own paragraph is a ready-made ticket body.

**`verifiable:` `scripts/_markdown_table.py`'s docstring is now stale.** (doc propagation, high confidence.) It is the single documented home for the escaping convention, and it still says `markdown_text_cell` is "Used by the venue-table emitters" — this PR adds two callers (`export_corpus_table.py`, `export_language_table.py`) without updating it. One-line fix, and it belongs in this diff rather than a follow-up: the next reader of that docstring is the person deciding whether a new emitter needs the helper.

**`verifiable:` A source row with `n_refined == 0` publishes the literal string `nan`.** (correctness + red team, both high confidence, both independently traced.) `scripts/figures/export_corpus_table.py:104` — such a row is built with only 5 of the 8 dict keys, so `pd.DataFrame(rows)` fills the percentage columns with `NaN` rather than leaving them absent; `row.get(c, "")` therefore never reaches its default, and `markdown_text_cell(nan)` emits `"nan"`. Pre-existing (the old code produced the same string) and untested before and after, so this is not a regression from this diff — but the diff touches the interpolation line, which puts it in the "pre-existing but touched" tier. Worth its own small ticket rather than a fix here.

**`consider:` The two emitters now express "keep the total row bold" two different ways.** (correctness, consistency and red team all raised this independently.) `export_corpus_table.py:108-110` escapes then wraps — the order ticket 0367's action 1 specifies, and the order the new `test_total_row_stays_bold_around_an_escaped_value` pins. `export_language_table.py:131` instead bakes `"**Total**"` into the row data and passes the whole string through `markdown_text_cell`. That works only because the escaper's character set excludes `*`, which is an implementation detail of a shared helper rather than local code. It is documented and pinned by `test_total_row_keeps_its_emphasis`, so a widening of the character set goes red immediately — hence `consider:` and not a blocker. But two emitters fixed in the same commit for the same defect class reading differently is the kind of asymmetry that costs a future reader a detour.

**`consider:` The language table got no no-churn pin.** (red team, high confidence.) `tab_corpus_sources.md` has `test_shipped_labels_are_untouched_by_the_escaper` driven by the real `SOURCE_META` labels; `tab_languages.md`, fixed in the same commit, has no equivalent driven by real `LANGUAGE_NAMES` values. Red team verified all 39 current values are clean, so there is no live churn — the gap is that a future name addition or an escaper change has nothing to catch a regression on that side. Cheap to add, given the corpus-table version already exists to copy.

**`nofollow:` `markdown_text_cell` is not idempotent** on `\`/`|`/`` ` ``-bearing values (double application over-escapes). No call site applies it twice; recorded because the helper is now shared by six emitters and the property is worth knowing before a seventh wraps a value that has already been through it.

**`nofollow:` `"TOTAL" in str(row["Source"])`** is a substring match (`export_corpus_table.py:106`, blamed to `14793afd3`, untouched by this diff). A label containing "TOTAL" as a substring would be spuriously bolded. No `SOURCE_META` label does today.

**`nofollow:` `Makefile:301` passes `$(REFINED)` as `--input`** while `export_language_table.py`'s docstring and default (`ENRICHED_PATH`) describe "the full enriched corpus." Pre-existing (`fdd59283`, 2026-07-24), unrelated to this ticket.

**`nofollow:` The caption rationale comment exists on `export_corpus_table.py:79-81` but not on `export_language_table.py:156`**, whose caption has the same static-authored-markup property. Presentation asymmetry only.

### Recommendation

Merge after folding in the `_markdown_table.py` docstring line, and file the two tickets (the gitignored-handoff-artifact one the body already promises, and the `nan` cell). The `consider:` items are the author's call.

MinhHaDuong (commented):
## `/review-pr` panel — round 2 (5 perspectives)

Correctness · Consistency · Scope · Red team · Doc propagation, each in a fresh
context. **3 approve, 2 comment, no blockers, no dissent.**

Gates re-run independently in the PR worktree: `make check-fast` 1218 passed /
11 skipped · `make lint` 212 passed / 10 skipped (the pre-existing
`8 scripts exceed 500 lines` warning) · `ruff check` clean on all six changed
source files · the three directly affected suites
(`test_corpus_table_export.py`, `test_language_table.py`,
`test_markdown_table_shape.py`, integration marks included since pandoc 3.1.3 is
present) 57 passed / 4 skipped. A full `make check` was also launched; it is
still running against two concurrent full suites from parallel sessions, and
its extra coverage over the above is the slow tier — network, real corpus data,
heavy compute — none of which this diff touches.

### What the panel confirmed rather than assumed

- **Red-before-fix, independently.** Correctness reverted each emitter to
  `bc82c26b~1` in place and re-ran the new tests: the corpus table renders
  `**TOTAL</td><td>all sources**</td>`, the language table drops its last cell.
  Both restored, tree clean.
- **The 8-target sweep is real.** Scope ran `generated_markdown_targets()`
  itself and got exactly the PR body's eight, then grepped all of `scripts/`
  for `" | ".join` / `"|".join` / `f"| {` / `lines.append("|` independently of
  that discovery surface. Every hit outside `scripts/figures/` is a regex
  alternation, a DOI filter string, a CSV field or a log line. The exit
  criterion holds literally.
- **The `export_citation_coverage.py` no-escape claim is true.** Three
  reviewers checked `render_table` line by line: `label` is two `int()` calls
  around a literal en-dash, `total`/`with_data` are `int(...)`, `pct` only ever
  reaches a row through `{pct:.0f}%`. No path puts free text in a cell.
- **The Makefile prerequisite is complete, not a hardcoded pair of two.**
  Consistency enumerated every direct and transitive importer of
  `_markdown_table` (`_deposit_variables.py`, `export_tab_venues.py`,
  `export_core_venues_markdown.py`, and the two fixed here; transitively
  `export_codebook.py`, `export_variables_table.py`, `export_deposit.py`) and
  cross-checked each against its rule. Every script that actually *calls* the
  escaper already declares the prerequisite. `export_deposit.py` has no Make
  rule and does not call the escaper.
- **The renumber left no residue.** 0364 / 0367 / `t0367-escape-corpus-table`
  grep clean across code, tests, docstrings, Makefile, `.erg` and
  `deliverables/`. `erg check tickets/` passes on 355 tickets.
- **The two promised tickets exist.** 0371 (gitignored included tables) and
  0375 (`nan` from `row.get`) are open in #1197, and both bodies accurately
  describe what the round-1 panel found. The `.gitignore:57-61` stanza is
  exactly as the PR body characterises it.
- **The `.qmd` contract survives.** `{#tbl-quality}` and `{#tbl-languages}` are
  untouched; both `data-paper.qmd` and `corpus-report.qmd` still resolve their
  `{{< include >}}`; no surrounding prose quotes a cell value verbatim.

### Red team: attacks that failed

Round 1 had already failed to break the escaper with embedded newlines, `---`,
leading `#`, HTML, a trailing backslash, U+FF5C, `a\|b` and `a||b`. Round 2 went
past that list and also failed with, all pandoc-verified:

non-string types (NaN / None / `np.int64` / `np.float64` / list / dict /
`Timestamp` / `NaT`) · empty and whitespace-only cells · a full first data row
of `---` / `:-:` / `-` / `:--` / `--:` across every column · U+2028, U+2029,
NEL U+0085, VT `\x0b`, FF `\x0c`, RTL override U+202E, ZWJ, combining marks ·
backtick pairing across cells · `C:\Users\|pipe` backslash-pipe adjacency · a
raw tab.

The single-pass `str.translate` design is doing its job: because every backtick
is escaped, no cross-cell code span can form, and because the delimiter line is
a hardcoded literal immediately after the header, no data value can occupy that
position.

### Findings — all minor

**1. `verifiable:` `tests/test_language_table.py:196` — `require_pandoc()` in a
test that never renders.**

`test_shipped_language_names_are_untouched_by_the_escaper` calls
`require_pandoc()` at line 196, but its body only reads the emitted file back
and counts `\` and `*` (lines 210-220). The two `render_gfm` calls in this file
are at lines 147 and 176 and belong to the other two tests. Evidence:

```
$ grep -n "require_pandoc\|render_gfm" tests/test_language_table.py
16:from _gfm_render import cell_texts, render_gfm, require_pandoc, row_with
136:require_pandoc()          # test_pipe_bearing_language_code…
147:flat = render_gfm(…)
166:require_pandoc()          # test_total_row_keeps_its_emphasis
176:flat = render_gfm(…)
196:require_pandoc()          # ← this test never renders
```

The consequence is a no-churn regression pin that silently skips on any machine
without pandoc — the sibling pin in `test_corpus_table_export.py:201` has no
such gate. One-line fix: drop line 196.

**2. `verifiable:` `tests/_gfm_render.py:46` — `row_with` returns the first
match, which a future multi-row test can trip on.**

Red team built a two-row table pairing the real `SOURCE_META["oecd"]["label"]`
("OECD DAC key documents") with this ticket's own fixture
`PIPE_SOURCE = "OECD DAC | CRS key documents"` — a shared prefix — and
`row_with(rendered, "OECD DAC")` deterministically returned the decoy row.
Reproduced at synthesis through the real `_write_md_table` + pandoc:

```
decoy label: 'OECD DAC key documents'
row_with('OECD DAC') -> OECD DAC key documents     # the piped row is never reached
```

The
two shipped tests are not fooled today (each drives a one-row DataFrame), so
this is not a defect in #1191. But the helper is now on its third suite
(0325, 0339, 0370) and its docstring promises "the rendered `<tr>` carrying
`needle`" — singular. Suggested hardening, in this diff or a follow-up:
`assert len(matching) == 1` instead of `assert matching`.

**3. `consider:` `scripts/figures/export_language_table.py:158` — the Total row
is found by position, with no assertion tying index to label.**

`total_row = len(table) - 1`. Correctness traced `main()` (lines 94-138) and
confirmed it is structurally guaranteed today: `rows.sort(key=…)` runs *before*
`Other`, `Unclassified` and `Total` are appended, nothing sorts after, and the
`Total` append is unconditional — the empty-corpus case also holds, since
`Share (%)` is the hardcoded `"100.0"` rather than a division. So this is not a
current bug. It is an unpinned invariant in a file where the sibling emitter
finds its total row by reading `"TOTAL" in str(row["Source"])`. A one-line
`assert table.iloc[total_row]["Language"] == "Total"` would make a future
reordering fail loudly instead of bolding the wrong row.

**4. `consider:` `export_corpus_table.py:112`, `export_language_table.py:162` —
`f"**{cell}**"` on an empty cell renders as literal `****`.**

Verified with pandoc at synthesis — CommonMark needs non-empty emphasis
content:

```
empty       -> <td>****</td>
double-star -> <td>******</td>
```

Kept at `consider:` rather than `verifiable:` because it is unreachable from
either call site today: both wrap targets are a computed percentage or the
hardcoded label `Total`. Recorded, not worth guarding.

**5. `nofollow:` a zero-refined source ships a literal `nan`.**

Two reviewers reproduced it independently: `row.get(c, "")`'s default never
fires because pandas back-fills the missing percentage columns with real NaN
under a key that *is* present, so `markdown_text_cell(nan)` → `"nan"` and the
emitted table carries `| EmptySource | 10 | 0 | 0 | nan | nan | nan | nan |`.
Both confirmed the pre-fix `str(v)` produced the identical string — this PR
does not introduce or worsen it — and that ticket 0375 in #1197 describes it
accurately. Correctly deferred.

**6. `consider:` `scripts/figures/export_language_table.py:143-152` — the
rationale block is longer than the code it annotates.**

Nine comment lines above six lines of loop. Scope judged the escape-then-wrap
half load-bearing (it documents a defect class the review round actually found)
and the "hostage to a later widening of the character set" half as restating
the PR body for a reader who has it in `git log` anyway. Trimmable to two
sentences without loss. Author's call.

**7. `consider:` follow-up — `export_language_table.py` has no `_write_md_table`
sibling.**

The asymmetry behind finding 1: `export_corpus_table.py` factors its row loop
into `_write_md_table(summary, path)`, which is what lets its no-churn pin run
in-process and stay in the fast tier. `export_language_table.py` keeps the same
loop inline in `main()`, forcing its pin into a CLI subprocess and the
integration tier for an assertion that needs neither. Pre-existing structural
difference, not introduced here, but extracting the equivalent helper would
de-tier the test and drop the superfluous pandoc gate at the same time.

### Verdict

**comment** — no blockers, nothing that should hold the merge. Finding 1 is a
one-line deletion worth taking in this diff; findings 2, 3 and 7 are small
hardenings that would fit a follow-up ticket alongside 0371/0375; 4, 5 and 6
are recorded for the trail.

The fix itself held against everything two rounds of adversarial probing could
construct, the sweep is complete against an independently-run discovery surface
rather than against its own claim, and the no-churn invariant — the one that
decides whether regenerating these tables on a corpus machine is a no-op — is
pinned on both emitters.

MinhHaDuong (commented):
## `/review-pr` round 4 — five-agent panel

**2 approve (correctness, scope), 3 comment (consistency, red team, doc propagation). No blockers, no dissent.**

Panel briefed with rounds 1–3's findings as known-fixed, so nothing below is a repeat. Everything tagged `verifiable:` was reproduced, with the command shown.

---

### Independently reproduced (not taken on the PR body's word)

- **Red-before-fix, both emitters.** Reconstructed the old code from `git show origin/main:` and rendered through the real `tests/_gfm_render.py` oracle. Old `_write_md_table` on `"OECD DAC | CRS key documents"` → `['OECD DAC', 'CRS key documents', '1,200', '900', '300', '12%', '88%', '77%']` — eight cells, `%Refs` (`66%`) silently gone. Old `export_language_table.py` on code `"a|"` → `A | ␣ | a | ␣`, both counts vanished.
- **No churn on shipped output.** Built a synthetic 11-language / ~4,640-row `enriched_works.csv` with no escapable characters and ran the `origin/main` and post-PR emitters side by side: **byte-identical**, including the `**Total**` → `Total`+wrap-after-escape change.
- **`row_with` hardening is safe.** All 6 call sites across `test_corpus_flow.py`, `test_corpus_table_export.py`, `test_language_table.py`, `test_rendered_table_fidelity.py`, `test_rendered_venue_table_fidelity.py` use single-matching-row fixtures. Suites run: 17 passed.
- **The 0340 shape guard will not false-positive on escaped cells.** `split_row()` (`tests/test_markdown_table_shape.py:51-84`) hand-scans and consumes `\`+next-char as one unit rather than counting raw `|`, and `test_escaped_free_text_keeps_the_row_whole` pins exactly that. File run: 36 passed, 5 skipped. So on a corpus machine the newly-escaped artifacts pass the guard.
- **Makefile prerequisites complete.** Derived mechanically (grep imports → grep rules): all 8 targets whose script imports `_markdown_table` declare `scripts/_markdown_table.py`, and `tab_citation_coverage.md` correctly omits it (docstring-only change, no import).
- **Exit criterion holds.** Swept all of `scripts/`, not just `scripts/figures/`. Every emitter interpolating corpus-derived free text routes through the helper; header rows are hardcoded literals everywhere; no pipe table is built outside `scripts/figures/`.

---

### Findings

**1. `verifiable:` — the docstring caller list omits the third caller this same PR adds.** *(doc propagation / correctness, high)*

`scripts/_markdown_table.py:23-24` now reads "Used by the venue-table emitters, the corpus-sources table and the language table (ticket 0370)." But `export_corpus_flow.py` is also a caller, and it became one *in this diff*:

```
$ git show origin/main:scripts/figures/export_corpus_flow.py | grep -c markdown_text_cell
0
$ grep -c markdown_text_cell scripts/figures/export_corpus_flow.py
1
```

This is round 2's finding recurring for the third caller of the same PR — the docstring is the one documented home of the escaping convention, and it is again one caller short. One-line fix. (`export_retrieval_protocol.py` is also unnamed, but was already a caller on `origin/main` — pre-existing, not this diff's.)

**2. `verifiable:` — the sweep table says 8 targets; the discovery surface returns 10.** *(doc propagation, high)*

```
$ python -c "from test_markdown_table_shape import GENERATED; print(len(GENERATED))"
10
```

Missing from the PR body's table: `tab_retrieval_protocol.md` and `tab_corpus_flow.md` — the latter being one this diff itself fixes. Round 3's panel reported checking the "8" by running `generated_markdown_targets()` itself; that verification did not hold. Body-only, no code impact, but it is the artifact future sweep bookkeeping will read.

**3. `consider:` — ticket 0371 (open in #1197) undercounts the gitignore blind spot.** *(doc propagation, high)*

```
$ git check-ignore -q deliverables/_shared/tables/<f>
tab_corpus_sources.md     IGNORED
tab_languages.md          IGNORED
tab_corpus_flow.md        IGNORED
tab_citation_coverage.md  IGNORED
tab_venues.md             tracked
tab_variables.md          tracked
```

0371 is written for two files. `tab_corpus_flow.md` meets its criteria identically — `{{< include >}}`d by `data-paper.qmd:106`, ignored, and escaped by this very diff. As written, 0371 would close the blind spot for two of the four and leave `tab_corpus_flow.md` in it. #1197 is still open; fixable there. (The misleading `.gitignore:62-66` comment the body flags is real — confirmed.)

**4. `consider:` — this diff raises 0376's severity from hypothetical to one malformed field away.** *(red team, high)*

`normalize_lang` returns any exactly-2-character value unchanged, so the `Code` column carries the raw corpus `language` value verbatim — genuinely uncurated, unlike every other column this diff touches. `@` fits in two characters. Reproduced through the real emitter, not a hand-built table:

```
$ printf 'language\n'; for i in $(seq 250); do printf '"@x"\n'; done   # -> enriched.csv
$ python scripts/figures/export_language_table.py --input enriched.csv --output tab_languages.md
| @X | @x | 250 | 100.0 |

$ pandoc -f gfm      -t html tab_languages.md   # the test oracle
<td>@X</td><td>@x</td>
$ pandoc -f markdown -t html tab_languages.md   # Quarto's actual reader
<td><span class="citation" data-cites="X">@X</span></td>
```

Not a defect in this diff — 0376 is already filed and ordered oracle-first. Recording it as evidence for where 0376 sits in the queue: the newly-escaped `Code` column is a likelier carrier than the hand-edited labels 0376 was filed against.

**5. `consider:` — two idiom divergences survived round 3's harmonisation.** *(consistency, medium)*

- *Escape-every-cell* (`export_corpus_table.py`, `export_language_table.py`) vs *escape-only-the-column-we-think-is-unsafe* (`export_corpus_flow.py`, `export_citation_coverage.py`). The diff's own comment argues "the escaper exists so no emitter has to know" which column carries a `|` today — and then two of the four emitters it touches make exactly that claim. Both claims are correct *today* (verified: every other cell in both is `int`-formatted). The question is whether the convention is "escape everything" or "escape what you must and document why", since the PR argues the first and ships the second.
- *Total-row identity*: substring (`"TOTAL" in str(row["Source"])`, `export_corpus_table.py:144`) vs position (`len(table) - 1`, `export_language_table.py`). The red team reproduced the substring form spuriously bolding a plausible non-aggregate label:

  ```
  Source = "Global TOTAL Assessment Registry"  ->  | **Global TOTAL Assessment Registry** | **10** | ...
  ```

  Pre-existing — the line is byte-identical on `origin/main` — and unreachable with today's eight `SOURCE_META` labels. Flagged because round 3 harmonised these same two files on escape/wrap ordering and left this second divergence between them.

**6. `consider:` — `scripts/figures/export_citation_coverage.py:40-42`**, two em-dashes in one docstring paragraph. `prose/_all.md` allows one.

**7. `nofollow:` — `tests/_gfm_render.py`'s `_CELL` matches `<td>` only**, so a needle that hit a header row would make `cell_texts` return `[]` rather than error. No current call site can reach it (all header content is a hardcoded literal); noting it beside the `row_with` hardening the same file just received.

---

### Attacks that failed (recorded so round 5 doesn't repeat them)

Empty corpus (0 works) · every language below `MIN_COUNT` · a corpus language literally named `"Other"` · reaching the `****` empty-Total-label shape (`"Total"` is a hardcoded literal, no lookup can miss) · NFKC pipe look-alikes beyond U+FF5C (÷, ∣, ¦, U+FE31, U+FFE8 — none normalise to ASCII `|`, and pandoc does not NFKC input) · tab inside a cell · header-row interpolation (literals in all four emitters) · reordering the Total row via sort (`rows.sort()` runs before Other/Unclassified/Total are appended, so it is structurally last).

### Gates

- `make check-fast` — exit 0, zero `FAILED`.
- `make lint` — 234 passed, 11 skipped.
- Touched suites (`test_corpus_table_export.py`, `test_language_table.py`, `test_corpus_flow.py`) — 48 passed, 4 skipped, 1 pre-existing failure: `test_generated_flow_closes_to_the_refined_corpus` (`@pytest.mark.slow`, from ticket 0327, unmodified here, `FileNotFoundError: refined_works not found`).
- Full `make check` not re-run: the `test_corpus_acceptance.py` failures the body documents are the known worktree/corpus-path trap, not this diff.

### Verdict

**comment** — no blockers. Finding 1 is a one-line docstring fix and worth taking before merge, since it is the same defect round 2 already corrected once in this file. Findings 2 and 3 are bookkeeping and can be taken in #1197. The rest are recorded, not requested.

PR #1192, opened 2026-07-27
fix(0352): archive script-list guard reads the whole SCRIPTS array

MinhHaDuong:
**Ticket:** tickets/0352-archive-script-list-guard-truncates-sile.erg

## What was wrong

`tests/test_archive_script_paths.py` promises, in its own docstring, that it
resolves *every* script path the archive tooling will `cp` or invoke. It parsed
the array with:

```python
re.search(r"SCRIPTS=\((.*?)\)", content, re.DOTALL)
```

`(.*?)` is non-greedy, so the capture ended at the **first** `)` after
`SCRIPTS=(` — including one inside an ordinary explanatory comment. Every path
below that point went unchecked while the guard stayed green. A guard that
silently inspects a prefix is worse than none, because the docstring's promise
is what the next mover relies on.

Ticket 0339 hit this, could not fix it inside a Markdown-escaping PR, and left
a "keep this array free of parentheses" warning in the build script instead — a
tripwire depending on the next editor reading a comment, which is the exact
failure mode the guard exists to remove.

## What changed

1. **The parse is a pure function** taking file *content*
   (`_parse_scripts_array`). The old form both read and parsed `BUILD_SCRIPT`,
   so it could only ever be exercised against the real array — the ticket's
   specified red test could not be written against it. Committed separately and
   behaviour-preserving, so the defect stayed reproducible for the test.
2. **Line-anchored block delimiter.** The array is shell source with one entry
   per line and its terminator alone on a line, so `^SCRIPTS=\([ \t]*$ … ^\)[ \t]*$`
   is sufficient. The ticket offered bracket counting as the alternative; the
   line anchor is simpler and fails *loudly* (no match at all) if the array's
   shape ever changes, where a bracket counter would keep quietly returning
   something.
3. **Entry invariants.** Comments are stripped per line, so a path named only in
   prose cannot masquerade as an entry; every non-comment line must yield
   exactly one `scripts/*.py` path; and exactly one `SCRIPTS` declaration may
   exist, counted over a pattern that also matches the `SCRIPTS+=(` append form.
4. **The stale paren warning is gone**, and its replacement comment uses
   ordinary parentheses deliberately.

## Evidence

The rewritten comment makes the real array a **live regression witness**, not
just a synthetic fixture. Against the current `build_analysis_archive.sh`:

| parser | paths read |
|---|---|
| old `SCRIPTS=\((.*?)\)` | **13** |
| new line-anchored | **20** |
| independent line scan | **20** entry lines |

The 7 the old regex drops are led by `scripts/_venue_naming.py` and
`scripts/_markdown_table.py` — precisely the helpers 0339 surfaced. All 20
resolve on disk.

Red/green was genuine: the three fixture tests failed on the extracted-but-
unfixed parser and pass after. The length ratchet re-derives the block by a
plain line scan rather than reusing `SCRIPTS_ARRAY_RE`, because a ratchet
measured with the regex under test shrinks along with it and reports nothing.

## Review rounds

Two `/review-pr` panels ran. Both found further instances of *this ticket's own
defect class* — a guard that checks a subset and stays green — and all of them
in code this PR introduced, so they were fixed here rather than deferred:

- **Token boundary.** `SCRIPT_PATH_RE` had no boundaries, so `../scripts/utils.py`,
  `old_scripts/utils.py`, `scripts/utils.pyc` and `scripts/utils.py.bak` each
  yielded `scripts/utils.py`. An entry the archive's `cp` cannot use would have
  resolved against the real file and passed.
- **Append form.** The first "exactly one array" check counted
  `SCRIPTS_ARRAY_RE` matches, which are anchored to literal `^SCRIPTS=\(` — so
  `SCRIPTS+=(` produced no second match and the check never fired. Now counted
  over declarations.
- **Makefile sibling.** `_makefile_script_paths` stripped whole-line comments
  only, leaving a trailing `# see scripts/x.py` readable as a prerequisite.
- **Ratchet false alarm.** The line scan matched delimiters exactly while the
  regex tolerates trailing whitespace, so a padded `SCRIPTS=(` raised
  `ValueError` on a file the parser reads fine.

Each is pinned by a test. Real parse counts are unchanged throughout — 20 array
paths, 20 line-scan entry lines, 20 Makefile references.

## Known gap, deliberately not fixed here

The guard validates the `SCRIPTS` *declaration*; it never checks that the `cp`
loop consumes it unmodified. A `cp` line outside the array, or a slice such as
`CORE_SCRIPTS=("${SCRIPTS[@]:0:6}")`, still passes. This is **pre-existing** —
the old parser did not check the declaration-to-loop binding either — and out of
0352's scope, whose exit criteria are about reading the array completely. It
warrants its own ticket; filing it belongs on a ticket-filing PR, not this
feature branch.

## Gate

- `make lint` — 190 passed, 6 skipped. Test stays in the `adherence` tier.
- `make check` — 1878 passed. The 7 failures are all
  `test_corpus_acceptance.py` file-existence checks: this worktree holds only
  `.dvc` pointers, not materialized corpus data. Verified environmental, not
  code — pointed `CLIMATE_FINANCE_DATA` at the real corpus and the same tests
  ran **47 passed, 12 skipped**. This PR touches no data path.
- `/verify-adherence` — PASS, no mechanical or semantic findings. It
  independently reproduced the 13-vs-20 result rather than trusting the ticket.
- `build/build_analysis_archive.sh` change is comment-only; the archive ships
  exactly what it shipped before.


MinhHaDuong (commented):
## `/review-pr` — four-perspective panel

Correctness · Consistency+DocPropagation · Scope · Red team. Each agent read ticket 0352's exit criteria and the diff independently, and probed the real functions rather than trusting the PR body.

**Panel verdict: comment (approve with one follow-up).** Three agents returned `approve`; the red team returned `request-changes`. That dissent is preserved in full below — the author decides.

---

### Verified independently

| Claim | Agent | Result |
|---|---|---|
| Old regex reads 13 paths, new reads 20, against the real `build_analysis_archive.sh` | Correctness | Reproduced. Old parse stops after `scripts/figures/export_tab_venues.py`; the 7 dropped are led by `_venue_naming.py` and `_markdown_table.py`. |
| `FIXTURE_PAREN_COMMENT` genuinely distinguishes old from new | Correctness | Reproduced. Old parser reads only `['scripts/utils.py']` from the fixture, so the existence guard over it reports zero missing — a silent pass. New parser surfaces the bad path. |
| The refactor commit is behaviour-preserving, so the defect stayed reproducible for the red test | Scope | Confirmed: `git show caaf5fd7:tests/test_archive_script_paths.py` still carries the unmodified non-greedy `re.search`. |
| `build_analysis_archive.sh` change is comment-only | Scope | Confirmed: array entries byte-identical, only the warning comment's text changed. |
| No other file parses or duplicates the SCRIPTS array | Consistency | Confirmed by repo-wide grep for `SCRIPTS=`, `parenthes`, `non-greedy`, `0339`, `0352`. Only remaining hits are inside this test file and ticket 0352's own body. Ticket 0339's closed file never carried the warning — the tripwire lived only as a code comment, so there is no stale leftover. |

Gate re-run by the panel, not taken from the PR body:
- `make lint` — **185 passed, 6 skipped**.
- `make check` — **1873 passed, 7 failed, 78 skipped**. All 7 are `test_corpus_acceptance.py::TestFileExistence` / `TestCorpusAlign` misses on `data/catalogs/*`: this worktree holds `.dvc` pointers, not materialized corpus. Environmental, unrelated to this diff.
- `tests/test_archive_script_paths.py` alone — 8/8 pass.

---

### Dissent — red team, `request-changes`

Quoted verbatim, high confidence:

> **`SCRIPT_PATH_RE` has no left boundary: silently extracts the wrong path from a prefixed line.** `SCRIPT_PATH_RE = re.compile(r"scripts/(?:[\w-]+/)*[\w.-]+\.py")` has no anchor/lookbehind before `scripts/`, so it matches `scripts/...` as a *substring* of a longer token […] The array literally says `../scripts/utils.py` […] but the guard validates and reports `scripts/utils.py` — a real file that happens to already be in the repo. The per-entry "exactly one path" invariant does not save this: the wrong path still counts as exactly one. […] This directly falsifies the PR's own docstring claim ("any mismatch between them fails loudly here") for this specific failure mode, and it's exactly the "silently checks a subset and stays green" class the ticket exists to close — just relocated from block-boundary to token-boundary.

Reproduced by the synthesizer:

```
'../scripts/utils.py'     -> ['scripts/utils.py']
'old_scripts/utils.py'    -> ['scripts/utils.py']
'$DIR/scripts/utils.py'   -> ['scripts/utils.py']
```

**Synthesis position: `verifiable:` comment + follow-up ticket, not a blocker.** `SCRIPT_PATH_RE` is unchanged by this diff (`git diff origin/main..HEAD` touches only its two call sites); it is pre-existing on `main`, and no current entry in the array triggers it. Under the project's code-quality escalation table that is "pre-existing but touched → comment + new ticket", not `request-changes`. Blocking a strict improvement on a latent pre-existing gap would be the wrong trade. The red team is right that the fix is cheap — a `(?<![\w/])` lookbehind — and right that it belongs to the same bug class; it is a good candidate for the author to take inside this PR if he prefers to close the class in one move. **Recommend filing it as a follow-up ticket if not.**

---

### Non-blocking findings

`verifiable:` **First-`SCRIPTS=(`-wins is a shared blind spot of both parsers** (red team, medium). `re.search` and `lines.index("SCRIPTS=(")` both take the first occurrence, so an earlier heredoc-generated array would defeat the parser and the ratchet identically:

```
heredoc decoy block first ->  PARSE: ['scripts/decoy_from_heredoc.py']
                              LINESCAN: ['    scripts/decoy_from_heredoc.py']
```

Not exploitable against the current file. The "independent implementation" decorrelates against wrong-*length*-of-the-same-block, not against wrong-*block*. Worth one line in the module docstring stating that assumption.

`verifiable:` **The sibling `_makefile_script_paths` carries the pre-fix version of this bug class** (red team, medium). It strips only whole-line `#` comments, not trailing inline ones, so `foo.csv: scripts/real.py  # superseded scripts/deprecated.py` yields three prerequisites. Dormant against the real `Makefile.analysis-manuscript`. Failure mode is spurious-loud, not silent-pass, so lower severity than Finding 1 — but it belongs in the same follow-up ticket.

`consider:` **`_entry_lines_by_line_scan` is a second parser kept permanently, and its independence is partial** (scope, medium). It duplicates the same comment-stripping convention (`line.split("#", 1)[0].strip()`) as the parser, so it guards the regex-truncation class but not a shared misreading of the comment convention. The PR body already flags this for `/simplify`; the panel's position is that `/simplify` should act on it rather than let the duplication become permanent by default.

`consider:` **Whitespace divergence between the two implementations surfaces as `ValueError`, not `AssertionError`** (red team + correctness, high). Trailing space on `SCRIPTS=(` or `)` makes the regex match and `lines.index` raise. Loud either way, so no masking risk — only a confusing error type for whoever hits it. The PR body's "Note for review" already names this divergence honestly.

`consider:` **The array must stay a column-0 top-level declaration** (red team, medium). Wrapping it in a function body indents `SCRIPTS=(`, the anchored regex stops matching, and the guard fails cleanly with "must declare a SCRIPTS=( ... ) array". Acceptable behaviour, but it is a real constraint on future refactors that is not stated near the regex.

`consider:` **Test-tier placement of the new fixture tests** (consistency, medium). The three fixture tests are pure string parsing with no file I/O; the file's blanket `pytestmark = pytest.mark.adherence` keeps them out of `make check-fast`. Defensible — they are contract guards, which is what the adherence tier is for — but they are the cheapest and most defect-sensitive tests in the file and would earn their place in the fast loop.

`nofollow:` **Fixture brittleness** (red team, low). `test_existence_guard_catches_a_missing_path_below_such_a_comment` depends on `scripts/no_such_helper_below_a_paren_comment.py` not existing on disk. Negligible.

`nofollow:` **`FIXTURE_PATH_IN_COMMENT` and its test exceed the ticket's stated Test section** (scope, high). A natural corollary of the comment-stripping the fix needed anyway, one fixture and one test, no packaging risk. Deliberately not flagged.

`nofollow:` **No standalone red commit in history** (scope, low). `ed484f27` bundles the fix, both invariants, the second parser, and all five tests; the ticket log's "3 failures at commit 1" was a manual verification step. The commit is still single-purpose, and the refactor/fix split — which is the one that mattered here — was done correctly.

---

### What the panel confirmed is handled correctly

Probed and found to fail *loudly* rather than silently, which is the property this ticket exists to protect: a `# )` line inside the array body; a misindented real terminator; two paths on one line; line continuations; `${DIR#…}` parameter expansion; quoted entries. Reordering or adding an entry is handled correctly and order-independently.

The ticket's three Actions are all delivered, the exit criteria carry concrete evidence rather than bare ticks, and the `.erg` log entry is append-only and correctly formatted. `make lint` keeps the test in the `adherence` tier as the ticket's invariant requires.

MinhHaDuong (commented):

**Panel verdict: comment.** Correctness, Consistency and Doc-propagation approve; Red team and Scope requested changes. Dissent is preserved verbatim below — the author decides. Nothing here is a correctness regression: the ticket's own defect is genuinely closed, verified independently three times.

`make check-fast` — exit 0, no failures. `make lint` — 187 passed, 6 skipped (reviewer-run). `pytest tests/test_archive_script_paths.py` — 10 passed.

### Reproduced, not taken on trust

| parser | paths against the real `build_analysis_archive.sh` |
|---|---|
| old `SCRIPTS=\((.*?)\)` | **13** |
| new line-anchored | **20** |
| independent line scan | **20** entry lines |

The PR body's table, the ticket's 13-vs-20 claim, the 7 dropped paths, and every `SCRIPT_PATH_RE` boundary example (`../scripts/utils.py`, `old_scripts/utils.py`, `scripts/utils.pyc`) were each re-derived by two reviewers running the old and new regexes rather than reading the claim. All match. `_venue_naming.py`'s "missing until 0339" claim checks out against `git log -S` (commit `9b86989b`). Every ticked ticket checkbox has real backing in the diff.

---

### Findings

**1. `verifiable:` — `SCRIPTS+=( ... )` is invisible to the new "exactly one array" assertion.** [red team, confirmed by synthesis]

The comment at `tests/test_archive_script_paths.py` on `len(blocks) == 1` says a second array "would be parsed by neither this nor the line scan — a silent blind spot of the 0352 kind", and asserts against it. `SCRIPTS_ARRAY_RE` is anchored to literal `^SCRIPTS=\(`, so the append form does not match at all and the assertion never fires:

```
attack3 SCRIPTS+= : paths parsed = 20 | ghost seen = False
attack3 line-scan entries = 20
```
(injected `SCRIPTS+=(\n    scripts/figures/plot_does_not_exist.py\n)` before the `cp` loop; full suite: 10 passed, 0 failed)

`+=` is the idiomatic bash way to extend an array — the natural edit for a contributor adding one script without touching the 20-line literal. Cheapest fix: have the array regex *detect* `SCRIPTS+=(` and fail loudly, which is exactly what the existing assertion's comment already promises.

**2. `verifiable:` — the guard validates the declaration, never that the declaration drives the `cp`.** [red team]

Two attacks, each proven end-to-end against a scratch copy of the repo with the real pytest suite (**10 passed, 0 failed** in both):

- a `cp --parents scripts/figures/plot_totally_missing_fig.py "$TMP/"` line added *outside* the array — nothing scans the rest of `build_analysis_archive.sh` for `scripts/*.py` the way `_makefile_script_paths()` scans the whole Makefile template;
- `CORE_SCRIPTS=("${SCRIPTS[@]:0:6}")` substituted into the loop at line 82, leaving all 20 real entries declared and resolving — the archive ships 6 of 20, `test_build_and_makefile_agree_on_entry_points` included stays green, because it too reads the *static* array.

This gap is **pre-existing, not introduced here** — the old parser never checked the array-to-loop binding either. Per the project's escalation table (pre-existing but touched) this is a comment plus a follow-up ticket, not a blocker on 0352. It matters more after this PR than before only because the module docstring now makes "every script path the archive tooling will actually `cp`/invoke" a strengthened promise, and the promise still outruns the coverage by one level: the guard verifies an array, not a build.

**3. `verifiable:` — whitespace divergence turns the ratchet into a traceback, not a verdict.** [correctness + consistency + red team, three-way convergence; already self-flagged in the PR body]

`SCRIPTS_ARRAY_RE` tolerates `[ \t]*` on the delimiter lines; `_entry_lines_by_line_scan` uses exact `lines.index("SCRIPTS=(")`. One trailing space after `SCRIPTS=(`:

```
ws: regex parses 20 paths
ws: line scan ValueError: 'SCRIPTS=(' is not in list
```

Loud, so not a bypass — the PR body is right that it never returns a silently wrong count. But an editor's format-on-save produces an uncaught `ValueError` traceback pointing at the guard rather than at the whitespace, and the tempting "fix" is to loosen the line scan, which is the invariant-inversion this test exists to prevent. Matching the regex's tolerance in the line scan (or catching the `ValueError` into the intended assertion message) costs one line.

**4. `verifiable:` — `SCRIPT_PATH_RE`'s lookahead is asymmetric with its lookbehind.** [correctness]

```
.bak: ['scripts/utils.py']
pyc : []
```

`(?!\w)` rejects `scripts/utils.pyc` but not `scripts/utils.py.bak`, while the lookbehind `(?<![\w/.-])` does exclude `.`. Dormant — no such entry exists — but it is the same near-miss class commit `ceecff3c` set out to close, left half-closed.

**5. `consider:` — commit `ceecff3c` has no ticket trace.** [scope, dissenting]

Verbatim, so the disagreement is on the record: *"`ceecff3c` fixes a genuinely different defect class (token-level over-match) than the one the ticket describes (block-level non-greedy truncation). The ticket's Actions/Test/Verification/Exit-criteria were authored and checked off in `375ebe2d` before this commit landed, and `ceecff3c` added no new log entry, Action item, or exit criterion. […] Split `ceecff3c` into three single-purpose commits."*

Synthesis note, against that: the commit message states its own provenance — *"Review of PR #1192 found the ticket's defect class surviving one level down"* — which is the harness's normal in-flight review-fix path, not undisclosed creep, and the three sub-fixes share one cause (a guard checking a silent subset). The residue that stands is traceability, not scope: a one-line ticket log entry recording the review round would close it. Not blocking.

**6. `nofollow:` — module docstring rewrite (`6ccb8e99`) and the `build_analysis_archive.sh` comment reword (`83dab77e`).** Both flagged as unmapped-to-Actions by Scope; both are the minimal edit consistent with removing the stale paren warning and with making the docstring's promise true. Recorded, no action.

### Where the panel agreed

The red/green was genuine — running the *old* parser against `FIXTURE_PAREN_COMMENT` yields `['scripts/utils.py']` only, so the fixture distinguishes a fixed parser from a lucky one exactly as the ticket's Test section demanded. The two-independent-parsers design and the length ratchet are earned by the ticket's own reasoning, not gold-plating. `findall` with one named group returns strings, so `len(blocks) == 1` counts correctly. The Makefile inline-comment strip is a verified no-op on the real template (20/20 both ways) — strictly stronger, not a behaviour change. No stale docs, no other consumer of `SCRIPT_PATH_RE`, marker tier and naming match the sibling adherence tests.

### Suggested disposition

Findings 1, 3 and 4 are one-line changes inside this PR's own new code. Finding 2 is the pre-existing gap and wants its own ticket — *the archive guard verifies the SCRIPTS declaration, not the loop that consumes it* — which is a strictly larger claim than 0352 made and deserves to be scoped on its own.

🤖 five-perspective panel: correctness · consistency · scope · red team · doc propagation

MinhHaDuong (commented):
## /gaze code review — PR 1192 (Agent C, standard tier: correctness + consistency)

**Verdict: approve**

Verified directly against `origin/main...HEAD` and the live worktree, `make check-fast` green (1242 passed, 11 skipped, 0 failed), `ruff check tests/test_archive_script_paths.py` clean, `tickets/erg check tickets/` PASS (355 tickets), all 13 tests in the changed file pass in isolation.

### Correctness

- **13→20 count is exactly reproducible.** Ran both the old non-greedy `SCRIPTS=\((.*?)\)` and the new `_parse_scripts_array` against the *current* `build/build_analysis_archive.sh` content: old regex's capture stops mid-comment at 13 entries (drops `_venue_naming.py`, `_markdown_table.py`, and everything below), new parser reads all 20. Matches the ticket's verification evidence exactly.
- **`SCRIPT_PATH_RE` token boundaries hold up.** Traced the lookbehind/lookahead by hand and confirmed empirically: `../scripts/utils.py` fails the lookbehind (preceding `/`), `old_scripts/utils.py` fails it too (preceding `_` is `\w`), `scripts/utils.pyc` and `scripts/utils.py.bak` both fail the lookahead. All four reduce `found` to length ≠ 1, so `_parse_scripts_array` raises — matches `test_a_path_is_not_matched_inside_a_longer_path` / `test_a_py_path_is_not_matched_inside_a_longer_extension`.
- **`SCRIPTS_DECL_RE` correctly counts both declaration forms** and doesn't false-positive on the real file: grepped `SCRIPTS` in `build_analysis_archive.sh` — only the one `SCRIPTS=(` declaration and one `${SCRIPTS[@]}` reference in the `cp` loop, and the reference doesn't match `^SCRIPTS[ \t]*\+?=` (doesn't start the line with `SCRIPTS`).
- **Makefile-sibling fix is real but not yet load-bearing.** `_makefile_script_paths` now strips inline (not just whole-line) comments via the shared `_strip_inline_comment`. Grepped `Makefile.analysis-manuscript` for `#.*scripts/` — the two hits are both whole-line comments the old check already stripped, so the new parser produces the same 20-path result today (verified). The fix is defensive hardening against a class of bug, not a fix for a currently-tripped one — worth knowing when reading "20 old / 20 new" for this one file, since the ticket's headline number (13→20) is about the array parser, not this sibling.
- **Length ratchet is genuinely independent.** `_entry_lines_by_line_scan` uses `.index("SCRIPTS=(")` / `.index(")")` string scanning, not `SCRIPTS_ARRAY_RE` or `SCRIPT_PATH_RE` — confirmed by reading it; a regression in either production regex cannot silently satisfy its own ratchet.
- **`cp` loop still consumes `"${SCRIPTS[@]}"` unmodified** — confirmed by reading the loop at line 82. The PR's disclaimed gap (a slice like `${SCRIPTS[@]:0:6}` would still pass the guard) is accurately described as pre-existing and out of scope; nothing in this diff touches the loop.

Edge cases checked, none are live defects:
- Zero-entry array (`SCRIPTS=(\n)\n`): `blocks[0]` is empty, loop doesn't run, `paths == []`, hits the existing `assert paths, "...holds no scripts/*.py paths"` — loud failure, not silent pass.
- Indented terminator (`    )` instead of `)`): `^\)[ \t]*$` requires the line to *start* with `)`, so an indented closer isn't recognized. Simulated both sub-cases: (a) no later bare-`)` line in the file → `assert blocks` fails loudly with a clear message; (b) a later unrelated bare-`)` line exists → the non-greedy body over-captures through it, and the per-line "exactly one path" invariant then fails loudly on the swallowed non-path lines. Either way it's a loud, if not maximally legible, failure — not a silent truncation. `consider:` — the error message in sub-case (b) would name the wrong line as the culprit; not worth a fix given the fixture-asserted contract already documents the array must have its terminator alone on a line.

### Consistency

- `pytestmark = pytest.mark.adherence` is already present (unchanged by this PR) — correctly tiered per `.claude/rules/coding-python.md`'s testing table (hygiene/guard test that reads files). Matches the ticket's own Invariants section.
- No `Optional`/`List[`/`Dict[`/`Tuple[` introduced; style matches `coding-python.md`.
- `tickets/0352-archive-script-list-guard-truncates-sile.erg`: no `Status:`/`Closed:` header, log entries only appended (diff shows one new `2026-07-27T14:40Z` line, nothing rewritten), exit criteria and verification checkboxes ticked with concrete evidence (counts, fixture name, test failure count at commit 1). Complies with `tickets/AGENTS.md` and the %erg 0.1 format. `erg check tickets/` passes at 355 tickets.
- `consider:` `_strip_inline_comment`'s `line.split("#", 1)[0].strip()` idiom is now the 4th independent inline reimplementation in the test suite (also in `test_cited_works_available.py`, `test_evict_analysis_intermediates.py`, `test_import_path_model.py`). Pre-existing pattern this PR follows rather than introduces net-new duplication of; no shared test-helpers module exists yet to extract it into. Not worth blocking on — flagging in case a future pass wants to consolidate.
- `nofollow:` the stale "keep this array free of parentheses" comment removal (chore commit `83dab77e`) and the rewritten comment in the same block read cleanly and match what `git diff` shows; no residual reference to the old warning anywhere else in the repo (not re-grepped exhaustively beyond `build/` and `tests/`, low risk given the ticket's own Action 3 named exactly this comment).

### Dissent / open questions
None — correctness and consistency perspectives concur; no disagreement to preserve.

---
*Agent C, /gaze tier=full, standard risk (correctness + consistency pair). No subagent-spawning tool was available in this environment; both perspectives were run directly by this agent against the live diff and worktree rather than via forked sub-agents.*

PR #1193, opened 2026-07-27
test(0346): generalize the facade-patch guard and finish the sweep

MinhHaDuong:
**Ticket-ref:** tickets/0346-run-report-tests-patch-the-wrong-module-a.erg

Deliberately `Ticket-ref:`, not `Ticket:`. All three exit criteria are discharged, but Action 2 — deleting the eight leaked fixture files and re-running `dvc add` — is still open and can only be done once this lands and sibling worktrees rebase. Closing 0346 on merge would leave that recipe as prose in a closed ticket with nothing tracking it. The ticket stays open until the deletion runs.

PR #1174 stopped the two leaks. This closes the residue: the second half of exit criterion 1 (`monkeypatch`, not hand-rolled save/restore) and exit criterion 3 (the sweep, recorded in the ticket).

## The guard was keyed on the constant that bit us

`CATALOGS_DIR` was a name in the guard, not a mechanism. It is now derived from the source: an AST walk over `scripts/` collects every constant a function re-imports from `pipeline_loaders` inside its own body — that function-local import re-reads the definition site on every call, which is exactly why rebinding the `utils` re-export redirects nothing.

That set is `{CATALOGS_DIR, POOL_DIR}` today. **`POOL_DIR` is the same defect one call site short of firing**: `pool_path()` re-imports it and then calls `os.makedirs`, so a test patching the facade would create directories inside DVC-tracked `data/pool/`. No test exercises the pool helpers yet — which is precisely why a name-keyed guard would miss the first one that does.

## Scoped to the facade, not to any module

The first draft flagged `test_dedup_vars` for patching `compute_vars.CATALOGS_DIR`. False positive: `dedup_stats` passes the directory to `latest_run_report` as an argument (0349 made it injectable), so the patch does reach it, and it is the idiom 0249 recommends. Flagging it would push authors toward the wrong fix, so the narrowing is pinned by its own test rather than left as a comment.

## The 0251 guard's exemption rested on a false claim

`TestNoWrongNamespacePatch` exempts bare attribute assignment, and named *these very tests* as legitimate: "a test patching utils' own namespace before calling a function that lives in utils itself". `save_run_report` lives in `pipeline_io`. That wrong justification is what let the leak run unflagged for four and a half months; its docstring now says so and points at the guard that does cover bare assignment.

## Second commit: the guard failing on its own terms

The pre-PR adherence pass found two blind spots, both closed here rather than deferred:

- Rebind analysis was **per file**, so a facade-only patch was exempted whenever any sibling function patched the definition site — one correct neighbour would launder every no-op beside it. Now grouped per enclosing function. `_patched_merge_dirs` does both in one fixture and stays exempt; verified, not assumed.
- The facade name was matched **literally** while the defining module was alias-resolved, so `import utils as u` slipped through — the exact limitation this ticket faults 0251 for accepting.

## Verification

Exit criterion 2 was discharged by **mtime, not content** — the leaked writes were byte-identical, so a content diff looked clean. A full suite run (1846 passed) left all 224 mtimes under `data/catalogs/` untouched in both the author's checkout and this worktree. The 7 failures are `test_corpus_acceptance` file-existence checks, the known absent-DVC-data gap in a worktree, unrelated to this change.

- `make lint` — 185 passed, 6 skipped
- `make check-fast` — 1217 passed, 11 skipped
- `erg check tickets/` — PASS (351 tickets)
- `/verify-adherence` — PASS, `mechanical_failures: []`

## One step deliberately left open

Deleting the eight leaked fixture files stays deferred, as the ticket recorded: sibling worktrees still run pre-fix `tests/` and would recreate them before the deletion could be committed. The recipe is in the ticket; run it once this merges and the other sessions rebase. The files are inert meanwhile — `latest_run_report` ignores every non-timestamp run id.

### DIFF
--- a/tests/test_robustness_observability.py
+++ b/tests/test_robustness_observability.py
@@ -99,62 +99,37 @@ class TestUtilsHelpers:
         # Allow duplicates in fast test envs but at least one call succeeds
         assert len(ids) >= 1
 
-    def test_save_run_report_creates_file(self, tmp_path):
-        # Patch CATALOGS_DIR to tmp_path for isolation
-        import utils
+    def test_save_run_report_creates_file(self, tmp_path, monkeypatch):
         from utils import save_run_report
 
-        orig = utils.CATALOGS_DIR
-        orig_pl = pipeline_loaders.CATALOGS_DIR
-        utils.CATALOGS_DIR = str(tmp_path)
-        pipeline_loaders.CATALOGS_DIR = str(tmp_path)
-        try:
-            data = {"elapsed_seconds": 1.5, "rows_written": 42}
-            path = save_run_report(data, "test-run-001", "my_script")
-            assert os.path.isfile(path)
-        finally:
-            utils.CATALOGS_DIR = orig
-            pipeline_loaders.CATALOGS_DIR = orig_pl
+        monkeypatch.setattr(pipeline_loaders, "CATALOGS_DIR", str(tmp_path))
+        data = {"elapsed_seconds": 1.5, "rows_written": 42}
+        path = save_run_report(data, "test-run-001", "my_script")
+        assert os.path.isfile(path)
 
-    def test_save_run_report_json_schema(self, tmp_path):
-        import utils
+    def test_save_run_report_json_schema(self, tmp_path, monkeypatch):
         from utils import save_run_report
 
-        orig = utils.CATALOGS_DIR
-        orig_pl = pipeline_loaders.CATALOGS_DIR
-        utils.CATALOGS_DIR = str(tmp_path)
-        pipeline_loaders.CATALOGS_DIR = str(tmp_path)
-        try:
-            data = {"counter_a": 10, "counter_b": 5, "elapsed_seconds": 2.0}
-            path = save_run_report(data, "run-xyz", "test_script")
-            with open(path) as f:
-                payload = json.load(f)
-            assert payload["script"] == "test_script"
-            assert payload["run_id"] == "run-xyz"
-            assert payload["counter_a"] == 10
-            assert payload["elapsed_seconds"] == 2.0
-        finally:
-            utils.CATALOGS_DIR = orig
-            pipeline_loaders.CATALOGS_DIR = orig_pl
+        monkeypatch.setattr(pipeline_loaders, "CATALOGS_DIR", str(tmp_path))
+        data = {"counter_a": 10, "counter_b": 5, "elapsed_seconds": 2.0}
+        path = save_run_report(data, "run-xyz", "test_script")
+        with open(path) as f:
+            payload = json.load(f)
+        assert payload["script"] == "test_script"
+        assert payload["run_id"] == "run-xyz"
+        assert payload["counter_a"] == 10
+        assert payload["elapsed_seconds"] == 2.0
 
-    def test_save_run_report_sanitizes_run_id(self, tmp_path):
-        import utils
+    def test_save_run_report_sanitizes_run_id(self, tmp_path, monkeypatch):
         from utils import save_run_report
 
-        orig = utils.CATALOGS_DIR
-        orig_pl = pipeline_loaders.CATALOGS_DIR
-        utils.CATALOGS_DIR = str(tmp_path)
-        pipeline_loaders.CATALOGS_DIR = str(tmp_path)
-        try:
-            path = save_run_report({}, "run id/with spaces&special!", "script")
-            assert os.path.isfile(path)
-            # filename should not contain spaces or slashes
-            basename = os.path.basename(path)
-            assert " " not in basename
-            assert "/" not in basename
-        finally:
-            utils.CATALOGS_DIR = orig
-            pipeline_loaders.CATALOGS_DIR = orig_pl
+        monkeypatch.setattr(pipeline_loaders, "CATALOGS_DIR", str(tmp_path))
+        path = save_run_report({}, "run id/with spaces&special!", "script")
+        assert os.path.isfile(path)
+        # filename should not contain spaces or slashes
+        basename = os.path.basename(path)
+        assert " " not in basename
+        assert "/" not in basename
 
     def test_retry_get_importable(self):
         from utils import retry_get
@@ -449,36 +424,28 @@ class TestRunReport:
             reports = os.listdir(reports_dir)
             assert not any("enrich_abstracts" in r for r in reports)
 
-    def test_save_run_report_counter_consistency(self, tmp_path):
+    def test_save_run_report_counter_consistency(self, tmp_path, monkeypatch):
         """Manually verify that the report structure is consistent."""
-        import utils
         from utils import save_run_report
 
-        orig = utils.CATALOGS_DIR
-        orig_pl = pipeline_loaders.CATALOGS_DIR
-        utils.CATALOGS_DIR = str(tmp_path)
-        pipeline_loaders.CATALOGS_DIR = str(tmp_path)
-        try:
-            missing_before = 10
-            missing_after = 3
-            total_filled = missing_before - missing_after
-            data = {
-                "missing_before": missing_before,
-                "missing_after": missing_after,
-                "total_filled": total_filled,
-                "elapsed_seconds": 5.0,
-            }
-            path = save_run_report(data, "consistency-test", "enrich_abstracts")
-            with open(path) as f:
-                payload = json.load(f)
-            assert (
-                payload["missing_before"] - payload["missing_after"]
-                == payload["total_filled"]
-            )
-            assert payload["elapsed_seconds"] > 0
-        finally:
-            utils.CATALOGS_DIR = orig
-            pipeline_loaders.CATALOGS_DIR = orig_pl
+        monkeypatch.setattr(pipeline_loaders, "CATALOGS_DIR", str(tmp_path))
+        missing_before = 10
+        missing_after = 3
+        total_filled = missing_before - missing_after
+        data = {
+            "missing_before": missing_before,
+            "missing_after": missing_after,
+            "total_filled": total_filled,
+            "elapsed_seconds": 5.0,
+        }
+        path = save_run_report(data, "consistency-test", "enrich_abstracts")
+        with open(path) as f:
+            payload = json.load(f)
+        assert (
+            payload["missing_before"] - payload["missing_after"]
+            == payload["total_filled"]
+        )
+        assert payload["elapsed_seconds"] > 0
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_run_report_isolation.py
+++ b/tests/test_run_report_isolation.py
@@ -68,42 +68,68 @@ def call_time_constants():
     return found
 
 
-def _module_aliases(tree):
-    """Names by which this file refers to the defining module."""
-    aliases = {DEFINING_MODULE}
+def _aliases_of(tree, module):
+    """Names by which this file refers to ``module``.
+
+    Applied to the facade as well as to the defining module: resolving one and
+    matching the other literally would leave ``import utils as u`` unflagged,
+    which is the alias blind spot the ticket-0251 guard already has.
+    """
+    aliases = {module}
     for node in ast.walk(tree):
         if isinstance(node, ast.Import):
             for a in node.names:
-                if a.name == DEFINING_MODULE and a.asname:
+                if a.name == module and a.asname:
                     aliases.add(a.asname)
     return aliases
 
 
-def _rebinds(tree, constants):
-    """Yield ``(module_name, constant)`` for every rebinding in the file.
+def _rebinds_in(node, constants):
+    """Yield ``(module_name, constant)`` for a rebinding performed by ``node``.
 
     Three shapes, all in use in this suite: bare attribute assignment
     (``utils.CATALOGS_DIR = x``), the ``monkeypatch.setattr`` string target
     (``setattr("pipeline_loaders.CATALOGS_DIR", x)``), and the two-argument
     object form (``setattr(pl, "CATALOGS_DIR", x)``).
     """
-    for node in ast.walk(tree):
-        if isinstance(node, ast.Assign):
-            for tgt in node.targets:
-                if (isinstance(tgt, ast.Attribute) and tgt.attr in constants
-                        and isinstance(tgt.value, ast.Name)):
-                    yield tgt.value.id, tgt.attr
-        elif isinstance(node, ast.Call) and getattr(node.func, "attr", None) == "setattr" \
-                or isinstance(node, ast.Call) and getattr(node.func, "id", None) == "setattr":
-            args = node.args
-            if len(args) >= 2 and isinstance(args[0], ast.Constant) \
-                    and isinstance(args[0].value, str) and "." in args[0].value:
-                mod, _, const = args[0].value.rpartition(".")
-                if const in constants:
-                    yield mod, const
-            elif len(args) >= 3 and isinstance(args[0], ast.Name) \
-                    and isinstance(args[1], ast.Constant) and args[1].value in constants:
-                yield args[0].id, args[1].value
+    if isinstance(node, ast.Assign):
+        for tgt in node.targets:
+            if (isinstance(tgt, ast.Attribute) and tgt.attr in constants
+                    and isinstance(tgt.value, ast.Name)):
+                yield tgt.value.id, tgt.attr
+    elif isinstance(node, ast.Call) and (
+            getattr(node.func, "attr", None) == "setattr"
+            or getattr(node.func, "id", None) == "setattr"):
+        args = node.args
+        if len(args) >= 2 and isinstance(args[0], ast.Constant) \
+                and isinstance(args[0].value, str) and "." in args[0].value:
+            mod, _, const = args[0].value.rpartition(".")
+            if const in constants:
+                yield mod, const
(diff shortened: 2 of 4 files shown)


MinhHaDuong (commented):
## `/review-pr` — five-agent panel (correctness, consistency, scope, red team, doc propagation)

**Verdict: comment.** The change is a real improvement over what it replaces and nothing in it regresses. Four gaps in the new guard are reproducible and worth recording; none of them is a defect this PR introduced, and none blocks the ticket's exit criteria (which are about *writes* into `data/catalogs/`).

**Gate evidence, run in this worktree:**
- `pytest -m adherence` — 185 passed, 6 skipped
- `pytest -m "not slow and not integration and not adherence"` — 1217 passed, 11 skipped
- `ruff check tests/` — clean

---

### Dissent, preserved

The red team returned **request-changes**; correctness returned **comment**; scope, consistency and doc propagation returned **approve**. The red team's case, verbatim:

> "A guard built to close a defect that cost 4.5 months of silent corpus writes, whose own two biggest blind spots are 'a production consumer that reads the facade at call time' and 'the test suite's own dominant mocking style,' has not closed the class of bug it says it closes."

I re-ran both claims myself and land on *comment* rather than *request-changes*, for reasons stated under finding 1. The author decides.

---

### Findings

**1. `verifiable:` The docstring premise "only `utils` is a pure re-export with no reader of its own" is false — there are call-time readers of the facade.**

`tests/test_run_report_isolation.py:139-146`. Three production sites do a function-local `from utils import <CONST>`, which reads `utils.<CONST>` at call time. `utils` snapshots the value at module import, so patching `pipeline_loaders` does **not** redirect these:

- `scripts/figures/plot_alluvial_html.py:85` — `from utils import CATALOGS_DIR, load_refined_embeddings` inside `load_paper_data`
- `scripts/analysis/compute_vars.py:381` — `from utils import EMBEDDINGS_PATH, REFINED_EMBEDDINGS_PATH`
- `scripts/analysis/compute_vars.py:535` — `from utils import REFINED_CITATIONS_PATH, REFINED_WORKS_PATH, normalize_doi`

Probe against the live helper — a test patching only the definition site while calling `load_paper_data` passes:

```
defsite patch for a facade-call-time reader: []
```

`call_time_constants()` matches only `node.module == DEFINING_MODULE`, so `EMBEDDINGS_PATH` and the three `REFINED_*` constants are never discovered at all.

**Why this is a comment and not a blocker.** The assertion message says "patch `pipeline_loaders.<CONST>` **too**", i.e. additive — an author following it keeps the facade patch and gains the definition-site one, which is correct for both consumer classes. And `load_paper_data` *reads* `refined_works.csv`; it does not write into DVC-tracked data, which is the leak class 0346 exists to stop. So this is a missing arm of the model, not a live hazard. It does deserve either a docstring correction or a follow-up ticket, because the sentence as written will be read as settled fact by the next author.

**2. `verifiable:` `_rebinds_by_scope` keys on the bare function name, so two same-named tests share one exemption bucket.**

`tests/test_run_report_isolation.py:121-132`. The per-function scoping is the headline fix of commit 57c62a50 — but `scope = node.name` collides across classes. A correct patch in `TestA.test_x` launders a genuine no-op in `TestB.test_x`, which is the file-level laundering the commit set out to eliminate, one nesting level down:

```
same-name methods in two classes: []      # expected: ['probe.py:test_x: CATALOGS_DIR']
```

Latent today — I grepped every test that touches `CATALOGS_DIR`/`POOL_DIR` and there is no name collision among them. Fix is a qualified scope path (ancestor chain tuple) instead of `node.name`.

**3. `verifiable:` keyword-argument `monkeypatch.setattr` is invisible.**

`tests/test_run_report_isolation.py:100-111` reads `node.args` only, never `node.keywords`:

```
keyword-arg monkeypatch.setattr: []
```

`monkeypatch.setattr(target=utils, name="CATALOGS_DIR", value=str(tmp_path))` is a valid pytest call and a pure facade-only no-op. Uncommon style; cheap to close.

**4. `consider:` `mock.patch` / `mock.patch.object` are not recognized.**

`_rebinds_in` matches only a func named `setattr`:

```
mock.patch string form: []
```

14 test modules import `unittest.mock`. None currently patches `CATALOGS_DIR` or `POOL_DIR` through it, and the guard this PR replaces did not cover it either — so no regression, but it is the one idiom most likely to reintroduce the defect under a future rewrite.

**5. `consider:` `conftest.py` and nested test directories are unscanned.**

`_test_sources()` uses `os.listdir(TESTS_DIR)` filtered to `test_*.py`. `tests/conftest.py` — the idiomatic home for an isolation fixture — is invisible to both guards in this file. It does no `CATALOGS_DIR`/`POOL_DIR` patching today and there are no `tests/**/test_*.py` subdirectories, so this is a coverage boundary, not an active hole.

**6. `consider:` an unparseable file under `scripts/` errors the guard rather than failing it.**

`call_time_constants()` calls `ast.parse()` on every `.py` under `scripts/` with no guard. A stray syntax error anywhere in the tree turns six adherence tests into tracebacks instead of an actionable assertion.

**7. `nofollow:` ticket-body numeric imprecision.**

The `18:35Z` log line reads "1846 passed, 7 known corpus-absent failures" (1853) against the body's "1,931-test run". 1931 is the real collection count (verified), so the gap is skipped tests uncounted in the log line. Cosmetic. Separately, "`_patched_merge_dirs` patches `utils.BASE_DIR` alongside `cm` and `pipeline_loaders`" reads as claiming parity across three modules; `pipeline_loaders.BASE_DIR` is never touched by that fixture (`tests/test_pipeline_e2e.py:52-77`). Harmless, since nothing re-imports `BASE_DIR` at call time — which the ticket itself says, and which I confirmed by AST scan.

**8. `nofollow:` `call_time_constants()` is the only public helper among five.**

Its siblings (`_aliases_of`, `_rebinds_in`, `_rebinds_by_scope`, `_facade_patch_offenders`) are all private, and `_facade_patch_offenders` is likewise called directly by tests. Cosmetic.

---

### Confirmed clean

**Doc propagation** verified every load-bearing factual claim in the new docstrings and the ticket body against source, all CONFIRMED: `POOL_DIR`'s call-time re-import in `pool_path`/`load_pool_ids`/`load_pool_records` plus the `os.makedirs` at `pipeline_io.py:278`; no test exercising the pool helpers; `save_run_report` living in `pipeline_io` and re-importing at `pipeline_io.py:213`; `dedup_stats` passing the directory to `latest_run_report` as an argument (`compute_vars.py:360`, parameter added by 0349 at `pipeline_io.py:145`); `BASE_DIR` having no call-time re-importer anywhere in `scripts/`; and `TestNoWrongNamespacePatch.FORBIDDEN` genuinely exempting bare attribute assignment, exactly as the corrected docstring now says. No stale reference anywhere in the repo to the old guard name `test_no_test_patches_catalogs_dir_on_the_facade`. Ticket 0251's own file does not repeat the falsified rationale, so no retroactive note is needed there.

**Scope** found the diff confined to four files, all implicated by 0346, with nothing unrelated. The generalization from a name-keyed regex to AST discovery is judged proportionate rather than over-engineered on the strength of what it actually found — `POOL_DIR`, a latent instance of the same defect that a `CATALOGS_DIR`-keyed sweep would have missed entirely. Commit 57c62a50 fixes blind spots in code this same PR introduced, which is finishing the deliverable, not scope creep. The `test_script_hygiene.py` docstring rewrite is a direct, cited consequence of the sweep. The one unticked exit-criterion box (fixture deletion) is deferred with a recorded and sound rationale.

**Consistency** found the missing type hints, the backslash continuations, the `adherence` markers and the long narrative docstrings all consistent with sibling files in `tests/` — including `TestNoWrongNamespacePatch` in this very diff. `E501` is not in the ruff select list, so the long lines are a non-issue.

---

### Suggested follow-up

Findings 1–3 are cheap and mechanical: scan `node.keywords` in `_rebinds_in`, key `_rebinds_by_scope` on a qualified ancestor path, and either correct the "no reader of its own" sentence or extend `call_time_constants()` to discover function-local `from utils import <CONST>` as a second, symmetric class. Worth one follow-up ticket rather than a reroll here — the guard as merged is strictly stronger than the regex it replaces, and holding the PR would leave the weaker one on main.

🤖 `/review-pr` — comments only, no fixes applied, no merge gate. Merge remains the author's call.

MinhHaDuong (commented):

Correctness · Consistency · Scope · Red team · Doc propagation. All read-only, `model: sonnet`. Gate run in this worktree: `make lint` 187 passed / 6 skipped; fast tier 1217 passed / 11 skipped. Both match the PR body.

**No blocker.** The guard does what it claims on every shape the suite actually contains — `_facade_patch_offenders(_test_sources())` returns `[]`, `call_time_constants()` returns exactly `{CATALOGS_DIR, POOL_DIR}`, and the alias case the 0251 guard admits it misses (`import utils as u`) is genuinely closed. Every factual claim the panel spot-checked held: `compute_vars.py:381,535` and `plot_alluvial_html.py:85` do function-local `from utils import …`; `pool_path` does call `os.makedirs`; 0349 did make `latest_run_report`'s `catalogs_dir` injectable; the 0251 exemption rationale really was false. That is a high bar met, and it is the bar this PR sets for others.

**Dissent, preserved.** Red team returned **request-changes**; the other four returned comment/approve. Its case verbatim: *"The two comment/docstring-substring findings are the most concerning — they are silent, accidental, and directly undermine the stated purpose of the guard … with a one-line comment that requires no malicious intent."* I did not adopt that verdict, because the line it targets is pre-existing (`git show origin/main:tests/test_run_report_isolation.py` line 104) and untouched by this diff — pre-existing-and-untouched routes to a ticket, not to a bounce. The author may weigh it differently.

---

### One decision before merge

**`Ticket:` closes 0346 while its own Action 2 is undone, and nothing else tracks it** — *scope, high confidence*, `verifiable:`

The PR body's `**Ticket:** tickets/0346-…` closes 0346 on merge. The three Exit criteria are all `[x]` with real evidence, so `erg-pr-merge` closing it is legitimate. But the Verification box `[ ] The eight existing fixture files are gone` stays unchecked, and the deferral is defensible on its stated grounds — sibling worktrees still run pre-fix `tests/` and would recreate the files. The gap is bookkeeping: a repo-wide search found no other ticket referencing that cleanup, so once 0346 closes, the `rm` + `dvc add` recipe survives only as prose inside a closed ticket, where no ticket sweep will surface it. Either demote to `Ticket-ref:` and close 0346 once the deletion lands, or file a one-line follow-up ticket now.

---

### Findings worth acting on, none blocking

**1. `call_time_constants()` hand-rolls `os.walk(SCRIPTS_DIR)` instead of the sanctioned enumeration** — *consistency, high confidence*, `verifiable:`
`tests/test_run_report_isolation.py:66`. `tests/_script_discovery.py::all_script_files()` documents itself as "the ONE sanctioned script enumeration — every guard routes through it … so a relocation can never narrow a guard's file set" (ticket 0260), and closes exactly the defect class of hand-rolled per-guard walkers. `test_script_hygiene.py` already routes through it. The new walker also diverges on coverage: `all_script_files()` excludes `scripts/archive*/`, this one includes it. No live difference today (nothing under `archive*/` re-imports from `pipeline_loaders`), so this is a latent divergence, not a bug — but the wider set for a *discovery* helper is a choice that should be stated rather than incidental.

**2. Two residual facade-alias shapes the new `_aliases_of` does not resolve** — *red team, high confidence*, `verifiable:`
Reproduced against the branch's own helpers:

```
control_bare       -> ['control_bare.py:test_x: CATALOGS_DIR']
alias_via_assign   -> []      # u = utils ;  u.CATALOGS_DIR = str(tmp_path)
dotted_string      -> []      # monkeypatch.setattr("scripts.utils.CATALOGS_DIR", …)
tuple_target       -> []      # utils.CATALOGS_DIR, other = str(tmp_path), 1
patch_multiple     -> []      # mock.patch.multiple(utils, CATALOGS_DIR=…)
sysmodules         -> []      # sys.modules["utils"].CATALOGS_DIR = …
```

`_aliases_of` walks `ast.Import` with an `asname` only, so name-to-name aliasing is invisible; `_rebinds_in`'s `ast.Assign` branch requires the target to be a bare `ast.Attribute`, so `ast.Tuple` and `ast.Subscript` targets fall through; `rpartition` on the string form yields `mod="scripts.utils"`, absent from the alias set. Of these, `dotted_string` is the least contrived — `tests/test_regression.py:153` already writes `import scripts.analysis.compute_regression_hashes as mod`, so `scripts.`-qualified references are established style here (and `scripts/` has no `__init__.py`, so such an import would produce a *second* module object, making that patch doubly a no-op). The rest are shapes nobody writes by accident.

**Reuse pointer that covers most of this and finding 3:** `tests/test_arch_compliance.py::TestCorpusThroughLoaders` already carries a forward-pass alias detector with `_ALIAS_STMT` taint propagation through `x = y`, taint decay on rebind, and up-front comment stripping — pinned by `test_detector_ignores_alias_in_comment` and `test_detector_decays_taint_on_reassignment` (tickets 0198/0202). That is the same problem solved once in this repo already. Worth a follow-up that lifts it into a shared helper rather than a third hand-rolled walker.

**3. The spawn guard's module-wide exemption is a raw substring of the file** — *red team, high confidence*, `verifiable:`
`tests/test_run_report_isolation.py:263`, `if 'setenv("CLIMATE_FINANCE_DATA"' in src`. **Pre-existing on `origin/main`, not introduced here.** A comment or docstring merely *mentioning* the call exempts the whole module while a genuine `cwd=REPO_ROOT` offender sits below it; conversely `monkeypatch.setenv('CLIMATE_FINANCE_DATA', …)` with single quotes loses the exemption and false-positives (no `ruff format` step normalizes quotes in this repo). Since this PR widened `_test_sources()` to feed that guard more files, both directions now have more surface. Ticket, not a bounce.

**4. `test_discovery_is_not_vacuous` pins only `CATALOGS_DIR`** — *correctness, medium confidence*, `consider:`
The whole argument of the PR is that the guard must not be keyed on the constant that bit us. The anti-vacuity pin is. If `POOL_DIR`'s function-local imports were refactored away, discovery would silently narrow to one constant and every test here would still pass. A set-equality pin (`{CATALOGS_DIR, POOL_DIR}`) would fail loudly and force a deliberate re-derivation — the same reasoning the file already applies to `CATALOGS_DIR`.

**5. `_facade_patch_offenders` does not guard `ast.parse` against `SyntaxError`** — *correctness, medium confidence*, `consider:`
`call_time_constants()` wraps its parse and skips unparseable files with a stated rationale; `_facade_patch_offenders` (line 193) does not, so an unparseable test module fails the guard with a `SyntaxError` traceback rather than a guard message. Arguably correct — an unparseable *test* file is a real problem — but the asymmetry between the two is unexplained.

**6. `test_script_hygiene.py:916` undersells the overlap it describes** — *consistency, high confidence*, `verifiable:`
The rewritten docstring says bare assignment "is now covered by" the new guard, implying a clean split (old guard = `setattr` forms, new = bare assignment). `_rebinds_in` also matches `setattr` and `mock.patch`, as `test_facade_guard_sees_past_the_call_shape` proves. For `CATALOGS_DIR`/`POOL_DIR` the two guards now fully overlap, and a regression fails two tests in two files with different messages. Not wrong, but the docstring reads as a handoff where it is a superset.

**7. Ticket rationale for `monkeypatch` overclaims** — *consistency, medium confidence*, `consider:`
Ticket Actions §1: hand-rolled save/restore "leaks if an assertion fails midway." `finally` runs on `AssertionError` too. The switch is still right (less code, no restore to forget, teardown ordering), but the stated reason is not the one that holds.

**8. Line-number citations in docstrings** — *doc propagation*, `consider:`
`_facade_patch_offenders`'s docstring cites `compute_vars` (:381, :535) and `plot_alluvial_html` (:85). All three resolve correctly today — verified. Nothing tests that they still will. Naming the function instead of the line costs nothing.

**9. `call_time_constants` is the only public helper in a file of private ones** — *consistency*, `consider:`. Used only within the module.

**10. Narrative triplication** — *scope*, `nofollow:`. The POOL_DIR / facade-scoping rationale appears near-verbatim in the ticket body, the PR description, and two docstrings. Harmless; three places to keep in sync if the reasoning is ever corrected again.

**11. The "patch the defining module, not the facade" lesson is in tickets and tests, never in a rule file** — *doc propagation*, `consider:`. It has now cost three tickets (0249 → an incomplete 0251 guard → 0346). A line in `.claude/rules/` would be cheaper than a fourth.

---

### Checked and clean

- The `monkeypatch` conversion in `test_robustness_observability.py` is safe in all four blocks: `utils.save_run_report` **is** `pipeline_io.save_run_report` (same object, single re-export), and that function never reads `utils`, so dropping the facade patch loses nothing.
- `.isupper()` filters correctly — `"Foo_BAR".isupper()` is `False`, `"x2".isupper()` is `False`.
- `test_save_run_report_honours_the_patched_definition_site` cannot escape `tmp_path`, and its `startswith` is on a true child path (`catalogs_dir` + `"run_reports"`), so no sibling-prefix collision.
- Fast-tier placement of that unmarked test is correct: no heavy import, no subprocess, writes only under `tmp_path`.
- No production code touched, no drive-by refactor, no unrelated hunk. The `_test_sources()` widening and the `test_script_hygiene.py` docstring correction are both load-bearing for the sweep.
- Ticket 0251's own body does not contain the falsified rationale — it lived only in the `test_script_hygiene.py` docstring (commit 73f5c4e3), which this PR corrects. No closed ticket needs amending.

*Panel verdicts: correctness comment · consistency comment · scope comment · red team request-changes · doc propagation approve. Review only — no fixes applied, no merge.*

PR #1196, opened 2026-07-27
fix(0359): make paths.mk name each deliverable's real artifacts, and guard it

MinhHaDuong:
## What

`paths.mk` carries two per-deliverable contracts — `*_INCLUDES` (the shared files a document composes) and `*_FIGS` (the figures it embeds). Both are render-rule prerequisites, not just documentation, and both had drifted.

The ticket's sweep found 22 unrendered artifacts. Resolving the include graph from all 11 `.qmd` roots showed the `*_INCLUDES` side was worse:

| Variable | Declared but not composed | Composed but not declared |
|---|---|---|
| `TECHREP_INCLUDES` | 11 | 23 |
| `DATAPAPER_INCLUDES` | 4 | — |
| `MULTILAYER_INCLUDES` | 6 (all of them) | — |
| `CORPUS_REPORT_INCLUDES` | — | 1 |

Three render rules named no figure at all. Verified against `origin/main`, not inferred: `make deliverables/multilayer/multilayer-detection.pdf` renders happily with none of its five figures on disk, because its only artifact prerequisite was six includes it had stopped composing.

## Changes

- Both contracts corrected against each document's real transitive include closure. `TECHREP_INCLUDES` is now `$(ZOO_INCLUDES)` plus one section; `MULTILAYER_INCLUDES` is empty (the paper composes no shared include).
- Missing figure prerequisites wired into the corpus-report, data-paper, multilayer and multilayer-techrep render rules.
- New `CORPUSREPORT_FIGS` (`fig_dag` was filed under the data paper) and `SLIDES_FIGS` (the decks embed pipeline figures and nothing declared them).
- Two figures documents embed that no variable declared gained homes: `fig_companion_sensitivity`, `fig_composition_wide`. `ALL_FIGS` goes 34 → 35: those two added, `fig_sem_composition` removed — checked old-vs-new by expanding both versions.
- **`fig_sem_composition` deleted** (author decision, ticket Action 1) along with its two `.mk` rules, the `_clean` clustering chain that fed only them, its short-label config, and `tab_sem_composition.csv`. Orphaned by 0332's §4 cut, its label objection then refuted by 0328 — withdrawn for scope, not error. `tab_sem6_assignments.csv` keeps its own producer, so §1's `lit_*` bullet is untouched.
- New `tests/test_deliverable_artifacts.py` covering figures, includes and tables: 39 tests, both directions per contract, plus reachability over tracked Markdown under `_shared/{_includes,tables}`.

## The allowlist is the load-bearing part

Without it the guard either fails forever or gets deleted. Two hatches, both redundancy-checked so an entry that becomes reachable **fails** the suite:

- `# not-embedded: <file> — <reason>` in `paths.mk` — figures built on purpose, embedded nowhere.
- `config/unrendered-artifacts.txt` — includes and tables no document composes.

The 17 flat orphaned includes are **explained, not deleted**: their content disposition is ticket 0290's Action 2, an author call. This PR stops them posing as live render prerequisites and delivers 0290's Action 3 (the hygiene guard).

## Verification

- Mutation-checked six ways — drop an allowlist entry, allowlist a reachable file, declare a dead include, orphan a figure, orphan an include, unwire a render rule. Each fails exactly the test that should catch it; baseline green.
- The guard proved itself during the rebase: it caught `tab_retrieval_protocol.md`, newly deposited by ticket 0329 and composed by no document. Allowlisted with its reason.
- `make lint` 230 passed · `make check-fast` 1300 passed. Full `make check`: 1907 passed, 7 failed — all `test_corpus_acceptance.py` file-existence checks for corpus data absent from this worktree; the test file is byte-identical to `origin/main`, so they predate this branch.

## Judgement calls worth review

- `tab_venues_fr_caption.md` looked dead to the sweep but is not — `manuscript.mk` concatenates it into the Gide venues table. Allowlisted as a build input rather than deleted.
- `test_makefile_contract`'s citation-coverage pin moved from `TECHREP_INCLUDES` to `CORPUS_REPORT_INCLUDES`: the corpus report is `citation-quality.md`'s only consumer, so the test had been pinning a transitive edge that no longer existed.
- `CORPUS_REPORT_FIGS` is spelled to match its sibling `CORPUS_REPORT_INCLUDES`, not the ticket's `CORPUSREPORT_FIGS`.
- Artifacts are enumerated from `git ls-files`, not disk — a fresh worktree holds only the committed handoff subset, so a disk-based guard's verdict would be machine-dependent.

**Ticket:** tickets/0359-datapaper-figs-lists-three-figures-the-p.erg
**Ticket-ref:** tickets/0290-orphaned-shared-includes-audit.erg


MinhHaDuong (commented):

**Verdict: comment.** Two agents returned `request-changes`; I verified both claims independently and neither survives as a blocker, but both point at real questions. Dissent is preserved verbatim below — the author decides.

`make lint` 230 passed, `make check-fast` 1300 passed, in this worktree.

The mechanical core of the PR checks out. Verified independently, not taken on the body's word:

- The include resolver keeps the **root's** directory as the base throughout (`tests/test_deliverable_artifacts.py:174-189`, `top` is a loop invariant outside the `while pending`), so it avoids the false-positive trap the ticket documents. Cycles are guarded by a `seen` set.
- `ALL_FIGS` old-vs-new, expanded via `make -pn` in a throwaway `origin/main` worktree and diffed sorted: `+fig_companion_sensitivity.png`, `+fig_composition_wide.png`, nothing removed. The claim holds exactly.
- Two of the six claimed mutation-kills reproduced live (unwire `CORPUSREPORT_FIGS` from `corpus-report.mk`; declare a dead include in `CORPUS_REPORT_INCLUDES`); each failed exactly the named test, baseline green.
- All 17 flat includes named in the ticket sweep appear in `config/unrendered-artifacts.txt`, 1:1, none missing or extra. Both named judgement calls confirmed in code: `tab_venues_fr_caption.md` is concatenated by `manuscript.mk:62-63`, `codebook.md` is copied by `build/build_datapaper_archive.sh:47`.

---

### Dissent 1 — red team called a clean-room regression; the panel's other members did not

**Red team (request-changes, high confidence):** wiring `$(MULTILAYER_FIGS)` and `$(ZOO_*)` into `deliverables/multilayer/multilayer.mk:22,33` breaks a build that previously worked. Reproduced: `make -n multilayer-detection` on this branch dies with *No rule to make target `deliverables/_shared/figures/fig_companion_zseries.png`*; on `origin/main` the same invocation renders.

**My check widens the frame.** I ran all five render targets on both refs:

| target | `origin/main` | branch |
|---|---|---|
| data-paper | fails (`tab_corpus_sources.md`) | fails (same) |
| corpus-report | fails (`tab_corpus_sources.md`) | fails (same) |
| technical-report | fails (`tab_citation_coverage.md`) | fails (`schematic_S1_mmd.png`) |
| multilayer-detection | **renders** | fails (`fig_companion_zseries.png`) |
| multilayer-techrep | **renders** | fails (`schematic_S1_mmd.png`) |

Three of five already failed this way before the PR. The two that "rendered" are the exact defect this PR exists to fix — the body says so: they render happily with none of their five figures on disk. Failing loudly beats emitting a figure-less PDF, so I do not read this as a regression to revert.

`verifiable:` What it does change is the count: `make papers` on a fresh clone now stops at 5 of 5 rather than 3 of 5, because `papers:` (`Makefile:701`) depends on the render targets only, never on the `figures-*` targets (`Makefile:668-676`). Anyone whose habit is `make papers` without a prior `make figures` will meet the new error.

`consider:` The sharper signal underneath: **`fig_companion_zseries.png` exists nowhere** — not tracked, not in the author's primary checkout, not in this worktree. The multilayer paper's figure set is genuinely absent, which the old missing prerequisite was hiding. That is worth knowing, and it is upstream of this PR rather than caused by it. Per the harness rule that handoff artifacts a downstream workpackage consumes belong in git, the durable fix is committing them (or making `papers` depend on `figures-*`); either belongs in its own ticket, not here.

### Dissent 2 — scope says the ticket's one real question was allowlisted rather than answered

**Scope (request-changes, high confidence):** ticket 0359 Action 1 asked for a decision on `fig_sem_composition` — keep it with a stated consumer, or delete the figure, its `.mk` rules and `tab_sem_composition.csv` together. The PR keeps it, naming `deliverables/data-paper/revision-rdj26561/literature-confirmations.md` as consumer. That file is composed by no document; it is a revision-round process artifact of closed ticket 0310. Scope's reading: the marker mechanism used everywhere else in this PR for artifacts with a real pinned test is here being used to re-label "orphaned" as "not-embedded, stated consumer."

**My check partly rebuts it.** `fig_sem_composition.png` is deliberately committed (`.gitignore:46` whitelists it) and pinned by `tests/test_lit_confirmations.py`. The marker's own definition is *built on purpose, embedded nowhere*, which is precisely this figure's status, so the mechanism is being used as designed rather than stretched. The reason text at `paths.mk:114-119` is candid about it, naming 0332's §4 cut as the cause.

`consider:` **This still wants the author's eye before merge.** The ticket called it "the only real question here," and whether a revision-response artifact justifies keeping a maintained figure, its compute rules and its backing CSV is a content call, not a build call. The panel is split; the author is the tiebreak.

---

### Findings the panel agreed on

`verifiable:` **The PR description claims a rename that never happened.** "`tests/test_figure_sets.py` → `tests/test_deliverable_artifacts.py`" appears in the body and the merge commit will carry it. `git log --all --oneline -- tests/test_figure_sets.py` returns nothing: the file has never existed in this repo. The new test file is net-new in a single-commit PR. Three of five reviewers found this independently. No code impact; correct the body so the history stays honest.

`consider:` **`CORPUSREPORT_FIGS` vs `CORPUS_REPORT_INCLUDES`** (`paths.mk:128`). Every other deliverable spells its `*_FIGS` and `*_INCLUDES` pair the same way; corpus-report is the one place they diverge, and the split propagates into `Makefile` (`figures-corpusreport`) and the new test. The ticket itself proposed the squashed spelling, so it is deliberate — a one-line rename or a "why" note would close it.

`consider:` **Test tiering.** `test_shared_artifacts_are_reachable` and `test_allowlist_entries_are_earned` (`tests/test_deliverable_artifacts.py:281,294`) shell out through `_tracked()` → `subprocess.run(["git","ls-files",…])` but carry only `@pytest.mark.adherence`. The sibling with the identical pattern, `tests/test_handoff_artifacts_tracked.py:54`, stacks `adherence` + `integration`. No functional impact (adherence is already out of `check-fast`, and the 39 tests run in 0.08s) — a classification inconsistency only.

`consider:` **`NOT_EMBEDDED_RE` captures only the first line of a reason** (`tests/test_deliverable_artifacts.py:104`, no `re.S`). Both live markers in `paths.mk` are multi-line, so the "reason is load-bearing" word-count check is really validating line one. Harmless today, a latent hole in the guarantee.

`consider:` **The include regex has no comment-awareness.** A `{{< include … >}}` inside a fenced block or an HTML comment — a document explaining the syntax, say — would count as reachability. No such instance exists in the tree today, so this is a note for whoever writes the build documentation, not a defect.

`nofollow:` `_tracked()` runs `git ls-files` with `check=True` and no fallback, so a `.git`-less copy of the tree errors with a plumbing exception rather than skipping. The archive builders do not run the suite; enumerating from git rather than disk is the right call and the PR body defends it well.

`nofollow:` The 39-test count was verified against the parametrization and each instance covers a distinct failure mode. Not over-engineered.

---

### For the author, before merge

1. `fig_sem_composition` — keep, or delete figure + `.mk` rules + `tab_sem_composition.csv` together? Ticket 0359 Action 1, unresolved in the panel's view.
2. The multilayer figures do not exist anywhere. Ticket separately, or is their absence already known?
3. Fix the "renames `test_figure_sets.py`" line in the PR body.

PR #1198, opened 2026-07-27
ticket(0359): reachability guard for generated shared artifacts

MinhHaDuong:
## What

A guard that fails when the build graph produces an artifact no document consumes, plus the two `paths.mk` fixes the ticket asked for.

`tests/test_artifact_reachability.py` (adherence tier) checks three artifact classes under `deliverables/_shared/`:

| check | universe | rule |
|---|---|---|
| includes | every `_includes/**/*.md` on disk | transitively reachable from one of the 11 deliverable `.qmd` roots |
| figures | every `figures/*.png` named in the build graph | displayed by some prose file |
| markdown tables | every `tables/*.md` named in the build graph | included by some prose file |

Plus `test_allowlist_has_no_redundant_entries`, so an exemption dies the moment its artifact is consumed again — without it the allowlist rots into a permanent ignore-list.

Red-then-green: the first commit fails with an empty allowlist; at HEAD, emptying `ALLOWED` yields exactly 41 orphans — equal to the allowlist keys, no gaps and no redundant entries.

## Two design calls worth reviewing

**Two edge types, two checks — not one strict transitive walk.** Includes get strict reachability from a root. Figures and tables get "displayed by *some* prose file", reachable or not. A single strict walk would report 25 orphaned figures instead of 12, because 13 are displayed only from the 17 orphaned includes — one root cause, reported three times. It would also hide the useful property this split gives: when ticket 0290 settles those includes, those 13 figures newly fail on their own and force their follow-up.

**Universes are chosen so the verdict never depends on whether `make` has run.** Includes come from disk (all committed handoff artifacts); figures and tables from the build graph (most are gitignored and regenerable). CSV and JSON are excluded — scripts read them, documents do not. Dynamic outputs (`fig_lexical_tfidf_{year}.png`, behind `.lexical_tfidf.stamp`) are in no static universe and out of scope; that is stated in the docstring.

## The trap

Quarto resolves `{{< include >}}` against the **top rendering document's** directory, not the including file's. The resolver carries the root's directory as base through every level of recursion. `test_nested_zoo_includes_are_reachable` pins it: the 18 `_includes/zoo/*.md` are reached only via `techrep-zoo.md` as `../_shared/_includes/zoo/…` from the zoo deliverable's folder, and the natural wrong implementation reports all 18 as orphans. That defect is what produced this ticket's 13 false positives. `/verify-adherence` mutation-tested it — swapping in the wrong-base resolver fails the test.

## Scope: what this PR deliberately does NOT do

**The 17 flat includes are explained, not settled.** Ticket 0290 owns their disposition and its action 2 sends any include not clearly superseded by the `techrep/` set to the author. Deleting or reviving them here would pre-empt an author call. They are allowlisted under one grouped reason naming 0290. Conversely 0290's action 3 asked for exactly this guard, so that arm is now supplied.

**`fig_sem_composition` is kept, not deleted.** It lost its consumer to 0332's §4 cut today, but the data paper is mid-R&R and §4 may return in round 2. Retiring a committed figure together with its `lit-confirmations.mk` rules mid-revision is the author's call. The reasoning is in the ticket log so it can be overruled cheaply: delete the allowlist entry, `SEM_COMPO_FIG`/`SEM_COMPO_TAB`, and `tab_sem_composition.csv` together. `tab_sem6_assignments.csv` and §1's literature-confirmation bullet are untouched either way.

## paths.mk

`fig_dag` moves from `DATAPAPER_FIGS` to a new `CORPUSREPORT_FIGS` — `corpus-report.qmd:25` is its only consumer, and it sat in the data paper's list only because no corpus-report list existed. `ALL_FIGS` absorbs the new variable and is **byte-identical**: 35 paths, `make --eval` on the pre- and post-change trees, empty `diff`. So `make figures` builds exactly the same set. `figures-corpusreport` added for symmetry with the other five per-deliverable targets.

The co-citation map's built-not-embedded status is now commented above `DATAPAPER_FIGS`, pointing at `tests/test_global_map.py:175` — a reader no longer has to grep to learn it is intentional.

## Sweep correction

The re-run found **41** orphans, not the 22 the filing estimated: 17 includes (as filed), 3 markdown tables (the filing missed `tab_core_venues_top10.md`), and 21 declared figures. Recorded in the ticket log.

## Verification

- `make lint` — 187 passed, 6 skipped (was 180; +7 from this guard).
- `make check` — 1848 passed, 7 failed. All 7 are `test_corpus_acceptance.py` file-existence checks: `.env` sets `CLIMATE_FINANCE_DATA=data` relative, so it resolves to this worktree's own `data/`, which holds only `.dvc` pointers. Pre-existing worktree condition, unrelated to this diff. The three ticket-0366 failures expected on main are gone — 0366 merged at c852824c.
- `erg check tickets/` — PASS (352 tickets).
- `/verify-adherence` — PASS, two nits, both fixed in the last commit (adherence marker; author field in the log entries).

**Ticket:** tickets/0359-datapaper-figs-lists-three-figures-the-p.erg

---

## Review rounds 1 and 2 — what changed

**Round 1 blocker: the universe was blind to composed variables.** The build graph was read by text scan alone, so figure lists make builds with `$(foreach)` were invisible — `BIAS_FIGS` in `zoo-figures.mk`, `SENS_FIG_PCA`/`SENS_FIG_JL` in `divergence.mk`. Nine real orphans escaped while the suite passed. A text scan cannot reach them even in principle: `$(ZOO_FIGS)/fig_zoo_bias_$(m).png` has neither the directory nor the stem as literal text. Rather than hand-mirror the lists as literals (the repo's existing habit for `ZOO_RESULT_FIGS`), the guard now enumerates assigned variable names lexically and lets **make** expand them — so the universe also cannot narrow when someone refactors a literal list into a function, which is the failure this PR had just suffered from the other side. `test_variable_composed_figures_are_in_the_universe` pins it.

**Round 2, two silent-pass gaps, both fixed rather than deferred** — each is the exact failure mode the guard exists to prevent:

- *Render fragments went unexpanded.* The probe ran `make -f Makefile`, which `-include`s the eight Phase-2 concern fragments but never the six per-deliverable render `.mk` (`make papers` invokes those with their own `$(MAKE) -f`). Every variable they assign expanded to nothing. The probe now runs once per entry point; the universe goes 80 → 86 paths.
- *Commented-out prose counted as a consumer.* `<!-- x -->` read as displayed. 0332 orphaned `fig_sem_composition` by deleting its paragraph — had it commented the paragraph out instead, this ticket would never have been filed. HTML comments are stripped before any prose scan, pinned by `test_commented_out_prose_is_not_a_consumer`.

Also dual-marked `integration` alongside `adherence` (four checks spawn `make`), matching `test_handoff_artifacts_tracked` and `test_phase_layout`.

`tests/_mk_discovery` supplies every makefile list — `test_no_handrolled_mk_enumeration` caught this file twice, and was right both times.

## ⚠️ Collision with PR #1195 — needs a human decision before either merges

PR #1195 (ticket 0290, `t290-orphaned-includes`) is open and overlaps this work substantially. It deletes 14 of the 17 flat includes, relocates 3, rewrites `paths.mk`'s include sets, adds a competing guard `tests/test_include_graph.py`, and files ticket 0373 on figure-set drift.

The branches merge **without textual conflict**, so `git` will not warn anyone. I built the combined tree and ran the suite against it: **3 of 7 checks go red** — 17 allowlist entries name files #1195 deletes, 13 figures lose their last prose consumer, 1 table follows. The failure is loud and self-describing (`test_allowlist_has_no_redundant_entries` names each stale entry), so nothing lands silently broken, but whoever merges second owns the reconciliation.

Two guards covering the include half is duplication worth resolving deliberately rather than by merge order.

**Ticket-ref:** tickets/0290-orphaned-shared-includes-audit.erg

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,10 +143,10 @@ NCC_FIGS        := deliverables/_shared/figures/fig_ncc_divergence.png \
 # Zoo figures (ZOO_SCHEMATICS / ZOO_RESULT_FIGS, defined in paths.mk) are
 # deliberately excluded from ALL_FIGS / make figures: plot scripts not yet wired.
 # Add them to ALL_FIGS when plot_schematic_*.py and plot_zoo_results.py have targets.
-ALL_FIGS := $(MANUSCRIPT_FIGS) $(DATAPAPER_FIGS) $(MULTILAYER_FIGS) $(TECHREP_FIGS) $(NCC_FIGS)
+ALL_FIGS := $(MANUSCRIPT_FIGS) $(DATAPAPER_FIGS) $(CORPUSREPORT_FIGS) $(MULTILAYER_FIGS) $(TECHREP_FIGS) $(NCC_FIGS)
 
 # ── Default target ────────────────────────────────────────
-.PHONY: all setup manuscript papers corpus-report technical-report data-paper multilayer-detection multilayer-techrep zoo figures figures-manuscript figures-datapaper figures-companion figures-techrep figures-ncc stats check check-package check-fast lint test-durations venv-canonicalize smoke benchmark determinism-check regression regression-update audit-pdf-content check-corpus check-manuscript-data data corpus corpus-sync corpus-discover corpus-enrich corpus-extend corpus-filter corpus-align corpus-filter-all corpus-tables corpus-validate deploy-corpus clean rebuild archive-analysis archive-manuscript archive-datapaper analysis-figures analysis-tables analysis-stats manuscript-render manuscript-figures datapaper-render datapaper-figures corpus-handoff
+.PHONY: all setup manuscript papers corpus-report technical-report data-paper multilayer-detection multilayer-techrep zoo figures figures-manuscript figures-datapaper figures-corpusreport figures-companion figures-techrep figures-ncc stats check check-package check-fast lint test-durations venv-canonicalize smoke benchmark determinism-check regression regression-update audit-pdf-content check-corpus check-manuscript-data data corpus corpus-sync corpus-discover corpus-enrich corpus-extend corpus-filter corpus-align corpus-filter-all corpus-tables corpus-validate deploy-corpus clean rebuild archive-analysis archive-manuscript archive-datapaper analysis-figures analysis-tables analysis-stats manuscript-render manuscript-figures datapaper-render datapaper-figures corpus-handoff
 
 .DEFAULT_GOAL := manuscript
 
@@ -626,6 +626,7 @@ deliverables/_shared/figures/fig_ncc_alluvial.png: \
 
 figures-manuscript: corpus-handoff $(MANUSCRIPT_FIGS)
 figures-datapaper:  corpus-handoff $(DATAPAPER_FIGS)
+figures-corpusreport: corpus-handoff $(CORPUSREPORT_FIGS)
 figures-companion:  corpus-handoff $(MULTILAYER_FIGS)
 figures-techrep:    corpus-handoff $(TECHREP_FIGS)
 figures-ncc:        corpus-handoff $(NCC_FIGS)
--- a/paths.mk
+++ b/paths.mk
@@ -103,10 +103,21 @@ ZOO_INCLUDES := deliverables/_shared/_includes/techrep/overview.md \
 # Phase-2 rules that PRODUCE them live in the concern .mk (root), not here.
 MANUSCRIPT_FIGS := deliverables/_shared/figures/fig_bars_v1.png deliverables/_shared/figures/fig_composition.png deliverables/_shared/figures/fig_breaks.png
 
-DATAPAPER_FIGS  := deliverables/_shared/figures/fig_bars.png deliverables/_shared/figures/fig_dag.png \
+# fig_global_map_cocitation is built on purpose and embedded in no document: the
+# data paper carries the direct-citation map, the co-citation map backs the R1-14
+# reviewer response (ticket 0307). tests/test_global_map.py:175 pins that by
+# asserting the basename is absent from the data paper source — do not "fix" it
+# by embedding the figure. fig_sem_composition lost its consumer when ticket 0332
+# cut §4's semantic-cluster paragraph; kept pending the round-2 decision.
+# Both are recorded in tests/test_artifact_reachability.py's allowlist.
+DATAPAPER_FIGS  := deliverables/_shared/figures/fig_bars.png \
                    deliverables/_shared/figures/fig_global_map_direct.png \
                    deliverables/_shared/figures/fig_global_map_cocitation.png \
                    deliverables/_shared/figures/fig_sem_composition.png
+
+# fig_dag sat in DATAPAPER_FIGS only because no corpus-report figure list
+# existed; corpus-report.qmd:25 is its sole consumer (ticket 0359).
+CORPUSREPORT_FIGS := deliverables/_shared/figures/fig_dag.png
 
 MULTILAYER_FIGS  := deliverables/_shared/figures/fig_breakpoints.png deliverables/_shared/figures/fig_alluvial.png \
                    deliverables/_shared/figures/fig_breaks.png \
--- a/tests/test_artifact_reachability.py
+++ b/tests/test_artifact_reachability.py
@@ -57,6 +57,8 @@ import glob
 import os
 import re
 
+from _mk_discovery import all_makefiles
+
 REPO_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 DELIVERABLES = os.path.join(REPO_ROOT, "deliverables")
 SHARED = os.path.join(DELIVERABLES, "_shared")
@@ -79,11 +81,13 @@ def _deliverable_roots():
 
 
 def _makefiles():
-    """Every Makefile fragment that can name an artifact path."""
-    paths = [os.path.join(REPO_ROOT, "Makefile"), os.path.join(REPO_ROOT, "paths.mk")]
-    paths += sorted(glob.glob(os.path.join(REPO_ROOT, "scripts", "analysis", "*.mk")))
-    paths += sorted(glob.glob(os.path.join(DELIVERABLES, "*", "*.mk")))
-    return [p for p in paths if os.path.isfile(p)]
+    """Every Makefile fragment that can name an artifact path.
+
+    Shared discovery (ticket 0248): a `.mk` relocation updates one list, and this
+    guard cannot silently narrow its universe — which would turn a real orphan
+    into a false green.
+    """
+    return [str(p) for p in all_makefiles() if p.is_file()]
 
 
 def _read(path):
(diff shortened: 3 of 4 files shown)


MinhHaDuong (commented):

**Verdict: COMMENT.** The guard is correct at HEAD, the `paths.mk` split is verified safe, and both gates are green in this worktree:

- `make lint` → **186 passed, 6 skipped**
- fast tier → **1217 passed, 11 skipped**
- `tests/test_artifact_reachability.py` → 6 passed in 0.91s

Nothing blocks the merge. What follows is one dissent, two silent-pass gaps in a guard whose whole premise is catching silent orphaning, and three cheap corrections.

---

### Dissent preserved

The **scope** reviewer returned `request-changes`; correctness, consistency, red team, and doc propagation all returned `comment`. Scope's objection is not about code — it is that the PR body's evidence describes a pre-review state (findings 4 and 6 below) and that 15 allowlist entries record decisions in a Python dict with no covering ticket (finding 5). The author decides whether that clears the bar.

---

### Findings

**1. `verifiable:` — the make-expansion half of the universe is blind to all six `deliverables/*/*.mk` render fragments** (correctness, high; red team, medium — same class)

`_assigned_variable_names()` harvests variable names from all 15 files `all_makefiles()` returns, but `_make_expanded_artifacts()` invokes `make -f Makefile -f probe.mk`, and the main Makefile `-include`s only 8 of them (`paths.mk` + 7 `scripts/analysis/*.mk`, `Makefile:29,126-132`). The 6 render fragments are deliberately excluded (`Makefile:122-125`, ticket 0237). Reproduced:

```
$ grep -n GIDE_VENUES_FR deliverables/manuscript/manuscript.mk
60:GIDE_VENUES_FR := deliverables/_shared/tables/tab_venues_fr.md
$ make -s -f Makefile -f /tmp/probe1198.mk __q
GIDE_VENUES_FR=[]
```

This is the same blind-spot class that commit `a1e24352` was written to close, relocated one directory over. It does not bite today only because no render `.mk` uses `$(foreach)`/`$(patsubst)`/`$(wildcard)` — every artifact path in those six files is literal text (`fig_bars_v1.png`, `fig_composition.png`, `fig_breaks.png`, `tab_venues_fr.md`, …), so the text-scan half of `_declared()` still catches them. The first render fragment that composes a list the way `zoo-figures.mk` already does will vanish from the universe silently.

Related, same class: `test_variable_composed_figures_are_in_the_universe` pins only the four `BIAS_FIGS` entries, though the docstring names eight composed figures. Red team disconnected `divergence.mk`'s `-include` in an isolated copy — `returncode 0`, universe drops 85 → 80 lines, `SENS_FIG_PCA`/`SENS_FIG_JL` gone, canary still green.

Suggested fix, cheaper than either: assert the make expansion is non-empty and that every variable name harvested from a file the probe *did not load* is absent from the universe — or simply add the four sensitivity figures to the canary and pin the loaded-makefile set.

**2. `verifiable:` — a commented-out reference counts as a live consumer** (red team, high)

The prose and Makefile scans run against raw text; comments are never stripped. Reproduced:

```
IMAGE_RE   on '<!-- ![old](../_shared/figures/fig_zzz.png) -->'   -> ['../_shared/figures/fig_zzz.png']
INCLUDE_RE on '<!-- {{< include ../_shared/_includes/gone.md >}} -->' -> ['../_shared/_includes/gone.md']
MK_FIGURE_RE on '# deliverables/_shared/figures/fig_dead.png: dep'   -> ['deliverables/_shared/figures/fig_dead.png']
```

0332 orphaned `fig_sem_composition.png` by *deleting* a paragraph. Had it instead *commented it out* — a routine drafting move, and this corpus's `_includes/` already carries `<!-- WARNING: AI-generated -->` / `<!-- STALE -->` markers — the guard would report the figure displayed and stay green. The Makefile direction is the milder mirror: a commented-out rule creates a phantom orphan (loud, self-correcting). No live instance exists in the corpus today (grepped clean); this is a landmine, not a current failure.

**3. `verifiable:` — the module docstring's cost claim is false, and the marker diverges from the two siblings it cites** (consistency, high)

`tests/test_artifact_reachability.py:63-67` says "Cost is fast-tier — a pure-Python lexical scan, **no subprocess**, no heavy import". `_make_expanded_artifacts()` (`:192-220`) spawns `make` with `timeout=120`, and it is reached by 4 of the 6 tests. Project rules classify subprocess-spawning as `integration`; both cited siblings dual-mark:

```
tests/test_handoff_artifacts_tracked.py:45  @pytest.mark.adherence
tests/test_handoff_artifacts_tracked.py:46  @pytest.mark.integration
tests/test_phase_layout.py:22  pytestmark = pytest.mark.adherence
tests/test_phase_layout.py:90  @pytest.mark.integration
```

The new file is module-level `adherence` only. Functionally harmless — `adherence` already excludes it from `check-fast`, and the file runs in ~0.4 s — but the shipped sentence is untrue. Either add `@pytest.mark.integration` to the four subprocess tests or correct the docstring.

**4. `verifiable:` — PR body numbers describe the pre-review state** (scope + doc propagation, high)

- Body: `make lint — 185 passed, 6 skipped ... +5 from this guard`. Actual, re-run at HEAD: **186 passed, 6 skipped**; the file defines **6** test functions.
- Body and ticket exit criteria: "**32** orphans ... 12 declared figures". `ALLOWED` at HEAD has **41** entries (17 includes + 3 tables + 21 figures). 32 was accurate at green commit `7952768f`; commit `a1e24352` surfaced 9 more.
- The composed-variable fix — arguably the PR's most valuable correctness change — appears nowhere in the PR body, only in that commit's message.

Correctness independently confirmed the red/green claim is genuine: `ALLOWED` emptied at HEAD yields exactly 41 orphans, equal to the allowlist keys — no gaps, no redundancy.

**5. `consider:` — 15 allowlist entries say in their own text that no ticket covers them** (scope, medium)

6 `_UNSETTLED_REASON` + 9 `_ZOO_DIAGNOSTIC_REASON` entries carry "No ticket covers these yet ... this entry is the record until one is filed" and "untracked, like the entries above". This is a science repo, so the tooling severity floor does not apply. Recommend one tracking ticket enumerating them, mirroring how 0290 owns the 17 flat includes — otherwise the record lives only in a Python dict.

**6. `consider:` — ticket 0290's action 3 shipped here, but 0290 is untouched** (scope, medium)

0290 is open (no `Closed:` header, not under `tickets/closed/`) and its action 3 asks for exactly this guard. The PR body says so in prose, but there is no `Ticket-ref: tickets/0290-...` line and 0290's own log has no entry. A reader opening 0290 alone will not know action 3 landed.

**7. `consider:` — `figures-corpusreport` breaks the alignment of its five siblings** (consistency, low)

`Makefile:627-632`: the other five pad `corpus-handoff` to a common column; the new line uses a single space.

**8. `nofollow:` — the docstring re-tells `.claude/rules/architecture.md`'s trap narrative near-verbatim** (consistency, low). It cites the doc, so this is drift risk, not undocumented drift.

---

### Attacked and found clean

Red team's highest-value vector — whether moving `fig_dag` out of `DATAPAPER_FIGS` drops it from the Zenodo archive — **found nothing**: `build/build_datapaper_archive.sh:75` hardcodes its `cp` and never referenced `$(DATAPAPER_FIGS)`, so `fig_dag` was never in the archive before or after. `ALL_FIGS` gains `$(CORPUSREPORT_FIGS)` in the same commit; expansion verified at 35 paths, `fig_dag.png` once.

Also verified clean: no `$(shell …)` anywhere in the loaded Makefile tree (no network/corpus dependency, no timeout flakiness); the `_reachable_includes()` walk is cycle-safe and order-invariant (3 identical runs); no symlinked directories under `deliverables/`; no basename collisions across the 92 declared artifacts; `REPO_ROOT` derives from `__file__` so the suite passes when invoked from `/tmp`; the allowlist's `elif` dead branch cannot mask anything (`_includes/` and `figures/`+`tables/` are disjoint by construction, no Makefile rule writes under `_includes/`); `ruff` clean; `erg check tickets/` PASS (352 tickets).

Doc propagation checked **every** file:line citation in the new comments and docstrings — `test_global_map.py:175`, `Makefile:136`, `build_datapaper_archive.sh:47`, `manuscript.mk:62-63`, `corpus-report.qmd:25`, `r1-14-network-response.md`, `docs/ncc-pipeline-audit.md`, `zoo-figures.mk:139`, `divergence.mk:129-130`, the 11 roots, the 18 zoo includes — all correct, no rot. Tickets 0226/0240/0286/0290/0307/0332/0359 all exist and say what is claimed (0332's cut confirmed independently at commit `fd1e1094`). No doc, README, template Makefile, dvc.yaml, or archive manifest needs updating for `CORPUSREPORT_FIGS`/`figures-corpusreport`.

---

### Recommended before merge (all cheap, none blocking)

1. Fix the docstring's "no subprocess" sentence, or add `@pytest.mark.integration` (finding 3).
2. Correct the PR body's `185`/`+5`/`32` to `186`/`+6`/`41` (finding 4).
3. Add `Ticket-ref: tickets/0290-...` and a one-line log entry on 0290 (finding 6).

Findings 1 and 2 are real silent-pass gaps and deserve a follow-up ticket rather than an in-PR scramble — the guard is a clear net positive as it stands.

🤖 `/review-pr` — read-only panel, no fixes applied, no gate.

PR #1237, opened 2026-07-28
fix(0420): guard the whole render-placeholder class; fix the crossref it found

MinhHaDuong:
Extends the 0363 render oracle to the rest of the unresolved-reference class — crossrefs and citations — and fixes the live crossref defect the ticket was filed on.

## ⚖️ One prose addition awaits your arbitration

Beyond the mechanical three-token reference fix, this PR **adds one sentence to the fig-zseries caption**: *"Shaded ribbons around the method curves span the trimmed subsampling range — the central sixteen of twenty equal-$n$ draws (§4.5)."* Rationale: three prose passages send the reader to the figure for the subsampling ribbon, and the caption never mentioned it; the review panel verified the sentence numerically against `config/analysis.yaml` and the plot code. It is an autonomous manuscript-prose change, so it lands only if you keep it — say the word and I drop it.

## The ticket's premise was wrong, in a useful way

Ticket 0420 said `@fig-companion-zseries` is referenced three times and "the figure is defined nowhere — it left with the retired companion paper." The figure is at `multilayer-detection.qmd:205`, labelled `{#fig-zseries}` on the image `fig_companion_zseries.png`; line 175 already references it correctly. The three broken references restyle the image *filename* as a label. The "author call" the ticket anticipated dissolves into a mechanical fix, applied here. (Checked first that the prose promise is true: `plot_companion_zseries.py` genuinely draws the ribbon — `fill_between` over `z_trim_lo`/`z_trim_hi`.)

## The guard, extended by measurement

The signal per mechanism was probed on a real render — and two assumptions died on contact:

- **Crossrefs in body text** write `?@fig-name` to stdout; they join `?meta:` in one `placeholders_in()` scan.
- **Crossrefs in figure/table captions** write *nothing* to stdout — Quarto's markdown writer drops caption text wholesale (review round 1's blocker, reproduced before fixing). The guard unions the output scan with a stderr scan for `Unable to resolve crossref`; a red-first caption probe asserts both halves — stderr sees it, stdout does not — so a future "simplification" back to output-only fails the probe instead of reopening the hole. The instance the review named (`tab_traditions.md:7`) is orphaned and allowlisted pending 0290; the class is closed regardless.
- **Citations** write nothing to markdown either — citeproc renders `(key?)` into HTML/PDF but leaves `@key` untouched in markdown by design, so the `**@key?**` literal the ticket predicted is emitted by no writer, and a guard coded to it would have passed forever. The signal is stderr: `Citeproc: citation X not found`. And citeproc only runs where a bibliography is declared, so a new fast-tier check requires any citing deliverable to declare one (slides-gide is the only document with neither, passing vacuously).

All mechanisms share one render per document. Everything is red-first, on the defect's real shape (label mismatch, caption placement), with exit 0 asserted on every probe — the silence is the defect class.

The suite is renamed `test_render_placeholders.py`; both references to the old name are updated, plus ticket 0420's own Relevant-files entry (logged).

## Review trail

- **Round 1** (request-changes): caption-crossref blind spot — fixed as above; dead `PLACEHOLDER_RE` removed; README broadened; caption sentence added.
- **Round 2** (no blockers): pattern composed from the `PLACEHOLDER` literal; stdout/stderr crossref character classes harmonized; warning-count claim dropped; bibliography check widened to `declared_keys()` (header ∪ metadata-files). The panel itself filed ticket 0511 / PR #1251 for the pre-existing fresh-worktree render-skip gap — 0363's known coverage boundary, not widened here.

## Gates

- `make lint` — 279 passed. `make check-fast` — 1333 passed.
- `make check` — 2069 passed, 8 failed: the same 8 corpus-absent-in-worktree failures as the 0363 baseline (`.dvc` pointers only).
- `/verify-adherence` — PASS, no findings (noted for `/gaze` to skip the mechanical phase).

One environmental incident during gating, fixed: this worktree's fresh `.venv` (the shared `/data/envs/venv/oeconomia` target does not exist on padme, so the post-checkout hook falls back to a local env) had a corrupt dcor install — `import dcor` died with `LLVM ERROR: Symbol not found`, crashing 20 numerical tests. `uv sync --reinstall-package dcor` cured it. Not a code issue on any branch.

**Ticket:** tickets/0420-unresolved-crossrefs-render-as-ref-with.erg


MinhHaDuong (commented):
## /review-pr — five-perspective panel (correctness, consistency, scope, red team, doc propagation)

> **Verdict: REQUEST-CHANGES** (posted as comment — the forge refuses a request-changes review from the PR author's own account).

Panel verdicts: correctness **approve**, consistency **comment**, scope **comment**, doc propagation **comment**, red team **request-changes**. The red-team blocker was independently reproduced by the synthesizer before posting; it stands. Overall: **request-changes**, with a small, well-bounded fix.

### Blocker (red team, reproduced independently)

**A broken crossref inside a figure/table caption is invisible to the new guard.** Quarto's markdown writer does not support FloatRefTarget nodes, so it drops caption text entirely — the `?@label` placeholder never reaches stdout, and `placeholders_in()` sees nothing. The crossref filter still warns on stderr (`WARNING … Unable to resolve crossref @fig-ghost`), but nothing reads it: `WARNING` in this suite is the string `"Unknown meta key"`, and `missing_citations_in()` matches only the citeproc line. Reproduction (quarto 1.8.27):

```
$ cat capxref.qmd   # caption contains @fig-ghost; table caption contains @tbl-ghost
$ quarto render capxref.qmd --to markdown --no-execute
exit 0; output: caption text dropped, no ?@ anywhere in stdout
stderr: WARNING … Unable to resolve crossref @fig-ghost
        WARNING … Unable to resolve crossref @tbl-ghost
```

This is not theoretical: `deliverables/_shared/_includes/tab_traditions.md:7` is a table caption carrying `@fig-traditions` — exactly the shape (label drift under this class of defect) that ticket 0420 was filed on. As shipped, the exit criterion "an unresolved crossref cannot reach a rendered deliverable without a test failing" holds only for body-text crossrefs.

**Fix is one regex.** The stderr warning fires for body-text crossrefs too (verified: a broken body crossref emits both `?@fig-nowhere` in stdout *and* the stderr warning), so a stderr scan is a complete superset of the stdout check. Mirror the citeproc pattern: `re.compile(r"Unable to resolve crossref @?(\S+)")` in `_qmd_meta.py`, OR its hits into `unexpected` in `test_rendered_deliverable_has_no_placeholder`, and add a red-first probe whose broken crossref lives *inside a caption* (the stdout regex passes on it; only the stderr scan goes red). Keep the stdout `?@` pattern as defense in depth or drop it — either is defensible if the docstring says which signal is load-bearing.

### Minor findings

- `verifiable:` **Dead code: `PLACEHOLDER_RE` at `tests/test_render_placeholders.py:77`** (found by 4 of 5 perspectives). Its only call site was replaced by `placeholders_in(...)` in this same commit; `grep -n PLACEHOLDER_RE tests/` returns exactly its own definition. Ruff cannot see an unused module-level name. Delete the line while fixing the blocker. (correctness/consistency/scope/doc-prop, high confidence)
- `consider:` **Citation blind spot where no `bibliography:` is declared.** `slides-gide.qmd` is the one deliverable without a bibliography key; a stray `@key` there runs no citeproc, emits no stderr line, and the guard sees nothing. Currently unexploited. A one-line docstring note (or a check that documents with `@`-citations declare a bibliography) would close it cheaply. (red team, medium confidence)
- `consider:` **README ~line 77 still frames the suite around the `?meta:` mechanism only** — accurate but incomplete now that the renamed suite guards three mechanisms. One clause would fix it. (doc propagation, medium confidence)
- `consider:` **Ticket 0420 body line 42 ("Relevant files") still names `test_meta_macro_resolution.py`.** Moot at merge when the ticket closes; cheap to touch in the same commit if you're editing anyway. (consistency + doc propagation, high confidence)
- `consider:` **Figure caption at `multilayer-detection.qmd:205` never mentions the subsampling ribbon** that the three fixed sentences now direct readers to — it describes only the grey |Z| < 2 band. The ribbon is real (`plot_companion_zseries.py` draws `fill_between(z_trim_lo, z_trim_hi)`), so the prose is accurate, but the caption gives no cue. Out of 0420's scope; candidate follow-up ticket. (scope, medium confidence)
- `consider:` **Dissent preserved (scope):** ticket action 1 named the prose resolution "an author call"; the PR resolved it unilaterally as a mechanical three-token label fix. The panel verified the justification independently — figure exists at line 205 as `{#fig-zseries}`, the ribbon is genuinely drawn and wired through `divergence.mk` for the four methods the prose names — so the call looks right, but the author arbitrates at merge per project convention.
- `nofollow:` Crossref regex omits dots while the meta-key regex allows them — checked against every label in the repo (all hyphen-only) and against Quarto label syntax; intentional and correct. `(\S+)` trailing-punctuation over-capture in the stderr patterns is theoretical; no instance found.

### What the panel verified positively

- The three prose fixes are substantively correct, not just label-matching: `@fig-zseries` exists, and the generating script draws exactly the subsampling ribbon the sentences describe.
- Red-first discipline holds for both new mechanisms; the citation probe's stderr-is-the-signal reasoning was independently confirmed by a hand render (`[WARNING] Citeproc: citation nosuchkey2099 not found`).
- Style, tier markers, `#:` conventions, and cross-file docstrings are consistent; the rename is justified and both known references were updated; no drive-by changes (diff minus the five files is empty).
- Gates in the PR worktree: `make lint` 279 passed; `make check-fast` exit 0; target suites 27 fast + 13 integration passed, with `multilayer-detection` rendering clean.

**Requested change, concretely:** add the `Unable to resolve crossref` stderr scan + a caption-shaped red probe, and delete the dead `PLACEHOLDER_RE`. Everything else is tagged non-blocking.

MinhHaDuong (commented):
## /review-pr — five-perspective panel (correctness, consistency, scope, red team, doc propagation)

**Panel verdicts:** correctness **approve**, doc-propagation **approve**, consistency **comment**, scope **comment**, red team **comment**. No blocker. Local gates on the PR head (837b2510): `make lint` 279 passed; fast tier 1333 passed, 11 skipped; integration render tests 14 passed, 2 corpus-absent skips (run independently by two panel agents against Quarto 1.8.27, including all three red-first probes and the real `multilayer-detection` case, which passes).

**Verified core:** all four `@fig-zseries` references resolve against exactly one `{#fig-zseries}` label; the sibling techrep document and its include trees are clean; the rename left zero stale references (the two remaining `test_meta_macro_resolution` hits are an append-only log line in 0357 and 0420's own explanatory parenthetical); `erg check` passes (386 tickets); the union logic cannot let a crossref hide behind a `KNOWN_UNRESOLVED` meta-key entry (only meta keys are pinned there).

### For the author's arbitration (scope)

- `consider:` **New manuscript prose rode in with the mechanical fix.** The `fig-zseries` caption (`multilayer-detection.qmd:205`) gains a sentence — "Shaded ribbons around the method curves span the trimmed subsampling range — the central sixteen of twenty equal-$n$ draws (§4.5)." The scope reviewer verified it numerically (`compute_crossyear_zscore.py` + `subsample_trim: 2` in `config/analysis.yaml`: 20 − 2×2 = 16), so the content is accurate, but the PR body folds it into "pointing at the wrong name" rather than flagging it as an addition. Explicit sign-off requested here rather than a split.

### Confirmed findings (red team, evidence attached)

- `verifiable:` **The render oracle is green-by-absence on the number-bearing documents in every fresh worktree.** Probe: injected `@fig-totally-does-not-exist-xyz` into `data-paper.qmd` in this worktree → `test_rendered_deliverable_has_no_placeholder[data-paper]` **SKIPPED**, exit 0 (generated includes unbuilt; `make check` never builds `corpus-tables`; no CI exists to run it elsewhere). Pre-existing from 0363's design — not this PR's defect — but 0420 widens the same blind spot to crossrefs and citations on the same two documents. Filed as **ticket 0511** (PR #1251) per the pre-existing-but-touched rule; not blocking here.
- `consider:` **Narrative-only citations bypass the bibliography guard.** `test_citing_deliverable_declares_a_bibliography` keys on `[@`; a document citing only as bare `@key` with no `bibliography:` gets no citeproc run, no stderr, no placeholder (probe-confirmed). The docstring admits this and the invariant holds today (every citing deliverable has a bracketed cite), but nothing tests that invariant going forward. Dormant; noted for the record.
- Defended (probed, no action): stderr-wording drift is covered because the red-first probes run in the same tier against the same binary and would fail loudly; `render_to_markdown` passes no `--quiet`/log-level flag that could mute the warnings; the markdown oracle does not under-detect relative to HTML output (the caption `?@` is *visible* in HTML, and the stderr channel catches it identically in both writers).

### Minor (correctness / consistency)

- `verifiable:` **Crossref regex asymmetry** — `PLACEHOLDER_PATTERNS["crossref"]` (`_qmd_meta.py:44`) excludes `.` while `CROSSREF_MISSING_RE` (`:62`) allows it. Probe with a dotted label: stdout match truncates (`fig-companion`), stderr match is full (`fig-companion.zseries`). The union keeps the assertion failing, so no false pass — but the error message would show two different names for one defect. One-character fix; no dotted label exists today.
- `verifiable:` **Stale count in a doc comment** — `_qmd_meta.py:58–61` says the caption probe yields "two of these warnings on stderr"; reproducing the exact fixture on Quarto 1.8.27 yields **one**. The test asserts set equality, not count, so it's prose-only — but the comment is wrong as written.
- `consider:` **DRY regression** — `PLACEHOLDER = "?meta:"` (`_qmd_meta.py:34`) and the hand-written `r"\?meta:…"` in `PLACEHOLDER_PATTERNS` (`:43`) are now two independent sources of the same literal; the removed `PLACEHOLDER_RE` composed from the constant. Suggest `re.escape(PLACEHOLDER) + r"([A-Za-z0-9_.-]+)"`.
- `consider:` **Hypothetical false-fail** — `front_matter()` reads only the top-level YAML, so a `bibliography:` declared via `metadata-files:` would trip the new guard. No such document exists; noted so the failure message isn't a mystery if one ever does.
- `consider:` **Ticket body edited, not logged** — 0420's Relevant-files bullet was rewritten to note the rename rather than adding a log line. Body edits are legal in %erg 0.1 (only the log is append-only) and `erg check` passes, but a log entry would have kept the audit trail self-contained.
- `nofollow:` single blank line before the `QUARTO_REFUSES` block (`_qmd_meta.py:88`) where the file elsewhere uses two — stylistic, not ruff-flagged in this repo's rule set.

**Dissent:** none — the three "comment" verdicts are additive to the two "approve"s, not contradictory.

**Synthesis:** the guard is sound, honestly documents its own blind spots, and the red-first probes structurally cover the stderr-drift evasion. Nothing blocks the merge; the one item genuinely awaiting the author is the caption-sentence sign-off above.

PR #1238, opened 2026-07-28
fix(0402): stop inventing a publication year for undated works

MinhHaDuong:
Closes ticket 0402, filed from the 0354 wrap-up sweep for the silent-coercion class.

## The defect

`build_het_core.py` dated every yearless work 2020. Not a filtering convenience — the value fed `cit_per_year`, hence `score`, hence which works `_select_with_quotas` picks. An old undated work had its citation rate divided by ~6 years instead of its true age, inflating it several-fold on a number nobody measured.

**Not latent.** Against the pre-fix output on the current corpus, **43 of the 1000 selected works swap**.

## Correction to the ticket's own severity claim

The ticket asserted this reached a *published* table via `het_mostcited_50.csv` → `summarize_core_venues.py` → `tab_core_venues_top10.md`. The chain is real, but the endpoint is published nowhere: no `.qmd` includes that table, it is gitignored (blanket `deliverables/_shared/tables/*`, no negation), and `compute_vars.py` never reads the ranking. I overstated it at filing and corrected the ticket body rather than leaving it to be believed.

So no paper number changes. What survives is a correctness defect in a scoring pipeline whose output is an analysis intermediate. For the record, the (unrendered) venues table moves exactly one cell — OECD publications 59 → 57 — with ordering and venue membership unchanged.

## Policy: no age credit, not exclusion, not imputation

An undated work forfeits the age-normalised term and keeps the other two signals (citation-graph centrality, teaching-canon presence), neither of which depends on the year. So it stays eligible on the evidence that exists.

Two measurements say this is the safe choice:

- **No grey-literature, UNFCCC or OECD work is undated** — 0 of 210, 230, and 33 respectively. Excluding undated works would therefore *not* have biased against the corpus's grey-lit layer, but it would have discarded 389 works, 214 of them teaching-canon readings.
- **Undated works carry at most 8 citations**, so the forfeited term is small in every case.

Imputing a median age would be the same fabrication with a better excuse.

## Gap versus fault

`errors="raise"` on both coercions, splitting the two kinds of non-value: a blank year is a gap and becomes NA; `"n.d."` is a pipeline fault and stops the build. This is the ticket-0354 lesson applied one layer over — there, a coerced blank passed a nullable schema cleanly.

An absent citation count still reads as zero, now with the reason stated in place: a work OpenAlex never saw cited has nought citations, which is a measurement rather than a guess.

## Structure and the regression guard

The two coercions became `parse_years` / `citations_per_year`, so the undated-work policy is testable and lives in one place. `tests/test_het_core_ranking.py` covers the arithmetic and, separately, the **call sites** — an adherence guard rejecting a literal year inside a `fillna` and pinning `main()` to the helpers. That split is deliberate: what regressed here was a call site, not the arithmetic, and the guard was genuinely red on `fillna(2020)` with a message naming the consequence.

## Verification

- Red → green demonstrated (call-site guard red on the literal year).
- `make lint` 309 passed. Full suite **2158 passed, 47 skipped, 0 failed**.
- Old-vs-new artifacts compared on the same corpus, per the byte-check discipline.

Ticket: tickets/0402-build-het-core-fabricates-a-2020-year-fo.erg


MinhHaDuong (commented):

**Recommended verdict: request-changes on two cheap, introduced-by-this-PR items.** The scoring fix itself is sound: `parse_years`/`citations_per_year` are dtype-robust, the arithmetic is right, and the panel found no defect in the ranking logic. What should not merge as-is are (1) a retracted claim shipped as fact in the new test file and (2) a regression guard that does not hold.

Locally verified on the PR head (`9c13a960`): `make lint` → **319 passed, 12 skipped**; `make check-fast` → **1330 passed, 8 skipped**; `tests/test_het_core_ranking.py` → 8 passed.

---

### 1. The new test module ships the claim this PR spent a revision retracting — *doc propagation + consistency, high confidence*

`tests/test_het_core_ranking.py:3-6`:

> `build_het_core.py` scores the corpus and writes `het_mostcited_50.csv`, which `summarize_core_venues.py` and `export_core_venues_markdown.py` carry into the **published** `tab_core_venues_top10.md`. So the age term in that score reaches a table in the paper.

The PR body and the ticket both retract exactly this. I re-checked every leg of the retraction independently, and the retraction is correct:

- `grep -rn tab_core_venues_top10 --include=*.qmd --include=*.tex --include=*.md` → one hit, `docs/plan-makefile-modularization.md:83`, a build-plan doc. No `.qmd` includes the table.
- `.gitignore:44` is `deliverables/_shared/tables/*` and `grep -n core_venues .gitignore` → 0 matches, so there is no negation whitelisting it (unlike ~19 sibling tables).
- `scripts/analysis/compute_vars.py` → 0 references to `het_mostcited`, `core_venues`, or `het_core`.

So the docstring is a false statement, in a file that outlives this PR discussion, contradicted by the ticket's own dated correction (`2026-07-27T22:40Z ... Corrected this ticket's own severity claim`). The same claim is repeated at **`tests/test_het_core_ranking.py:96`**, inside the assertion message a future developer sees on a real bounce: *"hence which works reach the published venues table"*. Both need the correction the ticket already wrote.

This is the doc-propagation defect class the project already has memory on — stale prose reaching the durable record. Two-sentence fix.

### 2. `verifiable:` The adherence guard does not hold — bypass reproduced live

`TestNoYearImputationSurvives` is the PR's stated regression protection ("what regressed here was a call site, not the arithmetic"). It is a lexical grep and the defect walks past it. I patched `main()` to reintroduce the *exact* fabrication:

```python
df["year_num"] = parse_years(df["year"]).fillna(2020.0)
```

```
$ uv run python -m pytest tests/test_het_core_ranking.py -q
8 passed in 0.52s
```

(Patch reverted; `git status --porcelain` clean.) The regex `fillna\(\s*(1[89]\d\d|2[01]\d\d)\s*\)` requires exactly four bare digits between the parens:

```
'x.fillna(2020)'                      -> ['2020']         caught
'x.fillna(2020.0)'                    -> ESCAPES GUARD
'x.fillna(value=2020)'                -> ESCAPES GUARD
'x.fillna(CURRENT_YEAR)'              -> ESCAPES GUARD
'x = np.where(x.isna(), 2020, x)'     -> ESCAPES GUARD
```

`test_main_scores_through_the_helper` does not close the gap either — it only asserts the strings `"citations_per_year("` and `"parse_years("` appear *somewhere* in the file, not that nothing intervenes. Two reviewers found this independently. The guard is the deliverable here; a guard that passes the reintroduced defect verbatim is worth one more iteration.

---

### Comment tier

**3. `verifiable:` The teaching-canon guarantee is dead code (pre-existing, not introduced here).** `build_het_core.py:317` selects `s2[s2["teaching_count"] >= 2]`, commented "guaranteed inclusion for teaching canon papers (2+ institutions)". But `teaching_count` is assigned from `from_teaching` at line 450, which is strictly binary on the real corpus (`{'0': 32726, '1': 618}`), so `>= 2` is unsatisfiable. Confirmed by the script's own log on a real run: `Teaching canon guaranteed: 0`. Per the escalation table this is pre-existing-and-untouched, so a ticket rather than a change here — flagging that the author capped filing this session, so this may want to be recorded in memory instead.

**4. `nofollow:` One panel finding is refuted — recording it so it is not re-raised.** The red team concluded, at high confidence, that the PR's own eligibility rationale is *"empirically false on the real corpus"*: that 353 undated works reach S2 and **"0 of the 353 appear in the final selected 1000"**, making the policy *"indistinguishable from the 'drop them' alternative the ticket explicitly rejected"*, and recommending the docstring be rewritten to say undated works are de facto excluded. I re-ran the pipeline and measured the output directly:

```
Corpus: 33344 papers (389 undated — no citations-per-year credit)
Selected: 1000
selected with blank year: 30      # scores 0.1341 – 0.2676, vs a selection floor of 0.0423
```

**30 undated works are selected**, carried in on pagerank with no age credit and no citation count — which is precisely the behaviour the `citations_per_year` docstring claims. The rationale holds and the docstring is accurate on this point. Dissent preserved verbatim above for the author; the measurement is the answer to it.

**5. `consider:` The "gap versus fault" split may not be reachable at this layer.** The docstring justifies `errors="raise"` by distinguishing a blank year (gap → NA) from `"n.d."` (fault → stop the build). But the `year` column in `refined_works.csv` arrives already float-formatted — `{'2024.0': 4229, '2023.0': 3908, ...}` — evidence of an upstream numeric round-trip, and there are 0 non-blank non-numeric values in 33,344 rows. A genuine `"n.d."` would have been flattened to blank one stage earlier, so the fault branch likely cannot fire. That cuts both ways and resolves the panel's dissent on this point: the red team's warning that `errors="raise"` turns one bad grey-lit row into a full `make all` abort is **not a live risk** (verified: 0 offenders today), and the PR's fault-detection claim is correspondingly weaker than stated. No change requested; worth a sentence in the docstring rather than a policy change.

**6. `consider:` The zero-rate policy enlarges the tie mass at the selection boundary.** All 389 undated works now share `cit_per_year == 0.0` regardless of citation count, where before they had distinct (if fabricated) rates. On the real run, 15 of the selected 1000 share the minimum score 0.0423, and `_select_with_quotas` iterates `s2` in whatever order a non-stable `sort_values("score")` leaves ties in — so membership at the margin can move across rebuilds on raw row order alone. Pre-existing mechanism, measurably widened here. A `kind="mergesort"` plus a documented tiebreaker would make the selection reproducible; not blocking.

**7. `nofollow:` The headline numbers are unverifiable from the repo.** "43 of 1000 selected works swap" and "OECD publications 59 → 57" are the evidence behind two ticked exit criteria, but `het_mostcited_50.csv` is neither git- nor DVC-tracked, and no comparison script or log was committed. I regenerated the output (1000 rows, 30 undated) but cannot reproduce the *delta* without the pre-fix artifact. Not a blocker — the byte-check discipline was clearly followed — but the numbers rest on trust rather than an artifact.

**8. `nofollow:` Scope — the 0403 fix rides along, correctly explained.** Commit `9c13a960` (one line, `return found` in `tests/test_datapaper_archive_layout.py`) belongs to open ticket 0403, not 0402. The underlying claim checks out: `d7d9789f` on main dropped the return and main is genuinely red without it. One panel correction: the review flagged the commit's parenthetical *"(0402 is closed by the commit below; this repairs 0403's guard)"* as false — it is accurate, `86e18250` in this PR does close 0402. The only nit is that this should have been `Ticket-ref: tickets/0403-...` rather than `Ticket: none`. Separately, ticket 0403 has every box ticked and "Guard landed" in its log but still sits unarchived in `tickets/` — pre-existing bookkeeping drift.

**9. `nofollow:` Docstring voice is an outlier in the file.** `parse_years`/`citations_per_year` are 6–13-line narrative arguments with ticket numbers and rejected alternatives; every sibling (`field_score`, `is_institutional_report`, `robust_minmax`) is 1–3 factual lines. Arguably decision-record content that belongs in the ticket. Stylistic preference, recorded not pressed.

*Checked and clean:* helper naming fits the file's conventions; `CURRENT_YEAR = 2026` is pre-existing (line 42); the `export_deposit.coerce_integer_columns` precedent cited at line 203 is accurate; the `@pytest.mark.adherence` tier and hand-rolled `_source()` read match `test_io_discipline.py` / `test_script_hygiene.py` / `test_arch_compliance.py`, so no shared helper is being reimplemented; `year_num` does not leak downstream (the output writes the raw `year` string, and other files' `year_num` are independent locals); no tracked derived artifact is left stale; no manuscript prose describes this ranking (the manuscript's "core subset" is a separate `cited_by_count >= 50` path at `compute_vars.py:266-271`).

PR #1239, opened 2026-07-28
ticket(0371→0450): renumber clear of the frontier after a collision landed on main

MinhHaDuong:
`origin/main` carried two tickets numbered 0371. PR #1235 opened with this ticket at 0371 and merged 267 commits later; in that window a sibling filed `tickets/0371-included-shared-tables-are-gitignored.erg`.

Neither check could see it — branch-local `erg check` passes by construction, the cross-PR gate compares open PRs only, and this repo runs no CI.

Renumbered to **0450**, fifteen clear of a frontier at 0435 with sixteen more IDs in flight across open PRs. Not 0372: the next free ID is the most contended seat in the repo and chasing the frontier cannot converge while siblings are filing.

Cross-references updated in the archived 0365, which names this ticket three times.

Verification: `erg check tickets/` → PASS (384 tickets); no duplicate IDs across the tree; 0450 absent from `origin/main` and from every open PR.

**Ticket-ref:** tickets/0450-swap-agent-token-to-a-fine-grained-pat.erg

PR #1243, opened 2026-07-28
ticket(0470): close — fine-grained PAT installed and verified on both loader paths

MinhHaDuong:
Closes the browser-only half of 0365 decision 1. The author minted the fine-grained PAT; this records what was installed, how it was verified, and what deliberately stays open.

## The route changed, and the new route is better

The ticket planned: replace `AGENT_GH_TOKEN`'s value, change no `KEYS=` line. What happened: the new token went in under `AGENT_GH_TOKEN_CLIMATEFINANCE`, and the `KEYS=` entry became the rename-on-export form.

That is the right arrangement and the plan was wrong to rule it out. A fine-grained PAT is *per-repository* by construction, so the keystore will accumulate one github token per project and a bare `AGENT_GH_TOKEN` cannot name them all. The suffix names which repository the token is bounded to; `KEYS=` does the selection — exactly as the sibling `openrouter:…_CLIMATEFINANCE=…` entry already did. It also keeps the classic PAT reachable under its own name during the swap, which is what the ticket's own invariant asked for.

## Verification

All by behaviour under `env -i`, never by reading a value:

- identity is `MinhHaDuong`; this repo plus its PR, issue and label endpoints read
- three **private** repos 404 → scope is bounded, not account-wide
- create-then-delete of `refs/heads/probe-token-scope` → write access confirmed, so the swap cannot lock the agent out at the next push
- both application paths export it — the `BASH_ENV` loader (`make`) and `pipeline_loaders` → `pipeline_keystore` (`dvc repro`, bare `uv run`)
- the repaired 0365 merge gate read the new token live when it allowed PR #1239

One check the ticket specified would have misled: `gh api user/repos --jq length`. A fine-grained PAT reads every **public** repository implicitly, so a public repo resolving proves nothing. Only a private repo discriminates.

## Left open deliberately

The classic PAT stays in the keystore, unreferenced here — other projects may select `github:AGENT_GH_TOKEN` and revoking it would break them silently. Revocation is the author's call. Also recorded: fine-grained PATs expire within 366 days, so this is a scheduled outage whose symptom is a 404 that reads like a permissions bug.

## Incidental finding, not fixed here

`make lint` shows `origin/main` red: 4 failures in `test_datapaper_archive_layout`, reproducible on a branch with zero commits. Three are 0403's `test_document_names_only_shipped_products`; the fourth is that guard's own self-test failing with `TypeError: argument of type 'NoneType' is not iterable`, pointing at `d7d9789f` rather than at the prose. Logged on open ticket 0403. This repo has no CI, so nothing else would have reported it.

**Ticket:** tickets/closed/0470-swap-agent-token-to-a-fine-grained-pat.erg
**Ticket-ref:** tickets/0403-standing-guard-a-submission-document-mus.erg


MinhHaDuong (commented):
Procedural review (0365 option 3b — this gate asserts the review pass ran, not that an independent party approved).

Diff is tickets-only: 0470 closed and archived, one log line on 0403. Checked:
- `erg check tickets/` → PASS (384 tickets)
- the archive commit is 131 insertions, not a pure rename, so the `Closed:` header landed (the trap that silently drops it)
- `Label: needs-human` removed — the human step is done
- 0470 free of collisions: absent from every open PR
- no credential value in the diff, the commit message, or this PR

Substantive verification is in the ticket body: identity, bounded scope via private-repo 404, reversible write probe, both loader paths.

PR #1248, opened 2026-07-28
fix: main is red — products_named_in lost its return statement in d7d9789f

MinhHaDuong:
**Ticket-ref:** tickets/0403-standing-guard-a-submission-document-mus.erg

## What was actually broken

`origin/main` was red on four failures in `tests/test_datapaper_archive_layout.py`, all in the ticket-0403 guard (`TestSubmissionProseNamesOnlyShippedProducts`). All four were **one bug in the guard, not defects in the prose**.

Commit d7d9789f ("Potential fix for pull request finding", an auto-suggested patch) restructured the token loop of `products_named_in` to expand ed04's split-extension shorthand — `` `tab_retrieval_protocol.csv`/`.md` `` — and dropped the `return found` line in the rewrite. The function returned `None`, which showed two faces:

- the three parametrized document checks (`response-letter.md`, `summary-of-revisions.md`, `ed04-zenodo-restructure-upload.md`) failed `assert named` and reported "names no deposit product at all";
- the guard's own self-test `test_a_retired_product_is_caught` died on `TypeError: argument of type 'NoneType' is not iterable` — the tell that the detection path could not run at all.

The three submission documents were verified against the fixed guard: their deposit enumerations parse and name only shipped products. No prose was touched.

## Why the fix is this shape

- **Restore `return found`** — one line; the d7d9789f branch logic itself is correct (it exists to catch ed04's `` `.md` `` shorthand, which would otherwise let `tab_retrieval_protocol.md` escape the subset check).
- **Add `test_a_split_extension_shorthand_is_expanded`** — the shorthand branch had no red-proof, so a wholesale revert of d7d9789f would have gone green while silently losing that detection. The new test pins both the expansion and the returned set.
- **Red-proof replayed mechanically**: with `return found` mutated away, 5 tests fail — the new one with a plain assertion (`assert None == {...}`), not just the self-test's TypeError. With the fix, the full file passes (23 tests).

`CLIMATE_FINANCE_DATA=/home/haduong/Climate_finance/data make lint`: **318 passed, 12 skipped** — clean.

Also appends a diagnosis line to ticket 0403's log (referenced, not closed — it is the standing-guard ticket, not this repair).

### DIFF
--- a/tests/test_datapaper_archive_layout.py
+++ b/tests/test_datapaper_archive_layout.py
@@ -109,6 +109,7 @@ def products_named_in(text):
                     last_stem = re.sub(_PRODUCT_EXT + r"$", "", matches[-1])
                 elif last_stem and re.fullmatch(_PRODUCT_EXT, token):
                     found.add(last_stem + token)
+    return found
 
 
 class TestRenderRuleTracksTheIncludeClosure:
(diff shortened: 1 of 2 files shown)


MinhHaDuong (commented):
Reviewed the diff against the d7d9789f regression. The root cause is exactly the missing return: the restructured loop in products_named_in never returned found, so every caller got None — three documents falsely reported as naming nothing, and the self-test crashed with a TypeError instead of asserting. The one-line restore is the smallest correct fix, and the new test_a_split_extension_shorthand_is_expanded is a genuine red-proof: verified mechanically that with the return mutated away it fails on a plain assertion, and that the previously untested shorthand branch (the reason d7d9789f existed) is now pinned so a revert cannot silently drop tab_retrieval_protocol.md detection. make lint green (318 passed). No prose changed; ticket 0403 referenced, not closed. Trivial in size, correct in shape.

PR #1249, opened 2026-07-28
Merge gate honours the ticket-filing fast path (tickets-only diffs exempt from review count)

MinhHaDuong:
**Ticket:** none

## The contradiction

Two written policies collided the moment the merge gate's matcher was repaired (ticket 0365):

- `.claude/hooks/check-reviews.sh` denies every `gh pr merge` without 2 review cycles (1 with `review:trivial`).
- `.claude/rules/git.md`, "Ticket-filing PRs take the fast path", says a tickets-only PR merges on `erg check` plus an ID-collision scan — no review ceremony.

While the matcher held uncompilable permission-rule syntax the hook fired on nothing, so the contradiction was latent; PRs #1239 and #1243 hit it live.

## Which policy yielded, and why

**The gate yields — for tickets-only diffs only.** Ticket 0365 (option 3b) settled that this gate is procedural: it asserts "the review pass ran", not "an independent party approved", because the project has one forge identity. For a diff that is only `tickets/*.erg`, the substantive checks are `erg check` and the ID-collision scan; satisfying the gate would mean the same account posting a `review:trivial` label and a self-review — ceremony carrying zero information. Making the rule yield instead would re-impose exactly the friction the rule was written to remove ("the friction was procedural, not a check") and tax every autonomous ticket filing.

## What changed

- `check-reviews.sh`: before counting reviews, the hook fetches the PR's file list. If every changed file — rename sources included — is a `.erg` under `tickets/`, it allows without review cycles. The check is conservative: an empty list, a list at the 100-file pagination cap (possibly truncated), any non-ticket path, or any API/parse error means **no exemption** and a fall-through to the unchanged review count. The gate for code PRs is untouched, error behaviour is byte-identical to before, and the new block can neither crash the hook nor deny on its own.
- `tests/test_check_reviews.py`: new `TestTicketsOnlyFastPath` (integration tier) pins the boundary — filing PRs and close-renames into `tickets/closed/` allow at 0 reviews; mixed diffs, code renamed into `tickets/*.erg`, `tickets/.ergrc`, empty/truncated file lists, and garbage API responses all keep the full gate. Written red-first (0fa31b30).
- `.claude/rules/git.md`: the fast-path bullet now names the hook exemption, so rule and gate cite each other.

## Verification

- `rtk proxy uv run python -m pytest tests/test_check_reviews.py -q` → 24 passed.
- `make check-fast` → exit 0. (`make lint`'s 4 pre-existing failures in `tests/test_datapaper_archive_layout.py` are `origin/main`'s, under repair separately.)


MinhHaDuong (commented):
Review cycle 1 (procedural gate, ticket 0365 option 3b). Checked: the exemption regex requires every filename and rename source to fullmatch tickets/.+\.erg; empty and 100-file lists refuse the exemption; all error paths end in TICKETS_ONLY=no and fall through to the pre-existing review count, so code PRs and error behaviour are unchanged. Red-first test coverage in TestTicketsOnlyFastPath pins both the allow and the deny sides. Reconciliation direction (gate yields for tickets-only diffs, rule stands) matches the recorded rationale in rules/git.md and ticket 0365.

PR #1252, opened 2026-07-28
feat(0357): register corpus-report in the vars registry, drop its hand-typed thresholds

MinhHaDuong:
## What

corpus-report reads `../_shared/technical-report-vars.yml` but was absent from `compute_vars.DOC_VARS`, so it rendered **12 distinct keys (22 occurrences) as `?meta:` placeholders** and kept its filtering thresholds as hand-typed literals.

The two halves are one defect, which is why the ticket could not be split: declaring `filter_reranker_threshold` under `technical-report` trips `test_doc_vars_no_extras`, and not declaring it renders an empty macro. Neither branch works without wiring the document into the registry first.

## Registry shape

Sharing a vars file is the existing reality, so the registry now *says so* rather than leaving the render to discover it. `DOC_VARS_FILE` maps document → path; four documents point at the shared file and the writer unions their declared keys per path.

The alternative — one vars file per document — would have duplicated the same 25 corpus measurements four ways and given Quarto four files to load where the `metadata-files:` front matter names one.

`multilayer-detection-techrep` and `breakpoint-detect-method-zoo` were unregistered for the same reason and are registered here too.

The registry moved to `scripts/analysis/_vars_registry.py`: adding corpus-report's 25 keys pushed `compute_vars.py` from 798 to 863 lines against an 800-line cap. `compute_vars` re-exports both names, so every guard that reads the registry is unchanged.

## A live factual error, not just staleness

Flag 6's concept groups read *"climate, finance, development, environment"*. Those are the **query-time** groups of `config/openalex_queries.yaml`, not the four that `corpus_filter.yaml` actually scores against (`climate, finance, policy, energy`). The two sets diverged and nothing moved the prose. It is a macro now.

## What deliberately stayed literal

The annex's two `0.0049` narrate the **superseded** initial threshold — a historical fact that must not move when the config does. An HTML comment says so, and notes that the 81%/74%/76% figures are measured at the live threshold, so a recalibration means re-running the human-validation sample, not just editing config.

## Verification — not just a green suite

- **Render oracle**: all 11 deliverables render, **0 placeholders, 0 skips**. The skip count matters — an earlier run reported "38 passed, 2 skipped", and the skip was the tell that corpus-report was never actually rendered.
- **Propagation**: setting `reranker_threshold` to `0.0031` and rebuilding produced 5 occurrences in the rendered corpus report and 0 of the old value; restoring it returned both vars files byte-identical.
- **Byte-compare**: `data-paper-vars.yml` and `multilayer-detection-vars.yml` regenerate byte-identical to main (key order preserved with `dict.fromkeys`, not a set) — the proof that the module split is a pure refactor.
- **Mutation-tested**: renaming corpus-report in the registry fails both new guards with the messages they promise.

`KNOWN_UNRESOLVED` in `test_meta_macro_resolution` is now **empty** — corpus-report was the only entry it ever held.

## The guard, and why it is narrower than the ticket asked

The ticket asked for a guard against any config value hand-typed in an include. Measured, that is not implementable: a standalone-token scan for `"2"` matches 73 lines across the 34 shared includes, `"50"` matches 13 — section numbers, tier labels, sample sizes. A guard with that signal-to-noise is an allowlist with a test wrapped around it.

Scoped instead to config values whose literal form is a **decimal fraction** and that the vars layer already exposes as a macro. It reads the value from config at test time, so recalibrating `0.002` → `0.003` re-aims the scan rather than blessing the stale number.

## One guard shipped blind and was caught

The prose-propagation guard added in the last commit passed on first write. Replaying the **actual** README text failed to trip it: the document name and the word "absent" sat on different lines with a sentence boundary between, and the proximity regex's character class excluded both. It is paragraph-scoped now, and both new guards are pinned by replaying the exact text they were written for.

## Pre-existing failures on this branch — none of them mine

- ~~`test_datapaper_archive_layout` × 4~~ — **resolved.** `origin/main` was red: `products_named_in` returned `None` after Copilot Autofix `d7d9789f` dropped its `return`, reproduced on a pristine checkout. A parallel session landed the same one-line fix as `002cfd65` while this PR was open, so my #1242 was closed as superseded. `make lint` is green here: **325 passed**.
- **`test_reranker_cache_exists`** — the one remaining failure, and it is environmental. `llm_relevance_cache.csv` is gitignored *and* declared in neither `dvc.yaml` nor `dvc.lock`, so no fresh worktree can obtain it — `make check` is structurally un-greenable outside the primary checkout. Pre-existing, unrelated, not filed.

## Other residuals found, deliberately out of scope

- **Flag 2's "shorter than 50 characters"** traces to a Python literal (`scripts/filter_flags.py:127`), not to config — it *cannot* trace to config without a Phase-1 change.
- **`make data` does not check out `data/catalogs/run_reports.dvc`**, so a worktree running `make stats` immediately after writes `[MISSING]` into two committed data-paper variables. Suspected cause is the nested-DVC-output overlap at `dvc.yaml:160` (cf. ticket 0430, now closed).
- The reranker **model name** is a live config value narrated in prose but left literal — a name, not a threshold.

## Merge with #1237 — done, and union-verified

PR #1237 (ticket 0420) merged while this was open and **renamed `tests/test_meta_macro_resolution.py` → `tests/test_render_placeholders.py`**, widening it from unresolved meta keys to the whole unresolved-reference class. This branch edits that file, so the two are composed, not either taken whole.

Integrated by **merge, not rebase** — force-push is denied in this repo, and rewriting a pushed branch's SHAs would leave local unable to fast-forward.

One conflict, in `README.md`, and it was the substantive one: main's side still described corpus-report as unregistered with 12 `?meta:` placeholders — the defect this PR fixes — while naming the new file. Resolution keeps this branch's content, main's filename, and main's broader description of the guard's scope.

Union verified rather than trusting the clean exit, since a clean auto-merge is exactly how a sibling's additions vanish. The merged `test_render_placeholders.py` carries 0420's `test_render_oracle_flags_a_broken_crossref`, `…_a_caption_crossref`, `…_a_missing_citation` and `test_citing_deliverable_declares_a_bibliography` alongside this branch's emptied `KNOWN_UNRESOLVED`. No reference to the old filename survives anywhere in the tree.

Re-verified post-merge: render oracle **54 passed, 0 skips**; the four registry/threshold guard modules **118 passed**; vars files still regenerate with zero drift.

## Corrections to an earlier draft of this description

Two errors the review panel caught, both mine: it cited **ticket 0410, which does not exist** — the DVC-overlap ticket is **0430** — and it listed #1237 as still open.

**Ticket:** tickets/0357-corpus-report-is-outside-the-vars-registr.erg


MinhHaDuong (commented):

**Verdict: approve.** No blockers. Four perspectives returned `approve`, red team returned `comment` with all findings self-declared non-blocking. One finding is doubly-confirmed and worth a follow-up; the rest are style/record.

### Test gate

`make check` → **1 failed, 2232 passed, 38 skipped** (284s).

The single failure is `test_corpus_acceptance.py::TestFileExistence::test_reranker_cache_exists`, and the PR body's claim that it is environmental checks out on inspection rather than on assertion: `data/catalogs/llm_relevance_cache.csv` is gitignored (`data/catalogs/.gitignore:22`) *and* named in neither `dvc.yaml` nor `dvc.lock`, so no worktree can obtain it by any supported route. The test is byte-identical on `origin/main`. Not caused by this PR.

`make lint` green (325 passed), render oracle `test_render_placeholders.py` **54 passed, 0 skips**.

---

### Findings

**1. `verifiable:` — `DOC_VARS` and `DOC_VARS_FILE` have no parity invariant** *(correctness + red team, independently)*

The split turned one dict into two, and nothing asserts they agree. Three call sites index the second by the first's keys:

- `scripts/analysis/compute_vars.py:654` — `per_file.setdefault(DOC_VARS_FILE[doc_name], …)`
- `tests/test_handoff_artifacts_tracked.py:87` — `path = DOC_VARS_FILE[doc]`
- `tests/test_retrieval_protocol.py:293` — `path = DOC_VARS_FILE[doc]`

Evidence, both directions:
- *Forward*: injecting a `DOC_VARS` entry with no `DOC_VARS_FILE` counterpart and re-running the union loop raises a bare `KeyError: 'fake-doc'` — no diagnostic naming the actual cause. All three call sites fail the same way, so the guards cannot report it either.
- *Mirror*: a `DOC_VARS_FILE` entry with no `DOC_VARS` key is **silently never written** — no error at all, the file just does not appear.

Parity holds today — verified, `set(DOC_VARS) == set(DOC_VARS_FILE)`, 3 distinct paths — so nothing is broken now. This is a latent gap the split introduced, and it bites the next registry edit. A one-line assertion in `_vars_registry.py`, or a fast test asserting set equality, closes it.

**2. `consider:` — `SHARED_VARS` naming is off-convention** *(consistency)*

`_vars_registry.py:30`. The `_VARS` suffix elsewhere in the codebase (`RETRIEVAL_VARS`, `FLAG_RULE_VARS`) denotes a *list of variable names*; `SHARED_VARS` holds a *path*. `DOC_VARS_FILE` signals path via `_FILE`; this one reads as if it should be a list. Module-private, low blast radius.

**3. `consider:` — the `dict, not set` comment claims a reason that is not load-bearing** *(correctness + red team)*

`compute_vars.py:652-653` justifies `dict.fromkeys` by first-declared-order stability. But `write_yaml` emits `for key in sorted(v.keys())`, so the written bytes are order-independent regardless of what `per_file` holds. The choice is harmless; the stated justification is not the real one, which is the kind of comment that misleads the next reader into preserving something that does not matter.

**4. `consider:` — redundant `sys.path` insert** *(consistency)*

`tests/test_config_thresholds_not_hand_typed.py:39-40` inserts both `scripts` and `scripts/analysis`. Every sibling test (`test_retrieval_protocol.py`, `test_doc_vars_completeness.py`, `test_handoff_artifacts_tracked.py`) inserts only `scripts/analysis`, since `scripts` already comes from pytest's `pythonpath`. Style outlier.

**5. `nofollow:` — stale symbol in another ticket's prose** *(consistency)*

`tickets/0374-grouped-target-output-guard-is-blind-to.erg:31` argues from `compute_vars.py` "routing each of its three writes through `DOC_OUTPUT_DIR` internally". The *behaviour* 0374 depends on still holds — `main()` still ignores `io_args.output` and routes internally, now via `DOC_VARS_FILE` — but the symbol it names is deleted here. Fix at 0374 pickup, not on this branch.

**6. `nofollow:` — guard module wider than the ticket's one-test ask** *(scope)*

`tests/test_config_thresholds_not_hand_typed.py` ships five tests where the ticket's Test section asked for one. Each extra guards a specific silent-pass mode of the main guard (vacuity, planted-defect self-test, allowlist redundancy) — the exact failure class this project has been bitten by twice. Recorded as scope growth; no action wanted.

---

### Hypotheses the panel tried to break and could not

Stated because these were the highest-risk failure modes, and a clean result is only informative if the probe was real:

- **Include-closure completeness.** The panel was pointed at `data-paper` on the theory that it composes `corpus-filtering.md` and would render `?meta:` for the four new macros. It does not compose it — only `corpus-report.qmd` includes either edited file, verified by a whole-tree sweep of include paths. `corpus-report`'s registered key list covers all eight macros the two files now use. Render oracle confirms: 54 passed, **0 skips** (a skip is the tell that a deliverable was never actually rendered).
- **Both new guards mutation-tested red.** `test_config_thresholds_not_hand_typed` fails against `git show origin/main:…/corpus-filtering.md` (3 hits); the registry guard fails when `corpus-report` is removed from `DOC_VARS`. Neither is a rubber stamp.
- **Make/registry drift.** `set(DOC_VARS_FILE.values())` equals the three paths in `COMPUTED_STATS` exactly. `metadata-files:` front matter in all six documents matches `DOC_VARS_FILE`.
- **The ticket's invariant.** `0.0049` is still a literal, twice, in `annex-crossencoder.md`, with an inline comment stating why. The 81/74/76% figures correctly stayed literal — they are measured outputs of a calibration run, absent from config, so outside the guard's remit by construction.
- **The concept-group correction is real, and the leftovers are not stale.** `origin/main`'s prose read "climate, finance, development, environment" — `openalex_queries.yaml`'s *query-time* groups, not `corpus_filter.yaml`'s Flag-6 *scoring* groups (`climate, finance, policy, energy`). The four surviving "development, environment" occurrences elsewhere describe the query-time mechanism accurately and are correctly untouched.
- **The split was forced, not elective.** `origin/main`'s `compute_vars.py` was 798 lines against `TestModuleLength`'s hard `MAX_MODULE_LINES = 800`. Post-split: 675 + 209. The claimed 863-line counterfactual is consistent.
- **Ordering / byte-stability.** `write_yaml` sorts; output is order-independent. No non-determinism.
- **`PINNED_DOCS` is not a mute skip.** `test_no_pinned_document_is_also_generated` guards the overlap, and `test_render_placeholders.py` is parametrized over all discovered `.qmd`s including pinned ones.

---

🤖 Panel: correctness, consistency, scope, red team, doc propagation — all `model: sonnet`, fresh contexts.

PR #1120, opened 2026-07-24
Prose restructure: table 4 to prose, grouped variables table, Data section split, condensed conclusion

MinhHaDuong:
Author checkpoint directives, 2026-07-24 teabreak round:

- **Table 4 → prose**: one paragraph per OpenAlex limitation, absorbing the duplicate dedup paragraph.
- **Table 5**: Group column replaced by horizontal rules at group boundaries (raw-LaTeX longtable in a crossref div); groups named in the caption only.
- **Data section split**: '3. Data description' / '4. Descriptive statistics' (the three figures live in the second); conclusion renumbered to 5; CC BY source counts corrected (5 automated / 3 restricted).
- **Conclusion condensed** 411 → 341 words (measured); referee-reply paragraph and the task-specific-cleaning concession intact.
- Renders clean at 16 pp; prose ≈3,146 words; 3 stale test pins retargeted (negative-polarity guards kept).

Ticket: none
Ticket-ref: tickets/0274-rdj26561-rr-round-1-tracker.erg

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1258, opened 2026-07-28
fix(0338): the multilingual claim outran the language table, and one bucket went unnamed

MinhHaDuong:
All four external reviewers of the 2026-07-27 panel flagged the same gap: the data paper's multilingual framing ran ahead of its own language table. One of them found an arithmetic error inside it.

**Ticket:** tickets/0338-datapaper-multilingual-claim-vs-english-share.erg

## The denominator slip

The paper published *"the non-English layer counts 3,381 works"* — total minus English, silently folding in the works whose language could not be classified. The count became a variable under 0323, but nothing stopped the shortcut from returning: only two of the three buckets were named.

`compute_vars` now publishes `lang_unclassified_n` beside `lang_non_english_n`, so all three are named and a reader can check the sum on the page. §2.3 prints both.

**The rebuild changed the size of the defect, not its shape.** With 0297's language backfill landed, the unclassified bucket is **11** works, not 1,358 — total-minus-English now overstates the layer by 11 rather than by 67%. The guard still earns its lines: the shortcut it forbids is available again after any harvest that outruns the backfill.

Verified partition on the real corpus: 93.8% × 33,344 ≈ 31,272 English + 2,061 classified non-English + 11 unclassified = 33,344 exactly.

## The claim gap

Four wordings asserted a cross-language *capability* the corpus never demonstrates — the multi-language UNFCCC and OECD documents are indexed in English only (§2.1), grey literature is 0% non-English, and no retrieval evaluation by language exists:

| Where | Was | Now |
|---|---|---|
| Abstract | "embeddings to support cross-lingual analysis" | states the English share; "embeddings pre-computed by a multilingual sentence-transformer" |
| §1 | "place all works in a shared semantic space regardless of language" | "searched in eight languages, and embedded with a sentence-transformer trained to place texts of any of them in one semantic space" |
| §1 | "and cross-lingual studies" | "and case studies of scholarship published outside English" |
| §5 | "cross-lingual bibliographic object" … "comparing how the category travelled across languages" | "multilingual bibliographic object" … "adds a non-English layer that the English-only queries of the prior mappings cannot reach" |

§2.3 also now states plainly that no retrieval evaluation by language was run — a limitation the reviewers asked for.

## Action 6 (the optional diagnostic) is not done, and cannot be as specified

It asks for nearest-neighbour alignment on UNFCCC/OECD documents that exist in two languages. §2.1 records that the corpus indexes only the English version of those, so the pairs it needs are not in the data by design. The claims were calibrated down instead — the branch the ticket names for the no-diagnostic case. `test_language_claims.py` lifts its claim guard automatically if a diagnostic ever publishes an `xling_*` variable, so the door stays open without a code change.

## TDD trail

- **Red** (`1c7b794c` → rebased): 5 failures — the missing bucket, the unregistered var, the absent prose shortcode, and the forbidden claims in both the two-section and whole-body scopes.
- **Green**: `lang_unclassified_n` derived and registered; four prose passages calibrated.
- **Regenerate**: vars file, byte-compared (below).
- **Condense**: `compute_vars.py` hit 804 lines against `test_no_god_modules`' 800 cap; merged the language block's three comment paragraphs into one keeping every load-bearing fact, back to 798. *(That headroom is now moot: 0357 landed mid-review and split `DOC_VARS` out to `_vars_registry.py`, so the file is **676** lines. The condensation stands on its own merits — the cap is no longer what motivates it.)*
- **Review rounds** (below): two rounds, every finding fixed.

## Regeneration evidence

The vars diff is **exactly one line**. `make stats` was started and abandoned deliberately: it re-runs the Crossref citation-verification probe over the network, which would churn `verify_*` values this ticket does not touch. Instead `compute_vars.corpus_stats` ran against the checked-out corpus and reproduced **all 12 shared keys byte-identically**, proving the corpus behind the committed file is the one on disk. `make corpus-tables` then regenerated the tracked `tab_corpus_sources.csv` with **zero** diff — independent corroboration.

## Render oracle

`quarto render data-paper.qmd --to markdown` exits 0 with **zero** `?meta:` / `?@ref` placeholders in stdout, and stderr carries only `FloatRefTarget` notices (a markdown-format limitation, not a defect). The abstract renders "English accounts for 93.8%"; §2.3 renders "2,061 works … a further 11 carry no language tag".

## Gates

- `/verify-adherence`: **PASS** — 322 adherence tests passed, 8 skipped; no mechanical failures, no semantic findings. **The mechanical phase can be skipped on the next `/gaze` pass.**
- `make check`: **2,216 passed**, 1 failure remaining, not from this branch.

### The one failure, and why it is not mine

`test_corpus_acceptance.py::test_reranker_cache_exists` — `llm_relevance_cache.csv` missing. This is **ticket 0450 defect 2** verbatim ("a test asserts on a file no clean clone can have"), and a consequence of its defect 1: `make data` aborts at exit 255 on `catalog_merge_report.json`, a `dvc.yaml` out that `dvc.lock` never recorded, *before* it reaches the reranker cache. Open PR #1241 files both and states both reproduce on `origin/main`. Neither test touches a file this PR changes. Not fixed here — separate ticket, separate PR.

A second failure in an earlier run (`test_refined_embeddings_freshness`) was an artifact of my own `dvc checkout --relink`, which wrote the two files 5 ms apart; `dvc status` reported no modified outs, and a `touch` restored the ordering.

## Prose cost

+33 words on a document already over its 2,500-word limit, all of it limitation statements the reviewers asked for. That budget is settled in `revision-rdj26561/cut-plan.md`, not here.

## Not done, flagged for the author

The PR rewrites the Abstract and Conclusion of a manuscript under revision but adds no line to `revision-rdj26561/summary-of-revisions.md`. That file is organized by journal remark ID and this ticket came from the internal panel, not the journal's 18-remark ledger — so nothing is formally owed. Whether the editor should be told is an author call.

---

## Review round 1 — two guard defeats, both reproduced, both closed

1. **The escape hatch was forgeable.** Adding `xling_foo: "0"` to the vars file made both claim guards `pytest.skip()`, and the panel ran the suite green with "cross-lingual analysis" live in the abstract. A one-line variable should not disarm a guard. The conditional lift is **removed**; retiring a pattern now means editing the test alongside publishing the measurement, where a reviewer sees it.
2. **The phrase list was three literal substrings.** "travelled across **the eight** languages" walked past `"across languages"` — 7/7 passed. Now patterns, covering the interposed count, `crosslingual` / `cross-language` / `cross linguistic`, and the capability constructions.

Finding the `\b` fix cost a false positive worth recording: without it the optional-whitespace class straddles a word boundary and the guard fires on the abstract's own *"works across linguistic contexts"* — a retrieval claim, not a capability claim. A second red-test now pins the other polarity, because a guard that fires on the paper's own concessions pushes the author to delete the concession rather than the claim.

Also from round 1: `_count()` strips the thousands separator `compute_vars` emits above 999; §2.3's dangling "That" became "Coverage at that level"; the abstract says "refined works", matching §2.3's name for the same variable.

## Review round 2 — a coverage hole and a stale record

1. **The wide guard started at `## 1. Introduction`.** The title, the keywords list, and the Related-dataset bullet were unscanned; review replanted the originating string into each and all three passed. It now reads from byte zero, stripping only HTML comments. **Verified by replanting `cross-lingual analysis` into each of the three regions in turn — each now fails.** The title is the worst of the three: it is the PDF cover and the Zenodo citation string.
2. **Non-ASCII hyphens evaded `\-`.** U+2010/2011/2013 and the soft hyphen all render as "cross-lingual" and all missed. All four now in the class and pinned in the red-test.
3. **Three false statements in the permanent record, corrected.** The ticket claimed the guard "lifts automatically" on an `xling_*` variable — which round 1 made false when it defeated that hatch; it cited `compute_vars.py:280` for a derivation 0357's split moved to `:158-163`; and this PR body's 800-line-cap narrative is annotated above.

**Declined, with reason:** pinning `lang_unclassified_n`'s *value*. A rebuild that reintroduced a 1,358-work bucket is not itself a defect — the defect was quoting total-minus-English as a multilingual quantity, which the partition guard already forbids. Pinning the value would bless the next wrong one.

## Merged main rather than rebased

0357 (PR #1252) landed mid-review and moved `DOC_VARS` out of `compute_vars.py`, conflicting with this branch's registration. Force-push is denied here, so a rebase would strand the pushed PR — integrated with `git merge origin/main` instead. Resolved to main's side wholesale and re-added the entry to the registry, then **grep-verified the union rather than trusting the clean auto-merge**: `compute_vars.py` keeps main's registry import and `__all__` re-export, and this branch's `lang_unclassified_n` survives on both sides of the split (computation at `compute_vars.py:163`, declaration at `_vars_registry.py:146`).

Post-merge gate: **2,250 passed**, same single pre-existing failure. `mergeStateStatus: CLEAN`.

## Coordination warning — worktree `t334`

Branch `t334-datapaper-claims-calibration` (ticket 0334, **no PR yet**) has three local commits editing **the same four paragraphs** of `data-paper.qmd` from the same base as this branch, and its §5 still reads *"cross-lingual bibliographic object … comparing how the category travelled across languages"* — the exact overclaim this PR removes. Both branches rewrite the same single-line paragraphs, so git will conflict rather than silently drop either side; whoever merges second must resolve by hand and keep both intents. That session also carries its own `_vars_registry.py` move, now superseded by 0357 on main. Flagged, not touched — it is a live session's worktree.

## Second merge of main — composed with 0337's language ablation

0337 landed while this PR was open and rewrote **the same §2.3 paragraph**. The merge is a composition, not a pick: every sentence 0337 added is kept verbatim (per-stratum removal shares, the abstract-coverage control, the own-language rescoring result, the `tab_filter_ablation.csv` pointer), with this branch's classified/unclassified split layered on top. Verified by counting each side's markers in the merged file rather than trusting the auto-merge — four of 0337's, two of this branch's, all present.

**Two of my own sentences did not survive, and should not have:**

- *"balanced cross-language comparisons"* was the inherited wording, and this branch's own guard forbids `cross-language`. Now "balanced comparison" — the same limitation, without claiming the capability in the act of denying it.
- *"we report no retrieval evaluation by language"* became **false** the moment 0337 merged: its ablation *is* one, and finds that rescoring non-English works against an own-language query lifts the share passing the threshold by 11.5 points. Carrying it into main would have shipped a statement the paragraph below it contradicts. Dropped, and the ticket corrected to say so — the reviewers' "no retrieval evaluation by language" complaint is answered by 0337's measurement, not by my disclaimer.

Post-merge gate: **2,262 passed**, same single pre-existing failure (ticket 0450).

*Erratum:* the merge commit's message lost one backticked word to shell command substitution — it reads "this branch's own guard forbids ." and should read "forbids `cross-language`". Force-push is denied on this branch, so it is corrected here rather than by rewriting history.

## Also filed

**#1265** (ticket 0590) — round 1 found `multilayer-detection.qmd:226` still asserting the shared-semantic-space property this PR calibrated out of the data paper. Different deliverable and an author's judgment call, so filed rather than folded in here.

---

## Round 3 — the multilayer paper, and the t334 collision landing

**The multilayer claim, fixed here rather than ticketed.** `/review-pr` found `deliverables/multilayer/multilayer-detection.qmd` still asserting the property this PR calibrated out of the data paper. It was filed as ticket 0590 (#1265, merged); the author asked for a direct fix instead, so it lands here and this PR closes 0590.

Two sentences carried the unattributed indicative, not one:

| Where | Was | Now |
|---|---|---|
| §2 (corpus description) | "The multilingual embedding model **places** works in English, French, Chinese, Japanese, and German into a shared semantic space." | "…**is trained to place** works…" |
| §6.1 (LDA/STM comparison) | "Multilingual sentence-transformers (here `BAAI/bge-m3`) **place** documents in a shared semantic space regardless of language" | "…**are trained to place** documents…" |

Review found §6.1. **§2 was found by running a pattern over the file**, and it is the worse of the two: it describes the *dataset*, which is the register 0338 is about, where §6.1 is a methods comparison. Both now use the form the data paper already uses in §1 and §3 ("trained to place", "designed to group"), so the two deliverables state the same strength for the same model.

A guard pinning this shape was drafted and **reverted at the author's direction** — 115 lines to guard two phrases was out of proportion (commits `5816bd0b` / `662ee83f`, kept in history rather than force-pushed away). Recorded for whoever revisits it: the guard keyed on attribution rather than vocabulary, and it was a source-text pattern match, so a synonym for "place" evades it.

**The t334 collision landed.** PR #1260 merged while this branch was open, editing the *same three paragraphs* from the same base — the collision this PR flagged before it happened. Neither side is a superset, so the merge composes:

- *From main:* the query-level overlap probe rewritten as a bounded claim, the caveat that Scopus/WoS shares absorb OpenAlex coverage gaps, the scope-not-retrieval reading of the 40.1%, the `institutional-reports` label, and the clause saying what the institutional layer contributes.
- *From this branch:* the four claim calibrations and the classified/unclassified split.

Three of main's phrasings did not survive, each being the defect this ticket removes: `cross-lingual bibliographic object`, `permits comparing how the category travelled across languages`, and `embeddings that place all works in a shared semantic space regardless of language`. Main's §2.3 wording — "balanced **cross-language** comparisons" — would have failed this branch's guard, which is the guard earning its place.

Verified by counting each side's markers in the composed file rather than trusting the resolution: **13 required present, 4 forbidden absent, 0 failures.** A first attempt rebuilt the file from main's copy plus the three resolved paragraphs and silently reverted two calibrations sitting in *auto-merged* regions — exactly the drop `.claude/rules/git.md` warns about — so the final resolution touches conflict hunks only.

**Gate on the composed tree:** 1,387 fast + 334 adherence passed, 0 failed. t334's own prose guards (now adherence-marked) pass on the calibrated wording.

**Ticket:** tickets/0590-multilayer-paper-asserts-an-unvalidated-em.erg


MinhHaDuong (commented):
## `/review-pr` — five-perspective panel on #1258

**Panel verdict: comment.** Four seats approve/comment, one (red team) returns **request-changes**. Dissent preserved below — the author arbitrates.

The shipped content holds up under direct verification. Three independent seats recomputed the language partition against the real corpus in the worktree and got `31,272 + 2,061 + 11 = 33,344` exactly, byte-identical to `data-paper-vars.yml`. `make check-fast`: **1,348 passed, 8 skipped**. The two contested findings are about the *guard's* durability across the revision cycle, not about the manuscript this PR ships.

---

### Red team — request-changes (dissent, both reproduced and reverted)

**1. `verifiable:` The claim guards disarm on the mere presence of an `xling_` key, whatever its value.** `tests/test_language_claims.py:120-124`. Red team appended `xling_foo: "0"` to `data-paper-vars.yml`, reinserted the ticket's own defect wording ("cross-lingual analysis") into the abstract, and ran the file: **5 passed, 2 skipped** — green, with the forbidden claim live in the manuscript. A stray future var, a typo, or an abandoned experiment silently disarms both guards permanently. Fix is ~2 lines: require the diagnostic key to carry a truthy value, not just the prefix.

**2. `verifiable:` The forbidden-phrase list is three literal substrings, and the ticket's own original wording evades it under a one-word edit.** `tests/test_language_claims.py:170`. Red team replaced §5 with *"permits comparing how the category travelled across **the eight** languages"* — the ticket's cited defect with two words inserted — and got **7/7 passed**. Also confirmed misses: `crosslingual`, `cross lingual`, `cross-language`, `between languages`, `compare languages`, `translingual`, `multilingual analysis`. This guard exists to survive a manuscript under active revision (`revision-rdj26561/`); a literal-substring list is the wrong instrument for that job. Suggested: word-boundary regex on `cross[- ]?lingu`, `across (the )?(\d+ )?languages`, `compar\w* languages`, `between languages`.

Red team explicitly cleared what it could not break: the three-way partition (probed `NaN`, `None`, `""`, `"EN"`, `"en-GB"`, `"und"` — buckets summed to total every time); the `ValueError`-on-heading-rename path (fails loudly, not silently); and `test_guard_would_catch_the_wording_it_was_written_against`, which does catch an emptied `UNDEMONSTRATED`.

### Scope — the finding outside this diff

**`verifiable:` Ticket 0334's in-flight branch reintroduces the exact overclaim this PR removes.** Worktree `t334-2919364` (branch `t334-datapaper-claims-calibration`, no PR open yet) has three local commits rewriting the *same four passages* — abstract, §1, §2.3, §5 — from a pre-#1258 base. Its §5 still reads *"a single, cross-lingual bibliographic object … permits comparing how the category travelled across languages"*. With no CI in this repo, a merge without rebasing onto #1258 either conflicts or silently restores the wording. **Action:** the 0334 session rebases onto #1258 before opening, and whoever merges second runs `grep -c "cross-lingual\|across languages" deliverables/data-paper/data-paper.qmd` against `origin/main`. PRs #1253 and #1254 also touch `data-paper.qmd` but in non-adjacent paragraphs — low risk.

Scope also verified each ticket action against the file rather than the checkbox: actions 2 and 4 landed earlier under 0329 and 0323 and are correctly *not* duplicated here; action 6's infeasibility is genuine (§2.1:84 records English-only indexing of the bilingual UNFCCC/OECD docs, so the pairs the diagnostic needs do not exist). `nofollow:` action 5's "and why" is stated nowhere — but that sentence landed in PR #1120, not here.

### Consistency

**`verifiable:` Ambiguous pronoun in the new §2.3 sentence** — `data-paper.qmd:116`. *"…a further 11 carry no language tag. **That** supports case studies, not balanced comparison between languages."* The nearest antecedent to "That" is the unclassified count; the intended referent is the non-English layer, two clauses back. Survives to the rendered output (verified at markdown line 440). Suggest folding the counts into the claim sentence.

**`verifiable:` Two nouns for one quantity** — abstract `:24` says English accounts for *"% of **the result**"*; §2.3 `:116` says *"% of **refined works**"*, same variable. One concept, one term.

**`consider:` The condensed comment dropped the corpus-stage detail** — `compute_vars.py:280-287` no longer records that 3,381 was *"an enriched-corpus figure inside a refined-corpus sentence"*. The fact survives in `tests/test_citation_verification.py:130-133`, so nothing is lost repo-wide, but the two now narrate the same episode differently without cross-reference.

### Correctness

**`verifiable:` Latent `ValueError` in a test fixture** — `tests/test_language_claims.py:102` does `int(v["corpus_total"])` where line 89 correctly does `.replace(",", "")`. `_int()` always comma-formats, so scaling the fixture past 999 rows crashes the test. Reproduced with a 1,010-row frame. Masked today by a 13-row fixture; a trap for the next editor.

`nofollow:` The bucket normaliser is exhaustive by construction — `normalize_lang_display` (`scripts/pipeline_text.py:289-308`) maps every non-mapping value to `"unknown"`, and it is the same normaliser `@tbl-languages` uses, so prose and table cannot diverge. New tests run in 1.94s, correctly in the fast tier, well under the 5.0s ratchet.

### Doc propagation

`nofollow:` `lang_unclassified_n` is registered everywhere its siblings are (`DOC_VARS:109`, derivation `:293`, yml `:51`, qmd `:116`), and no other deliverable's registry needs it. Table 3's tracked artifact independently confirms the arithmetic: `418+308+295+256+784 = 2,061`, Unclassified row `11`, total `33,344`. The `{{< meta >}}` shortcode in YAML front matter is established prior art in this same file; `test_render_placeholders.py::…[data-paper]` passes on a real `quarto render`, abstract included.

**`consider:` One live deliverable still carries the flagged phrasing** — `deliverables/multilayer/multilayer-detection.qmd:226`: *"Multilingual sentence-transformers … place documents in a shared semantic space regardless of language."* The context differs (a tool-comparison sentence, not a validated claim about this corpus) and it is outside ticket 0338's declared scope. Worth a follow-up ticket if the claim discipline should extend to the method paper.

`nofollow:` `revision-rdj26561/` carries no remark response contradicted by the calibration — grep for the four phrases and `3,381` across `revision-rdj26561/*.md` returns zero. The PR body's "nothing formally owed" reads correct.

---

**Recommendation:** the two guard-robustness findings are cheap to fix and go to the durability of the only thing standing between this manuscript and the overclaim's return — worth landing before merge. The §2.3 pronoun is an author call on prose. The 0334 coordination flag needs an owner regardless of what happens to this PR.

MinhHaDuong (commented):

Correctness · Consistency · Scope · Red team · Doc propagation, all on `sonnet`, branch tip `aa2f0c1e`. Every finding below was reproduced by the synthesizer before posting.

**Verdict: comment** — four of five reviewers approve-with-comment; the red team returned **request-changes**, and its dissent is preserved verbatim in §1. Nothing found is a defect in shipped behaviour. What is worth fixing before merge is a guard-scope hole and three statements in the permanent record that are not true.

`make check-fast`: **1375 passed, 9 skipped, 0 failed**. `tests/test_language_claims.py` + `test_doc_vars_completeness.py` + `test_language_table.py`: **45 passed**.

### What holds up

The partition is not merely tested, it is structurally guaranteed: `normalize_lang_display` (`scripts/pipeline_text.py:289`) provably returns `"en"`, a real ISO 639-1 code, or the literal `"unknown"` for every input probed (`arz`, `sco`, `''`, `'   '`, `None`, `nan`, `'EN'`, `'en-US'`, `'und'`), so the three masks in `compute_vars.py:158-163` are mutually exclusive and exhaustive by construction. Two reviewers re-derived the counts straight from `data/catalogs/refined_works.csv`: 33,344 = 31,272 English + 2,061 classified non-English + 11 unclassified, exactly matching the published vars. `lang_unclassified_n` is correctly alphabetised in `_vars_registry.py:146`, the one-line `.yml` insertion sits where `write_yaml`'s `sorted()` puts it, and `3,381` is gone from both the paper and the response letter — the only surviving hits are the reviewer's own quoted text in `external-review/` and guard comments that cite it deliberately. Every ticket and PR cross-reference (0323, 0297, #1141) resolves and says what is claimed. The six regex patterns each match a phrase that was live on `main`; the two red-test functions are load-bearing, not ceremony.

---

### 1. The claim guard does not scan the title, the keywords, or the pre-Introduction bullet — **red team, CONFIRMED**

`verifiable:` `_hits` is applied to two slices only: `abstract: |` → `keywords: |` plus §5 (`tests/test_language_claims.py:148-154`), and `## 1. Introduction` → `## References` (`:220-228`). That leaves 437 chars of front matter ahead of the abstract and a 719-char gap after it unscanned. I replanted the *verbatim originating defect string* into each and ran the shipped patterns and slice logic:

```
defect replanted in TITLE        : narrow=MISSED wide=MISSED
defect in keywords list          : narrow=MISSED wide=MISSED
defect in Related-dataset bullet : narrow=MISSED wide=MISSED
```

`title:` at `data-paper.qmd:2` is the most exposed surface in the file — PDF cover, Zenodo record, citation string — and the Related-dataset bullet at `:37` carries the deposited dataset title verbatim. The current wording is correct ("Multilingual Retrieval and Grey Literature"), so there is no live defect; the hole is in coverage. Ticket 0338's exit criterion reads *"No multilingual claim in the paper exceeds what Table 3 demonstrates"* — the guard does not enforce that for the region a reader sees first.

Red team's verdict, verbatim: *"That's not a paraphrase-robustness nitpick, it's a scope-coverage bug in a guard whose entire job is scope coverage."* The fix is to start the wide slice at offset 0 rather than at `## 1. Introduction`.

### 2. Non-ASCII hyphens walk through `\bcross[\s\-]?lingu`

`verifiable:` `tests/test_language_claims.py:57` — the character class matches ASCII `-` only:

```
'cross-lingual analysis'  -> caught      'crosslingual analysis' -> caught
'cross‑lingual' (U+2011)  -> MISSED      'cross–lingual' (U+2013) -> MISSED
'cross­lingual' (U+00AD)  -> MISSED
```

All three render as "cross-lingual" to a reader. One-line fix: `[\s\-‐-―­]`. Worth bundling with §1 since both are edits to the same six lines.

### 3. Three statements in the permanent record are false

These land in `main` via the ticket body and the merge commit, and this project's `/verify-gate` treats file:line as load-bearing evidence precisely because it is checkable.

`verifiable:` **(a)** `tickets/0338-…erg:123-124` and the PR body both state *"`test_language_claims.py` lifts its guard automatically if a diagnostic ever publishes an `xling_*` variable."* The shipped code says the opposite at `tests/test_language_claims.py:42-50`: the `xling_*` escape hatch was tried, defeated in review (`xling_foo: "0"` disarmed the guard while the claim sat live in the abstract), and removed — *"There is no automatic lift."* Removing the hatch was the right call; the narrative describes the design that was rejected. Found independently by Scope and Correctness.

`verifiable:` **(b)** `tickets/0338-…erg:94` cites `compute_vars.py:280` as deriving the two language vars. The derivation is at `compute_vars.py:162-163`; line 280 is inside an unrelated function. Found independently by Correctness and Consistency.

`consider:` **(c)** The PR body's "Condense" section says `compute_vars.py` hit 804 lines against `test_no_god_modules`' 800-line cap and was trimmed to 798, "net line cost of this ticket is zero." The cap is real (`tests/test_script_hygiene.py`, `MAX_MODULE_LINES = 800`) and the claim was true against this branch's original base, but `aa2f0c1e` merged `main` and absorbed ticket 0357's independent shrinkage: the file is **676 lines** at branch tip, 124 clear of the wall. The condensation is harmless and in scope — all load-bearing facts survive the rewrite (arz/sco rationale, #1141 attribution, the 3,381-vs-2,023 provenance, the langdetect prefix-match reason) — but the stated pretext no longer describes the merged file.

### 4. Smaller findings

`consider:` **Vocabulary and paraphrase ceiling.** `translingual`, `interlingual`, `language-to-language`, "a bridge between English and non-English scholarship", "compares each language directly" (singular) all evade every pattern. Generic to keyword guards and the docstring is explicit that it pins wordings, not semantics — flagged only because this guard has now been defeated twice in review and keeps being extended reactively.

`consider:` **`lang_unclassified_n`'s value is pinned nowhere.** Only its presence (registry + `{{< meta >}}`) and its derivation formula (on an 11-row synthetic fixture) are tested. A rebuild that silently reintroduced a large unclassified bucket — exactly the pre-0297 state, 1,358 works — would pass every test in the file. The prose would still be *arithmetically* honest; it would just quietly stop being a footnote.

`consider:` **The partition test reconstructs English from a rounded percentage.** `tests/test_language_claims.py:116` derives the English count as `round(pct/100 * total)`. At real-corpus scale that yields 31,277 against an actual 31,272 — a 5-unit gap that would make the assertion spurious. Harmless today (the test only ever runs on an 11-row fixture where the round-trip is exact), but the technique does not survive being reused against real data.

`consider:` **`deliverables/_shared/tables/tab_languages.md` is untracked and regenerated by `make figures`.** In this branch's worktree it is current (31272 / 11 / 33344, matching the new vars). In the primary checkout on `main` the local copy is still pre-0297 (89.9%, 2023, 1358) — so a render run there without `make figures` first would print a table contradicting the prose this PR just fixed. Local-state hazard, not a PR defect; the code path is pinned equal by `test_language_table.py::test_prose_counts_bucket_codes_exactly_as_the_table_does`.

`nofollow:` **Section anchors are brittle but fail closed.** The six `text.index(...)` calls each match exactly once today; a renumbering would raise `ValueError`, not silently narrow scope. Acceptable.

`nofollow:` **`summary-of-revisions.md` and `docs/editorial-brief.md`.** Doc propagation settled the question the PR body left open: no remark in `revision-rdj26561/ledger.jsonl` touches the cross-language wording (this came from the internal 2026-07-27 panel), no test or rule requires logging every Abstract/Conclusion edit, and the editorial brief is scoped to `manuscript.qmd` by its own header. The PR's "not formally owed" reading is correct. Whether to tell the editor anyway remains an author call.

`nofollow:` **Word budget.** No automated gate covers the data paper; the 2,500-word limit is tracked by hand in `revision-rdj26561/cut-plan.md` under open ticket 0332, which already records the document ~881 words over independently of this PR. Deferring the +33 there is the established mechanism, not an evasion.

`nofollow:` **Action 6.** §2.1's *"For UNFCCC and OECD documents available in multiple languages, we index only the English version"* is pre-existing (commit `7614094e`, ticket 0274) and substantiates "infeasible as specified". Scope's one caveat: the ticket lists Action 3 (calibrate the claims) as an independent mandatory action, not as the sanctioned fallback for a failed Action 6 — the Outcome section's "the branch the ticket names for the no-diagnostic case" is a generous reading, not a false one.

---

**Recommended before merge:** widen the wide guard's slice to offset 0 (§1), add the non-ASCII separators (§2), and correct the two false statements in the ticket body (§3a, §3b) — the ticket is what `main` keeps.

🤖 `/review-pr` — Correctness · Consistency · Scope · Red team · Doc propagation

PR #1272, opened 2026-07-28
fix(0591): stop the genealogy chain and three sibling scripts from writing hash-seed order into artifacts

MinhHaDuong:
**Ticket:** tickets/0591-fig-genealogy-render-order-is-hash-seed.erg

## What was wrong

`load_edges()` in both genealogy renderers returned `list(set(...))`, so edge
draw order followed the interpreter's string-hash seed. Two runs of the same
code on the same inputs produced different bytes. The Makefile pins
`PYTHONHASHSEED := 0` globally, which is why every `make`-driven build looked
deterministic and the defect only surfaced in ticket 0571's refactor
byte-compare (PR #1269), where it read as a spurious regression.

Reading the chain turned up two sinks the ticket had not named:

- **`backbone_dois` was a set in both `load_model()`s.** It is enumerated to
  draw nodes, and passed to `sorted(..., key=cited_by_count)` — whose many ties
  fall through to set order, because Python's sort is stable.
- **`analyze_genealogy.py`, upstream of both, had the same defect in the model
  itself.** `compute_layout()` enumerates `backbone_dois` to assign the
  per-(band, year) jitter index, so the `y` coordinate written to
  `tab_lineages.csv` was hash-seed dependent. Fixing only the renderers would
  have left them reproducible from a non-reproducible input.

## The sweep (ticket action 2)

Two independent passes — a search agent over all of `scripts/figures` and
`scripts/analysis`, and the adherence pass's semantic reviewer — converged on
the same three files. All three are fixed here; none was a `list(set(...))`.
All three are the variant that reads as already-sorted: `sorted(<a set>,
key=...)` is *stable*, so tied keys fall through to hash order.

| File | Reaches | Fix |
|---|---|---|
| `export_tab_venues.py` | `tab_venues.md` — **git-tracked, included by the Œconomia manuscript** | `sorted()` on the journal-name union; `kind="stable"` on both `sort_values` |
| `plot_fig_traditions.py` | `fig_traditions.png` | DOI tiebreaker on the label-node sort |
| `plot_fig_traditions_pre2008_citers.py` | `fig_traditions_pre2008_citers.png` — **git-tracked, cited to RDJ26561 reviewers** | same |

`export_tab_venues.py` is the one that matters: `nlargest(5, "total")` keeps
the first row among equals, so *which journals a published table names*
depended on the hash seed. That is a live science-lane defect, not a latent
one.

Ruled out and worth recording so a later sweep does not re-flag them: sets used
only for membership or `len()`, `sorted(set(...))` with no key, and sets feeding
order-independent reductions (`np.median`, `sum`, modularity). One further
match, `compute_interpretation.py`, is the **same defect, dormant** — its
`vocab` set feeds a `sort_values(...).head(top_n)` where tied terms compete for
the cut — but no Make target drives it, so it reaches no deliverable today.
Left alone rather than widening the diff; recorded here so the next sweep
recognises it instead of rediscovering it.

## Tests

`tests/test_hash_seed_determinism.py` runs each producer twice under **two
different** hash seeds and byte-compares. Running under a single pinned seed —
as `tests/test_determinism.py` and the Makefile both do — would pass against
the defect, so the two-seed design is the whole instrument.

Four cases: `fig_genealogy.html`, `fig_genealogy.png`, `tab_lineages.csv`,
`tab_venues.md`. All four red before, green after. The genealogy three were
verified red against `origin/main`'s scripts in a scratch tree; the venue guard
was verified by reverting `export_tab_venues.py` to `origin/main` with the test
in place — it fails with its own assertion message, not an environmental error.

`make determinism-check` now runs both modules. It previously ran only
`test_determinism.py`, which pins the seed for both of its runs and so was
structurally blind to the class it is named for.

Coverage is deliberately uneven, and the module docstring says so rather than
leaving the gap silent: the two traditions renderers need a Louvain partition
over a real citation graph before any label ties exist, which no cheap fixture
reproduces. They are fixed by inspection.

## Design notes

- **Sorting for edges, insertion order for `backbone_dois`.** Edge insertion
  order is pandas row order over a DVC-managed CSV — stable, but not
  *canonical*, where an explicit sort is a property of the data.
  `backbone_dois` keeps CSV row order instead: `tab_lineages.csv` is itself
  generated and its row order is now pinned upstream, and re-sorting would
  discard the lineage grouping a reader of that file sees.
- **Membership tests rebuild a local set** (`backbone = set(backbone_dois)`).
  The citation frame is scanned row by row, so an O(n) list lookup there would
  have been a real regression.
- **The ticket's "one shared helper if they can share" is declined.**
  `load_model` / `load_edges` are byte-identical across the two genealogy
  renderers and the codebase has the precedent (`_band_scheme.py`,
  `_community_registry.py`), so the extraction is warranted — but it is a
  different change, with its own Makefile-prerequisite blast radius, and
  bundling it here would bury a determinism fix inside a refactor diff.

## Gate

- `make check-fast` — clean (1399 passed)
- `make lint` — clean (330 passed)
- `make determinism-check` — clean (7 passed)
- `make check` — 2229 passed, **8 pre-existing failures**, all corpus-file
  *existence* assertions in `test_corpus_acceptance` / `test_corpus_flow`. This
  worktree holds DVC pointers but no `make data`; `data/catalogs/` contains only
  `*.dvc` stubs. Nothing in the diff touches a corpus file, and the count is
  identical to the pre-branch baseline.
- `/verify-adherence` — PASS, no mechanical failures. Its semantic phase is what
  surfaced the three sweep hits, now fixed in this branch.
- `/review-pr` — five-agent panel, no blockers, recommend merge. The panel
  independently replayed the defect against all four covered producers and ran a
  20-seed probe confirming the `1`/`12345` pair is not coincidentally passing.

## Not done — tracked in ticket 0610 (PR #1273)

Three **git-tracked** artifacts were generated by the old ordering and no longer
match what their producer emits. None can be rebuilt here: this worktree has
`.dvc` stubs, not the corpus.

- `fig_traditions_pre2008_citers.png` — cited to RDJ26561 reviewers at
  `revision-rdj26561/r1-14-network-response.md:38`. Highest stakes of the three.
- `tab_venues.md` — Œconomia manuscript.
- `tab_venues_fr.md` — Make-generated from the above, French variant.

`fig_genealogy.{png,html}` and `fig_traditions.png` also change; none is tracked
and no golden hash pins them, so nothing in the repo needs regenerating for them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -858,10 +858,15 @@ audit-pdf-content:
 smoke:
 	$(PYTHON) -m pytest tests/test_smoke_pipeline.py -v --tb=short
 
-# Determinism check: run figure scripts twice on smoke data, diff outputs.
+# Determinism check: run figure scripts twice, diff outputs.
 # Catches unseeded randomness, leaking timestamps, floating-point non-determinism.
+# Two modules, and the second is the stricter one: test_determinism.py runs both
+# halves under a pinned PYTHONHASHSEED=0, so it cannot see a producer that walks
+# a set into its output, while test_hash_seed_determinism.py varies the seed
+# between the two runs and can (ticket 0591).
 determinism-check:
-	$(PYTHON) -m pytest tests/test_determinism.py -v --tb=short
+	$(PYTHON) -m pytest tests/test_determinism.py \
+		tests/test_hash_seed_determinism.py -v --tb=short
 
 # Regression hashes: compare Phase 2 output hashes against golden baseline.
 # Runs as pytest (one test per script, module-scoped fixture = scripts run once).
--- a/scripts/analysis/analyze_genealogy.py
+++ b/scripts/analysis/analyze_genealogy.py
@@ -115,21 +115,27 @@ def load_data():
 # ============================================================
 
 def select_backbone(works, doi_meta):
-    """Select highly-cited papers with abstracts and valid years."""
+    """Select highly-cited papers with abstracts and valid years.
+
+    Returns a DOI list in ascending DOI order. The order is part of the
+    contract, not an accident: downstream steps enumerate this collection to
+    assign layout jitter and to lay out CSV rows, so a `set` here would put the
+    interpreter's hash seed into `tab_lineages.csv` and, through it, into both
+    genealogy figures (ticket 0591).
+    """
     cite_threshold = _cfg["clustering"]["cite_threshold"]
     year_max = _cfg["periodization"]["year_max"]
     log.info("Selecting backbone (cited_by_count >= %d)...", cite_threshold)
 
     has_abs = works["abstract"].notna() & (works["abstract"].str.len() > 50)
     high_cited = works[has_abs & (works["cited_by_count"] >= cite_threshold)]
-    backbone_dois = set(high_cited["doi_norm"])
 
     # Filter to papers with valid year
-    backbone_dois = {
-        d for d in backbone_dois
+    backbone_dois = sorted({
+        d for d in set(high_cited["doi_norm"])
         if d in doi_meta and doi_meta[d]["year"] is not None
         and 1985 <= (doi_meta[d]["year"] or 0) <= year_max
-    }
+    })
 
     log.info("Backbone papers (with valid year): %d", len(backbone_dois))
     return backbone_dois
@@ -178,8 +184,8 @@ def assign_lineages(backbone_dois, doi_meta, doi_to_cluster):
         else:
             lineage[d] = doi_to_cluster.get(d, 0)
 
-    # Keep only papers with assigned lineages
-    backbone_dois = {d for d in backbone_dois if d in lineage}
+    # Keep only papers with assigned lineages, preserving the sorted order
+    backbone_dois = [d for d in backbone_dois if d in lineage]
     band_counts = {
         b: sum(1 for d in backbone_dois if lineage[d] == b)
         for b in range(N_COMMUNITIES)
@@ -196,16 +202,22 @@ def assign_lineages(backbone_dois, doi_meta, doi_to_cluster):
 # ============================================================
 
 def build_citation_dag(backbone_dois, cit):
-    """Build directed edges between backbone papers (cited → citing)."""
+    """Build directed edges between backbone papers (cited → citing).
+
+    Returns the edges sorted, so the list does not carry set iteration order.
+    ``backbone_dois`` arrives as an ordered list; the membership test runs once
+    per citation row, so it needs the set.
+    """
     log.info("Building citation DAG...")
+    backbone = set(backbone_dois)
     edges = set()
     for _, row in cit.iterrows():
         s = row["source_doi"]
         r = row["ref_doi"]
-        if s in backbone_dois and r in backbone_dois:
+        if s in backbone and r in backbone:
             edges.add((r, s))  # cited → citing
 
-    edges = list(edges)
+    edges = sorted(edges)
     log.info("Internal citation edges: %d", len(edges))
     return edges
 
@@ -292,8 +304,11 @@ def save_lineage_table(backbone_dois, lineage, positions, doi_meta, output_path)
             "y": round(y, 6),
         })
 
+    # kind="stable": the default quicksort would reorder the many ties in
+    # cited_by_count arbitrarily. A stable sort keeps them in the DOI order
+    # `rows` was built in, so the row order is a function of the data alone.
     lineage_df = pd.DataFrame(rows).sort_values(
-        ["lineage", "cited_by_count"], ascending=[True, False]
+        ["lineage", "cited_by_count"], ascending=[True, False], kind="stable"
     )
     lineage_df.to_csv(output_path, index=False)
     log.info("Saved lineage table -> %s (%d papers)", output_path, len(lineage_df))
--- a/scripts/figures/export_tab_venues.py
+++ b/scripts/figures/export_tab_venues.py
@@ -60,7 +60,11 @@ def main():
     acc_counts = acc["journal"].value_counts()
 
     rows = []
-    for j in set(eff_counts.index) | set(acc_counts.index):
+    # sorted(): the union is a set of journal-name strings, and `rows` order
+    # survives all the way to which venues `nlargest` keeps among ties below,
+    # so an unordered walk here put the hash seed into tab_venues.md — a
+    # git-tracked table the Œconomia manuscript includes (ticket 0591).
+    for j in sorted(set(eff_counts.index) | set(acc_counts.index)):
         ne = eff_counts.get(j, 0)
         na = acc_counts.get(j, 0)
         total = ne + na
@@ -73,7 +77,13 @@ def main():
             "total": total, "log_odds": lor,
         })
 
-    df = pd.DataFrame(rows).sort_values("log_odds", ascending=False)
+    # kind="stable" here and at the final sort below: log_odds ties are common
+    # (any two journals with the same efficiency/accountability counts share
+    # one), and the pandas default is quicksort, whose tie order is a numpy
+    # implementation detail. Stable sorting pins ties to the journal-name order
+    # `rows` was built in.
+    df = pd.DataFrame(rows).sort_values("log_odds", ascending=False,
+                                        kind="stable")
 
     # Assign lean labels
     def lean(lor):
@@ -107,7 +117,7 @@ def main():
 
     selected = pd.concat([eff_sel, shared_sel, acc_sel, key_rows])
     selected = selected.drop_duplicates(subset="journal")
-    selected = selected.sort_values("log_odds", ascending=False)
+    selected = selected.sort_values("log_odds", ascending=False, kind="stable")
 
     # Build markdown table
     lines = []
--- a/scripts/figures/plot_fig_traditions.py
+++ b/scripts/figures/plot_fig_traditions.py
@@ -146,7 +146,11 @@ def _draw_labels(ax, G, pos, trad_to_comm, comm_to_nodes, ref_counts,
     # Keep readable labels only; dedupe identical label texts
     # (e.g. two distinct references both rendered "Axel Michaelowa 2003"),
     # keeping the most-cited node.
-    candidates = sorted(label_nodes, key=lambda d: -ref_counts.get(d, 0))
+    # DOI is the tiebreaker, not decoration: `label_nodes` is a set, citation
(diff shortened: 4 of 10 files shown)


MinhHaDuong (commented):

Panel: correctness, consistency, scope, red team, doc propagation. All five ran against commit `8706ac7f`, each in a real checkout. **No blockers. Verdict: comment.** Individual verdicts: approve / comment / comment / approve / comment.

### What the panel confirmed with evidence

**The guard replays the defect on all four covered producers.** Red team reverted each fix individually in a scratch copy and reran: `list(edges)` in `plot_genealogy_html.py` → `test_html_renderer` fails with its own `fig_genealogy.html differs between hash seeds` assertion; same for the PNG; `sorted({...})` → `{...}` in `analyze_genealogy.py` → `test_lineage_model` fails with its own message; dropping the outer `sorted()` in `export_tab_venues.py` → `test_venue_table` fails with its own message. Four for four, all `AssertionError`, none an environmental failure. Correctness independently reproduced this by copying the test module onto an `origin/main` scratch worktree — all four red there, all four green here. This is the `feedback_red_test_the_guard_you_wrote` lesson satisfied properly, not asserted.

**The two-seed sample is not fragile.** Red team ran the mutated producers under 20 distinct `PYTHONHASHSEED` values on the actual fixtures: 20/20 distinct SHA-256 digests for both `analyze_genealogy.py` and `export_tab_venues.py`. The `SEED_A=1` / `SEED_B=12345` pair coincidentally agreeing is negligible at this fixture size.

**The seed override actually reaches the child.** `_run` builds `{**os.environ, "PYTHONHASHSEED": seed}` with the explicit key after the spread, and `source_root_env` does not clobber it — verified directly with `PYTHONHASHSEED=999` exported ambiently, and end-to-end by rerunning a mutated test with `PYTHONHASHSEED=0` exported in the parent shell (still correctly red). `BASH_ENV` cannot interfere: `subprocess.run([sys.executable, ...])` never spawns a shell. This was the highest-value attack available — a two-seed test that silently runs both halves under one seed is worthless — and it holds.

**No performance regression.** Every per-citation-row membership test in all three touched files uses a locally rebuilt `set(backbone_dois)`; the only list iterations are full O(n) walks that were already full walks on `origin/main`. Grepped all callers of `load_model` / `load_edges` / `assign_lineages` / `compute_layout` / `_infer_bands` / `build_citation_dag` repo-wide — no external consumer relies on the old set return type.

**The `kind="stable"` chains are total.** Both `rows` lists trace back to an already-ordered list or a `sorted()` union with no residual unordered input. `nlargest(n, col, keep="first")` preserves calling-frame row order for ties (verified empirically), so the `sort_values(stable)` → filter → `nlargest` → `concat` → `drop_duplicates(keep="first")` → `sort_values(stable)` chain is order-preserving by induction. Note for the record: plain `dict` iteration is insertion-ordered and never hash-seed dependent — the PR correctly targets only `set` walks.

**Docstring claims check out.** Doc propagation verified every added contract statement against the code, including the cross-module claim that `tests/test_determinism.py` "pins the seed to a single value" (`_smoke_env()` hardcodes `"PYTHONHASHSEED": "0"` — accurate). PR body test counts spot-checked and exact: `make lint` → 330 passed, `make check-fast` → 1399 passed. I reran the new module myself: 4 passed in 6.31 s.

**Sweep independently reproduced.** Scope re-ran the pattern hunt across all of `scripts/` and reached the same three files, checking every other hit against the PR's own stated exclusion criteria (membership-only sets, `sorted(set(...))` with no key, order-independent reductions, retired `archive_traditions/`, exploratory tools with no `--output`). `plot_genealogy_html.py:227`'s `sorted(backbone_dois, key=lambda d: d in popup_dois)` is fixed transitively by the upstream change in this same PR.

---

### Findings

**1. `verifiable:` Two tracked artifacts are stale against their fixed producer, not one.** [doc propagation, consistency, scope — converging]

The PR's "Not done" section names `tab_venues.md`. The Make graph gives it a dependent: `deliverables/manuscript/manuscript.mk:62` generates `tab_venues_fr.md` from `tab_venues.md`, and `paths.mk:37` puts it in `GIDE_INCLUDES`. Both are `git ls-files`-tracked and both feed a manuscript (`manuscript.qmd:331` `@tbl-venues`; the Gide variant via the FR file). The deferral itself is legitimate and correctly disclosed — this worktree holds only `.dvc` stubs, so regeneration is genuinely impossible here (independently confirmed: `data/catalogs/*.csv` absent). But with no CI in this repo, nothing regenerates them on merge, and the PR's own commit message calls this "a live science-lane defect, not a latent one." Worth a follow-up ticket to regenerate and byte-diff **both** files against the real corpus before the next manuscript render.

Confirmed clean by contrast: `tab_lineages.csv` and `fig_genealogy.{png,html}` are untracked (detracked in `ef36e871`), and no golden-hash fixture pins any of the six changed scripts (`tests/fixtures/smoke/golden_hashes.json` grepped clean).

**2. `verifiable:` A third tracked artifact carries the same risk and is not mentioned anywhere.** [doc propagation, scope]

`.gitignore:38` whitelists `deliverables/_shared/figures/fig_traditions_pre2008_citers.png`, it is `git ls-files`-tracked, it has a Make target in `network-limitations.mk`, and `deliverables/data-paper/revision-rdj26561/r1-14-network-response.md:38` points RDJ26561 reviewers at it as a potential electronic supplement. The PR fixes the same defect class in its producer (DOI tiebreaker on tied label nodes) but neither the PR body nor the test-module docstring lists it among artifacts whose content may shift on a corpus-present rebuild. Arguably higher stakes than `tab_venues.md` — it is referenced from a submission record. One line in the ticket or the follow-up would close this.

**3. `consider:` The traditions figures' top-5 truncation still has no tiebreak, upstream of the fix.** [red team, adjusted]

`plot_fig_traditions.py:129-131` does `sorted(comm_to_nodes[c], key=lambda d: -ref_counts.get(d, 0))` then `[:5]`; `other_nodes[:4]` at line 143 is the same shape. The DOI tiebreak this PR adds pins only the *final* label-dedup sort, downstream of both truncations. Red team tagged this `verifiable:` — I am downgrading it after checking the input: `comm_to_nodes` is a dict-of-lists built by `setdefault(c, []).append(doi)` over `partition.items()` (`scripts/_pre2007_traditions.py:157-159`), so it is insertion-ordered and **no set is involved** — the hash seed does not reach it. The residual is that truncation among tied `ref_counts` is deterministic but not *canonical*: it follows graph build order rather than a property of the data. Same argument the PR itself makes for sorting edges. Cheap to close (`key=lambda d: (-ref_counts.get(d, 0), d)` on both), but not a live defect.

**4. `consider:` `make determinism-check` does not run the new module.** [doc propagation]

`Makefile:864` wires only `tests/test_determinism.py` — the single-pinned-seed run-twice test. `tests/test_hash_seed_determinism.py`, the actual instrument for this defect class, reaches CI only through full-suite pytest collection, and this repo has no CI. The target literally named for this purpose skips the better test. Adding it to line 864 is a one-line change and would make the target mean what its name says.

**5. `consider:` The ticket asked for a shared helper and the PR grew the duplication instead.** [consistency]

`load_model()` and `load_edges()` in `plot_genealogy.py` and `plot_genealogy_html.py` are byte-identical apart from "figure"/"SVG" and "draw nodes"/"emit `<circle>` elements". The duplication predates the PR, but this PR pasted the same ~14-line docstring and the same fix into both copies. Ticket action 1 said "one shared helper if they can share," and the codebase has the precedent: `scripts/figures/_community_registry.py`, an underscore-prefixed shared module created for ticket 0307 for exactly this two-caller situation in `figures/`. Not extracting is defensible for a bugfix, but it is the ticket's own action item going unaddressed without a note.

**6. `consider:` Dissent worth recording on `compute_interpretation.py`.** [scope vs. doc propagation]

Both agents independently confirmed the PR's factual claim — not wired to any Make target, its declared output `tab_interp_2007_2011.csv` appears nowhere in the tree. They disagree on framing. Doc propagation: "correctly out of scope." Scope: the file carries the *identical* defect shape (`vocab = set(...) | set(...)` walked into `rows`, then `sort_values("abs_log_odds").head(top_n)` with no `kind="stable"`, so `.head()` truncation among ties follows the hash seed), and "reaches no deliverable" undersells that it is the same live bug held dormant only by nothing driving it. Both readings are correct; the PR body could say "dormant, same class" rather than "ruled out" so the next sweep does not rediscover it from scratch.

**7. `nofollow:` Makefile `PYTHONHASHSEED := 0` pin (line 58-60) unchanged.** [doc propagation, consistency] Still load-bearing — `scripts/_null_separation.py` and any Phase-2 script the sweep has not reached still rely on it. The comment reads as if the pin *is* the determinism mechanism, where after this fix it is a backstop for the unswept remainder. A provenance comment pointing at 0591 would match the PR's own style everywhere else, but nothing is wrong.

**8. `nofollow:` Marker discipline is correct, not a gap.** [consistency] `tests/test_script_hygiene.py:790-836` enforces `@integration` (not `@slow`) for any test file importing `subprocess`. Every test here goes through `_run()`. Measured: 4 passed in 6.65 s, slowest single test 3.29 s. No `slow` marker is owed and the fast-path budget ratchet is untouched.

**9. `nofollow:` `if d not in lineage` dedup has no reachable behaviour change.** [correctness, red team] Semantically equivalent to `d not in set(backbone_dois)` since `lineage[d]` is assigned unconditionally each iteration, and the check correctly precedes the assignment. A duplicate DOI in `tab_lineages.csv` would newly pin first-appearance order, but `analyze_genealogy.save_lineage_table` writes one row per DOI, so no such input exists under the current pipeline.

**10. `nofollow:` `build_citation_dag`'s return value is discarded in `analyze_genealogy.main()`** — called for its log-count side effect only. Pre-existing on `origin/main`, unrelated to this ticket.

---

### Recommendation

Merge. Findings 1 and 2 are the ones that matter and neither blocks: both are corpus-present regeneration work this worktree cannot do, and both are honest deferrals that just need a follow-up ticket naming all three tracked artifacts (`tab_venues.md`, `tab_venues_fr.md`, `fig_traditions_pre2008_citers.png`) rather than the one the PR body names. Findings 3, 4, and 5 are each a one-to-few-line change that could ride this branch or a follow-up, at the author's call.

The test design is the strongest part of this PR and survived the panel's hardest attacks intact. Running under two *different* seeds — where the Makefile's global pin and `tests/test_determinism.py` both pin one — is what makes it an instrument rather than a ritual, and the module docstring says so plainly. The uneven coverage is disclosed in the docstring rather than left silent, which is the right call.

PR #1282, opened 2026-07-28
data(0641): audit all 63 tracked Phase-2 artifacts against the pinned corpus

MinhHaDuong:
**Ticket:** tickets/0641-audit-all-tracked-phase-2-artifacts-for.erg

One systematic pass over every git-tracked Phase-2 artifact under
`deliverables/_shared/{tables,figures,_includes}/`, rebuilt from the corpus
that `dvc.lock` pins. Two live instances of this drift (0610, 0625) were found
without looking; this establishes the whole class.

## Method

- **Producers discovered from the Make database**, not a hand list: `make -n
  <target>` per tracked file, split on whether make printed a recipe or
  "nothing to be done". That distinction is load-bearing — 37 of the 63 files
  have no producer at all, and a naive "does make accept this target" check
  scores them all as buildable.
- **Rebuild**: `make -j4 NJOBS=6` on the pinned corpus (`make data` first).
  `PYTHONHASHSEED=0` and `SOURCE_DATE_EPOCH=0` are already exported by the
  Makefile, so determinism needed no extra handling. **No producer in the set
  touches the null-model chain** — nothing here cost hours of compute.
- **Clean means regenerated-and-identical.** Every rebuilt target's mtime was
  checked against the run window, so a target make silently skipped could not
  be scored clean.
- PNGs compared byte-wise, then pixel-wise, then column-profiled, to separate
  antialiasing from moved data.

## Classification

| Class | n | Members |
|---|---|---|
| **Clean** | 20 | `fig_composition`, `fig_composition_wide`, `fig_global_map_direct`, `fig_traditions_pre2008_citers`; `tab_citation_coverage_periods`, `tab_citation_verification`, `tab_corpus_flow`, `tab_corpus_sources`, `tab_dedup_error_estimates`, `tab_filter_ablation`, `tab_lit_confirmations`, `tab_network_limitations`, `tab_reference_counts`, `tab_reranker_validation`, `tab_retrieval_protocol.csv`, `tab_retrieval_protocol.md`, `tab_sem6_assignments`, `tab_semantic_robustness`, `tab_variables.md`, `tab_venues.md` |
| **Cosmetic** | 0 | — |
| **Substantive, left as committed** | 5 | `fig_bars_v1.png`, `fig_breaks.png`, `fig_global_map_cocitation.png`, `tab_null_separation_pre2007.csv`, `tab_pre2007_coverage.csv` |
| **No producer** | 37 | 34 `_includes/*.md`, `figures/scratch/scout_tradition_coupling.png`, `tab_venues_fr.md`, `tab_venues_fr_caption.md` |
| **Excluded** | 1 | `qa_cocitation_edges_report.json` |

Today's earlier merges verify clean: 0610 (`tab_venues.md`,
`fig_traditions_pre2008_citers.png`), 0625 (`tab_network_limitations.csv`),
0640 (venue captions). No regression from any of them.

`qa_cocitation_edges_report.json` was not touched — the author's decision today
is that it is a dated live-Crossref probe of record, and rerunning it is a new
experiment rather than a replay.

### Left as committed, with reasons

Five artifacts rebuilt differently, and all five were put back. Each is coupled
to something that makes republishing it an author call — and **not one of the
five couplings shows up in a grep of the citing prose for the artifact's
numbers.** Each was found by a different mechanism, which is the reusable part
of this audit.

- **`fig_bars_v1.png`** — A.2 declares `@fig-bars` frozen on the corpus as it
  stood at the first submission. **That it moved at all is the finding**:
  `plot_fig1_bars.py --v1-only` filters the live `in_v1` column of
  `refined_works.csv`, so the "frozen" figure is recomputed from a mutable
  corpus and four year-bars shifted. Its documented command no longer
  reproduces it. (`fig_composition.png`, the other figure A.2 calls frozen,
  reads the frozen `config/v1_tab_alluvial.csv` and rebuilt byte-identical —
  that is what a real freeze looks like.)
- **`fig_global_map_cocitation.png`** — the rebuild is a **regression, not a
  correction**. `config/community_registry.yml` keys hand-curated labels and
  colours on Louvain community *ids*; the corpus change repartitioned the graph
  and renumbered them. 7 of 9 registered communities lose their name and
  colour, and `plot_fig_global_map.auto_label` renders three as the literal
  string `nan nan` (the ticket 0550 NaN class). Nothing failed —
  `tests/test_global_map.py` checks the registry's *contents*, which stay true
  whatever the ids address. Committing it would have shipped a worse artifact.
- **`tab_null_separation_pre2007.csv`** — regenerated and committed, then put
  back. `tests/test_separation_provenance.py::test_a5_prose_matches_committed_csv`
  reads the six A.5 figures out of this CSV and asserts each appears in the
  manuscript's A.5 bullet. Its docstring states the intent: a regeneration that
  updates the CSV *and* the prose together stays green, and only a silent drift
  between the two turns it red. Committing the CSV alone was that drift, and
  the fix is not to rewrite the prose (`rules/git.md`, prose workpackages). The
  pair must land in one author-arbitrated PR.

  Found by: a **guard**. Grepping the prose for these numbers found nothing;
  the failing test found the coupling immediately.
- **`fig_breaks.png`** — `plot_fig2_breaks.py` draws a significance threshold
  at z = 1.5 and plots 2015 and 2021 as "not a break" markers; its docstring
  states the reading outright. On the pinned corpus the window-3
  Jensen–Shannon z for 2015 is **1.528** — above that line. So the regenerated
  figure shows the Paris marker sitting on the significance threshold, while
  `manuscript.qmd` L313 reads Paris as "a thematic diversification … rather
  than a rupture" and `.claude/rules/writing.md` codifies "Paris (2015) shows
  only a marginal Jensen–Shannon rise, not a rupture."

  The periodization's two breaks survive and strengthen (2007 cosine z 1.86,
  2013 JS z 1.66, both above threshold). What moves is the act II → III
  boundary claim, which is the author's to read.

  Found by: the **adherence pass**. This artifact has no guard and no quoted
  number — the load-bearing claim was compiled into the producer, as a
  hardcoded threshold and two hardcoded marker labels. Review then re-derived
  the table and found 2015 clears the line at **two** window widths, not one
  (w3 1.528, w4 1.594) — and at w4 by more than 2013 clears at w3. The figure
  plots w3 only.
- **`tab_pre2007_coverage.csv`** — n_works 1344 → 1654, works with an outgoing
  edge 407 → 667, share 30.3% → 40.3%. This one has no consumer of any kind:
  A.5 cites it qualitatively ("coverage of the early literature is sparse"),
  which 40% still supports. It was committed, and then reverted on review.

  `separation.mk` declares `separation: $(SEP_COVERAGE) $(SEP_NULL)`, and
  `compute_pre2007_coverage.py`'s docstring says its numbers "decide the
  interpretive regime for the separation null model". Committing one half of
  that pair while reverting the other puts A.5's evidence on two corpus
  snapshots — and `docs/audit-manuscript-stat-provenance.md` handoff item 2 is
  *already* an open, unresolved two-snapshot confusion about this same pre-2007
  graph (84/369 against 169/1,056, diagnosed as two `refined_citations`
  snapshots rather than a code bug). Manufacturing a second one to save a
  one-file diff is the wrong trade.

  Found by: a **Makefile pairing and a producer docstring**. Nothing in the
  prose, and no guard.

## Prose claims that move — author arbitration

`manuscript.qmd` A.5 quotes five values from `tab_null_separation_pre2007.csv`.

| Quantity | A.5 prose | Pinned corpus |
|---|---|---|
| pre-2007 co-citation graph | 84 nodes, 369 edges | 109 nodes, 466 edges |
| within-tradition co-citation share | 0.89 | 0.878 |
| configuration-model null | 0.37 | 0.353 |
| z | ≈ 25 | ≈ 28.1 |
| a-priori anchor nodes matched | "about three" | 7 (8 edges) |

A.2 licenses the drift in direction — these are declared current-corpus
diagnostics — so the fix is a transcription refresh, not a retraction. The
qualitative claims survive: the separation is still strong and still circular,
and the a-priori test still degenerates. The sentence "only about three
co-citation-core nodes match, all environmental economics" needs re-checking
against the seven-node result before it is restated.

Plus the `@fig-breaks` Paris reading above. Both route to ticket 0671.

**No submission record was touched.** Nothing under
`deliverables/data-paper/revision-rdj26561/` or `papiers/` changed, and every
artifact those files quote — `tab_network_limitations`,
`tab_dedup_error_estimates`, `tab_reference_counts`, `tab_lit_confirmations`,
`tab_semantic_robustness`, `tab_sem6_assignments`, `tab_variables`,
`tab_citation_verification`, `tab_filter_ablation`, `tab_retrieval_protocol`,
`fig_traditions_pre2008_citers` — came back clean.

## Two further findings, reported not fixed

- **`tab_venues_fr.md` / `tab_venues_fr_caption.md` have no producer.** They are
  hand-maintained French translations whose data body must be kept in sync with
  the generated `tab_venues.md` by hand. They match at HEAD; nothing would
  catch it if they stopped.
- **`tab_citation_verification.csv` is derived from a live-Crossref probe**
  (`qa_citations.py`, 300 links) — the same class the author froze for
  `qa_cocitation_edges_report.json`. It reproduced byte-identically this time,
  so no action was needed, but whether it should also be frozen is an author
  call.

## What the audit is worth, given it commits no artifact

The diff carries the ticket record and nothing else. That is the honest result,
and the ticket's exit criteria anticipate it — "match what their producers emit
from the pinned corpus, **or carry an explicit author decision to stay as
committed**".

Of 63 tracked artifacts, 20 are now *proven* reproducible against the pinned
corpus, 37 have no producer at all, 1 is excluded by the author, and all 5 that
moved are blocked on an author reading rather than on work this branch could
do. Every artifact quoted to the RDJ26561 referees came back clean. That is the
class established, which is what the ticket asked for.

The reusable lesson: **what makes an artifact prose-coupled is not always a
guard or a quoted number.** Five artifacts, four different detectors — a
failing test, a threshold hardcoded in a plotting script, an explicit freeze
claim in the appendix, and a Makefile pairing plus a producer docstring. A grep
of the citing prose for each artifact's numbers would have found none of them.
Three of the five surfaced only after the first attempt had already committed
them.

## Gates

`make lint` green (331 passed, 11 skipped). `make check`: 2313 passed, 43
skipped, 1 failed — `test_corpus_acceptance.py::test_reranker_cache_exists`,
pre-existing on `main` (ticket 0592) and unrelated to this diff.
`/verify-adherence` run pre-PR: mechanical phases all clean; its one blocking
semantic finding was the `@fig-breaks` Paris marker, now addressed by reverting
that artifact and routing it to 0671.

Follow-up tickets 0670 (community registry) and 0671 (manuscript corpus-dependent
claims) are filed in **PR #1283**, a tickets-only diff on the project's fast
path. They now also carry the audit's raw measurements — the registry-vs-corpus
community id tables and the full breakpoint z table — because that evidence
lives only in gitignored `data/derived/` in a throwaway worktree and would
otherwise be unverifiable short of a full pipeline re-run.

**Merge #1283 before this one.** This PR's `**Ticket:**` line closes 0641, and
0641's whole residue is 0670/0671; closing the tracker while its children sit
unmerged inverts the order.


MinhHaDuong (commented):
## `/review-pr` — five-perspective panel on #1282

Panel: correctness, red team, doc propagation, scope, consistency. All four
agents read the branch directly and re-derived the PR's claims rather than
taking the body's word. **Three verdicts: approve. One: request-changes.** The
dissent is preserved below and is an author call, not a mechanical defect.

### What verified true

Every specific, checkable claim in the PR body held up under independent
re-derivation. Worth saying plainly, because the body makes a lot of them:

- **The committed CSV is internally consistent.** `share_with_outgoing_edge` =
  667/1654 = `0.40326481257557434` exactly; `share_grey` = 2/1654 =
  `0.0012091898428053204` exactly; `n_academic + n_grey = n_works`. No
  hand-edited or half-regenerated values.
- **The 63-file tally closes.** `git ls-files deliverables/_shared/{tables,figures,_includes}`
  returns exactly 63. The classification rows sum to 63, the Clean row lists
  exactly 20 members, and every one of the 63 filenames appears in exactly one
  row — no double-count, no omission.
- **`z = 1.528` is real, not an assertion.** Re-read from
  `data/derived/tables/tab_breakpoints.csv`: 2015 `z_js_w3 = 1.5278490586809401`.
  2007 `z_cos_w3 = 1.855`, 2013 `z_js_w3 = 1.659`. `plot_fig2_breaks.py:57` does
  hardcode `ax.axhline(1.5, …)` and line 78 does hardcode the 2015/2021 markers.
- **The `nan nan` bug reproduces.** From the rebuild's own
  `global_map_cocitation.json`: community 4 → `['nan nan', 'nan nan']`, 1 →
  `['IPCC 2022', 'nan nan']`, 7 → `['Florian Weiler 2018', 'nan nan']`. Exactly
  three, as claimed. `analyze_global_map.py::member_label` uses the
  `or "?"` idiom that NaN walks straight through (the 0550 class).
- **"No numeric consumer" holds.** Repo-wide grep for `tab_pre2007_coverage` and
  for every old value (1344, 608, 407, 0.3028, 30.3, 1342, 0.00148) finds no
  consumer in any `.qmd`, test, or config. The three incidental hits are
  unrelated (a word count in `cut-plan.md:33`, a soft-membership mass in
  `docs/scout-tradition-coupling.md:68`).
- **The four reverts are genuine.** `git diff origin/main...HEAD` over
  `figures/` and `tab_null_separation_pre2007.csv` is empty — byte-identical to
  main, nothing staged and forgotten.
- **`tab_venues_fr.md` really does match.** Table body byte-identical to
  `tab_venues.md`; only the caption differs, translated.
- **"No producer" survives an adversarial check.** Spot-checked three
  `_includes/*.md` across different corners: `make -n` says nothing to be done,
  no Make rule targets them, and `grep -rn '_includes' scripts/ --include='*.py'`
  finds no writer. The one hit (`export_corpus_table.py:301`) is a stale comment
  — the script writes to `tables/`, not `_includes/`. This was the audit's
  highest-consequence classification and it is sound.
- **Gates reproduce.** `make lint` → 331 passed / 11 skipped. `make check` run
  fresh on this branch → **1 failed, 2313 passed, 43 skipped in 209s**, the
  single failure being `test_corpus_acceptance.py::TestFileExistence::test_reranker_cache_exists`
  (`llm_relevance_cache.csv missing at data/catalogs/`). Ticket 0592 is open and
  describes exactly that fresh-worktree failure, so the pre-existing attribution
  is accurate to the file. `erg check tickets/` PASS (407 tickets).

### Dissent — the one request-changes, preserved verbatim

**Red team, high confidence:**

> `compute_pre2007_coverage.py`'s docstring: its output "decide[s] the
> interpretive regime for the separation null model (`compute_null_separation.py`)."
> Both scripts take the identical Make prerequisites. […] These two tables are a
> documented pair, meant to be read from the same corpus snapshot.
>
> This PR commits one (`tab_pre2007_coverage.csv`: now 1654 works / 667 edges /
> 40.3%, current corpus) and reverts the other (`tab_null_separation_pre2007.csv`:
> stays at 84 nodes / 369 edges, old corpus). […] A.5 now cites two diagnostics
> computed on two different corpora, and nothing guards that pairing.
>
> The PR explicitly sets the precedent for the sibling table ("the pair must land
> in one author-arbitrated PR") — it should have applied the same rule to this one.

Verified: the docstring says exactly that (`compute_pre2007_coverage.py:8-9`),
and `separation.mk:20-28` gives both targets the same `$(REFINED)` /
`$(REFINED_CIT)` prerequisites. The two are a declared pair.

**The counter-case**, from correctness and doc propagation: no number from the
coverage table is quoted anywhere in `manuscript.qmd`, so no reader of the
rendered PDF sees an inconsistency. `manuscript.qmd:297` (A.2) does license
current-corpus appendix diagnostics generically, and reverting this file too
would reduce the PR's content diff to zero. The inconsistency is a repo-state
one, visible only to someone reading both CSVs — and ticket 0671 is where it
gets resolved either way.

**This is yours to decide.** Two coherent resolutions:

- **(a)** Revert `tab_pre2007_coverage.csv` too and land the whole A.5 evidence
  set — coverage + separation + prose — as one arbitrated PR under 0671.
  Consistent with the principle this PR states for its sibling.
- **(b)** Keep it committed, and add a line to 0671 recording that coverage is
  *already* on the pinned corpus, so the reconciliation must bring the
  separation table forward to match rather than treat both as pending.

Option (b) is the cheaper path and loses nothing, provided 0671 carries the
note. What should not happen is the asymmetry going unrecorded.

### Findings

`verifiable:` **The 2015 Paris marker crosses the threshold at two window widths,
not one.** The PR reports `z_js_w3 = 1.528`. From
`data/derived/tables/tab_breakpoints.csv`, 2015 also reads `z_js_w4 = 1.5940543`
— likewise above the `axhline(1.5)` at `plot_fig2_breaks.py:57`. Only `w2`
(1.186) stays below. That makes the `@fig-breaks` reading routed to 0671
somewhat more strained than the body states; worth carrying the w4 value into
0671 so the arbitration sees both.

`verifiable:` **The numeric evidence behind three of the four "left as
committed" decisions lives only in gitignored scratch.**
`data/derived/tables/tab_breakpoints.csv`, `global_map_cocitation.json`, and
`data/derived/t641_dryrun/` (63 `make -n` captures, one per audited artifact)
all exist in this worktree and all corroborate the body — but worktrees are
throwaway. Once pruned, "1.528" and "three communities render `nan nan`" become
unverifiable short of a full pipeline re-run, for numbers explicitly steering an
author decision. Suggest pasting the 2007/2013/2015 rows of `tab_breakpoints.csv`
and the three offending community labels into ticket 0671 / 0670 before the
worktree goes.

`consider:` **The four reverted artifacts have no freeze mechanism.** `make -n`
reports them up to date *right now*, but only because the reverting
`git checkout` stamped a later mtime than the leftover intermediates. On the
documented fresh-worktree path (checkout, then `make data`), the intermediates
would be newer and `make all` would silently regenerate all four back into the
degraded state — shifted bars, `nan nan` labels — with no guard to catch it.
`fig_composition.png` shows the real pattern: it is frozen because its rule
reads a committed input (`config/v1_tab_alluvial.csv`), not the live corpus.
This is 0670/0671 territory, not this PR's job, but the tickets should say so.

`consider:` **"Sparse" is a closer call than the body allows.**
`share_with_outgoing_edge` went 30.3% → 40.3%, a 33% relative jump, and the body
cites only the comfortable framing. The claim does survive — `median_refs_per_work`
is still `0.0` in both versions, and that is the stronger evidence for sparsity
— but `manuscript.qmd:321` ("coverage of the early literature is sparse") is
worth a one-line author gut-check in 0671, same class as the other qualitative
reads already routed there.

`consider:` **The ~88-line Result section may belong in `docs/`.**
`tickets/AGENTS.md` separates decisions (stay in the ticket) from the material
the decision was made about (natural location, referenced by path), and this
repo already has the convention: `docs/audit-manuscript-stat-provenance.md`,
`docs/audit-multi-output-scripts.md`, both tracked. A
`docs/audit-0641-phase2-artifacts.md` holding the 63-file classification table,
referenced from the ticket, would fit that precedent and leave the ticket to the
four decisions plus pointers. The counter-consideration is real — the audit
produced no other artifact, so the ticket is currently the only record — which
is why this is `consider:` and not a blocker.

`consider:` **The ticket log carries no completion entry.** Exit criteria flipped
to `[x]` and 88 lines of body added, but `--- log ---` still holds only the
creation line. Sibling ticket 0625, closed the same day in this same audit
lineage, used timestamped `note` entries for exactly this narrative. Not an
`%erg 0.1` violation and 0610 did the same thing, so the convention is not
unanimous — flagging for consistency within the lineage only.

`nofollow:` **Merge order.** 0670 and 0671 exist only on branch
`t670-file-0641-followups` (PR #1283, CLEAN); neither is on `origin/main`.
`erg-pr-merge` closes 0641 unconditionally on this PR's `**Ticket:**` line, so
merging #1282 first closes the tracking ticket while its four residual findings
sit in an unmerged PR. Both are CLEAN and #1283 is a tickets-only diff on the
fast path — merge #1283 first and this evaporates. Recorded, not a defect.

`nofollow:` **No evidence artifact for the 62 clean files.** Checkbox 1 ("every
tracked Phase-2 artifact rebuilt and diffed") rests on narrative for the 62
files that produced no diff. Spot-checks on three targets matched their claimed
classification, and an audit whose finding is "nothing changed" has nothing to
show by construction. Inherent to the ticket's shape; noted for future audits of
this form, which could commit the `make -n` capture set.

### Scope, consistency, history

Clean on all three. Exactly two files change; `git status --porcelain` is clean
with no tracked leftovers. The six-commit history is a documented
commit-then-revert, not churn — `35c358cc` regenerates three artifacts, then
`00d0306d` and `1ce3766d` each revert one with a commit body naming the specific
mechanism found (a failing provenance guard; a crossed hardcoded threshold).
That is exactly the shape `rules/git.md` permits. One `**Ticket:**` line,
resolving to a real file; 0670/0671 correctly mentioned in prose only, so this
PR will not close them. No ID collision — only two open PRs, queried
individually via `gh pr view --json files`, no overlap. CSV format matches
sibling tables (same `metric,value` shape, same full float repr, trailing
newline present).

### Panel verdict

**comment** — approve on the merits, with one author decision outstanding.

The audit's work is sound and its reporting is unusually honest: it discloses
that its own headline artifact class is unguarded, that a "frozen" figure is not
frozen, and that a rebuild it could have committed would have been a regression.
Nothing I could adversarially test came back false.

The single thing to settle before merge is the coverage/separation asymmetry
above — choose (a) or (b). Everything else is a `consider:` for 0670/0671 or a
merge-ordering nit. If you take (b), the only real ask is to get the scratch
numbers (`z_js_w3` **and** `z_js_w4` for 2015, the three `nan nan` community
labels) into the follow-up tickets before this worktree is pruned.

PR #1284, opened 2026-07-28
fix(0565): the generated source tables adopt Table 1's "Institutional reports"

MinhHaDuong:
**Ticket:** tickets/0565-datapaper-table2-grey-label-follows-rename.erg

Ticket 0334 renamed the curated-seed + World Bank layer to **Institutional reports** in the data paper's Table 1, Abstract, and prose. Table 2 and the deposited retrieval protocol kept printing **Grey literature** one page later, because both come from generators the corpus report also consumes. This closes the gap and pins it.

## Label decision: uniform, not parameterised

The ticket left open whether to rename everywhere or parameterise per document. The corpus report was read first, as it asks. It uses the old name in exactly two places — a row label in its own Discovery Sources table (`corpus-construction.md`) and the matching bullet lead — and nowhere defines or leans on "grey literature" as a term of art for that layer. Nothing resists the rename, so both consumers move. Parameterising would have bought two names for one thing plus a knob to keep in sync.

The corpus report's own table and bullet are renamed in the same pass. Renaming only the shared generator would have moved the contradiction rather than fixed it: the report would then have said "Grey literature" in §2 and "Institutional reports" in §3.

**What deliberately does not move:**

- Code identifiers: the `grey` source key, `from_grey`, `config/grey_sources.yaml`, `catalog_grey.py`, `grey_works.csv`.
- Prose where the World Bank repository harvest or the literature type genuinely applies — 0334's decision. Table 1's Coverage cell still reads "World Bank repository grey-literature harvest"; `citation-quality.md`, `metadata-quality.md`, and the data paper's reference-coverage sentences keep the generic use.
- The paper's title and the Zenodo retitle, out of scope per the ticket.

## Before / after, per label site

| Site | Before | After |
|---|---|---|
| `export_corpus_table.py` `SOURCE_META["grey"]["label"]` | `Grey literature` | `Institutional reports` |
| `export_retrieval_protocol.py` protocol row `Source` | `Grey literature` | `Institutional reports` |
| `export_retrieval_protocol.py` `GREY_CAPTION` | `The curated grey-literature seed list, in full` | `The curated institutional-reports seed list, in full` |
| `export_retrieval_protocol.py` seed-list heading | `### Curated grey-literature seed list` | `### Curated institutional-reports seed list` |
| `export_retrieval_protocol.py` `CAPTION` ×2 | `the grey seed list` / `The ISTEX and grey-literature languages` | `the institutional-reports seed list` / `The ISTEX and institutional-reports languages` |
| `_deposit_variables.py` `from_grey` gloss | `Provenance flag: grey-literature source` | `Provenance flag: institutional reports (curated seed + World Bank)` |
| `data-paper.qmd` §2.1 | `enumerates the curated grey-literature seed list` | `enumerates the curated institutional-reports seed list` |
| `corpus-construction.md` sources table | `\| Grey literature \| catalog_grey.py \|` | `\| Institutional reports \| catalog_grey.py \|` |
| `corpus-construction.md` bullet | `- **Grey literature:**` | `- **Institutional reports:**` |
| `README-datapaper.md` | `plus the grey-literature seed list` | `plus the institutional-reports seed list` |

Regenerated artifacts: `tab_corpus_sources.csv`/`.md`, `tab_retrieval_protocol.csv`/`.md`, `tab_variables.md`.

## Counts unchanged

Regenerated against the pinned corpus (`make data` in the worktree, 33,344 refined / 43,179 unified / 1,375,310 citation rows). **Every count is byte-identical** to the state ticket 0641's audit verified today; the artifact diff is labels and captions only. The institutional-reports row still reads `281, 210, 103, 0%, 100%, 95%, 32%, 97%`.

## The guard

`tests/test_datapaper_claims.py::test_generated_source_tables_use_table_1s_label` reads the label **off Table 1** rather than pinning a wording, so a future rename stays legal as long as every table moves together — per `writing.md`'s CI test-polarity rule. It keys on the "World Bank" fingerprint, which appears in all three tables in a different column each time, rather than on a column position or a source id.

It asserts the generator *and* its committed artifact for both tables: the generator alone would pass while a stale deposited CSV carried the old label, and the artifact alone would pass until the next `make`.

It deliberately does not read `tab_corpus_sources.md` — that file is gitignored, so a clean clone has none until a build with the corpus produces one, and it comes out of the same DataFrame as its CSV inside one `main()`, so it cannot drift alone. `tab_retrieval_protocol.md` *is* tracked and is checked.

### Red-test evidence

Each of the three label sites reverted independently on a scratch copy, then restored:

| Mutation | Guard output |
|---|---|
| `export_corpus_table.py` label → `Grey literature` | `AssertionError: Table 1 calls the layer 'Institutional reports' but export_corpus_table labels it 'Grey literature'` |
| `tab_corpus_sources.csv` row → `Grey literature` | `AssertionError: tab_corpus_sources.csv labels the layer 'Grey literature', not Table 1's 'Institutional reports' — run make corpus-tables and commit` |
| `export_retrieval_protocol.py` row → `Grey literature` | `AssertionError: Table 1 calls the layer 'Institutional reports' but the retrieval protocol labels it 'Grey literature'` |

Plus four in-suite fangs on the new helpers: the rename path, the ambiguous-table refusal (two World Bank rows), picking the protocol table out of a file that also holds the seed enumeration, and drift between paper and generated table.

`tests/test_retrieval_protocol.py`'s three pinned expectations follow (`~344/419/433`). The caption-provenance guard now looks for `institutional-reports` rather than `grey` in the disclosure, so it still fails if the disclosure drops the row.

## Verification

- `make lint` — 336 passed, 11 skipped, exit 0.
- `make check` — 2312 passed, 49 skipped, **1 failed**: `test_corpus_acceptance.py::TestFileExistence::test_reranker_cache_exists` (`llm_relevance_cache.csv` absent). Pre-existing and main's, ticket 0592 — this branch touches no harvest, corpus, or DVC file.
- `/verify-adherence` — **PASS**, no mechanical failures. Its three semantic nits were all fixed in `c0f28f5` except one left deliberately: `corpus-construction.md`'s framing sentence "academic and grey literature" is a generic literature-type use that 0334 explicitly preserves, and pinning a positive phrasing there is forbidden by the polarity rule.


MinhHaDuong (commented):

Panel: correctness, consistency, scope, red team, doc propagation (all `sonnet`, fresh contexts, PR head `c0f28f5`).

**Verdict: comment — no blockers.** Correctness returned *approve*; the other four returned *comment*. The rename is complete where it counts, the guard is well-built and genuinely red-tested, and the invariant holds: `git diff origin/main...HEAD -- deliverables/_shared/tables/` touches only label and caption text — every count byte-identical, no reordering, no CSV quoting change.

Independently re-run in the PR worktree: `make lint` → **336 passed, 11 skipped**; `make check-fast` → **1431 passed, 9 skipped**; `tests/test_datapaper_claims.py tests/test_retrieval_protocol.py` → **53 passed**.

Correctness red-tested all five label sites on scratch copies (never touching tracked files) and confirmed each revert to `Grey literature` flips the comparison — the guard's fangs are real. Red team confirmed `export_retrieval_protocol.py` regenerates byte-identical output from config alone, and that the seed-list table's own World Bank row does **not** leak into `world_bank_row_label` (the exact-header match in `pipe_table_rows` holds).

---

### Three sites the sweep missed — all one-line, all verified

**1. `verifiable:` `scripts/figures/export_retrieval_protocol.py:75`** — the `_cell()` docstring still reads "17 bibliographic titles from the **grey seed list**", five lines below the `CAPTION` and `GREY_CAPTION` this PR renamed to "institutional-reports seed list". Developer-facing only, but it is the one internal inconsistency left inside a file the PR otherwise swept clean. *(consistency, high)*

**2. `verifiable:` `config/grey_sources.yaml:1`** — header reads `# Curated grey literature for climate finance HET analysis`; line 2 already says `# These are key institutional reports`, so the rename is trivially safe. This is reader-visible in the deposit: `build/build_datapaper_archive.sh:49` ships the whole tree via `git archive HEAD`, and `data-paper.qmd` §2.1 points referees at exactly these comments — "so the query is readable without reading the code". A referee following that pointer from the "Institutional reports" row lands on a file that calls itself grey literature. *(doc propagation, verified here)*

**3. `verifiable:` `build/build_datapaper_archive.sh:79`** — comment reads "the paper points referees here for the query fields, per-tier term counts, and the **grey-literature enumeration**", describing precisely the artifact this PR renamed to "institutional-reports seed list". Same logical unit as `build/templates/README-datapaper.md`, which the PR *did* update. (Line 96's `grey` loop token is a code identifier — correctly untouched.) *(verified here)*

---

### Dissent — surfaced verbatim, author's call

The PR body defends two sites as deliberate. The doc-propagation reviewer disagrees at high confidence. Both readings are defensible; recording the split rather than resolving it.

**4. `data-paper.qmd:73` — Table 1's row describes itself in its own Coverage cell as what its label no longer says.**

```
| Institutional reports | Hybrid (curated seed + World Bank API) |
  17 selected reports [e.g., @buchner2013] + World Bank repository grey-literature harvest |
```

PR body: kept per 0334 — the World Bank repository harvest genuinely *is* a grey-literature harvest. Doc propagation: "self-contradiction inside one row," and one the new guard is structurally blind to, since `world_bank_row_label` checks only that such a row exists, never that a row's text agrees with its own label.

**If you do rewrite that cell, the literal string "World Bank" must survive it.** Red team's reproducer: strip it (e.g. generalizing to "multilateral repository harvest") and the new guard fails with `AssertionError: expected exactly one row mentioning the World Bank harvest, got set()` — a message that blames the wrong thing on an edit that has nothing to do with labels.

**5. `deliverables/_shared/_includes/corpus-construction.md:16`** — "The corpus assembles academic and grey literature from …", three lines above the row this PR renamed. PR body: generic literature-type use, and `writing.md`'s polarity rule forbids pinning a positive phrasing there. Doc propagation: high-confidence contradiction, and it flows into `corpus-report.qmd` too.

---

### Fragilities in the new guard — no current trigger

**6. `consider:` Three parsers now read the same `#tbl-sources` table.** `sources_table_labels` (L78), the new `tbl_sources_rows` (L136), and the pre-existing private `_tbl_sources_rows` (L245) each carry a near-identical anchor regex and return a different shape. Flagged independently by correctness, consistency, and scope. Nothing fails today; a future change to Table 1's markup has to land in three places, and the new public helper differs from the old private one by an underscore while doing something genuinely different — easy to misread. A `/simplify` pass could collapse them.

**7. `consider:` `tbl_sources_rows`'s regex can span an unrelated earlier table.** `r"\n\| *(Source *\|.*?)\n\n: [^\n]*\{#tbl-sources\}"` under `re.DOTALL` anchors on the *first* `\n| Source |` and runs non-greedily to the *first* `{#tbl-sources}` caption. Insert any other `Source`-headed pipe table above Table 1 and the match swallows both tables plus the prose between them — and unlike `pipe_table_rows`, this helper has no `set(line) <= set("|:- ")` filter, so delimiter and text lines leak into the row dicts silently. Red team's reproducer yields either a wrong label or `AssertionError: … got {'Example', 'Institutional reports'}`. Only one such table exists today.

**8. `consider:` `pipe_table_rows` silently drops an all-placeholder data row.** `set(line) <= set("|:- ")` classifies any line built solely from those characters as a delimiter, so a legitimate row like `| - | - | - | - | - |` vanishes with no error. Found independently by correctness and red team; no such row exists in current data.

**9. `consider:` `from_grey` is now the only `from_*` variable whose gloss does not echo its identifier.** `tab_variables.md:31` glosses it "institutional reports (curated seed + World Bank)"; the parenthetical does most of the bridging work, but the deposit has no note anywhere stating that the `grey` identifier family and the "Institutional reports" label name one thing. `scripts/_deposit_variables.py:109`'s `source` enum still lists the literal `"grey"`.

---

### Out of scope — recommend a follow-up ticket

**10. `nofollow:` Two other deliverables narrate this same layer with the old term, unguarded.**

- `deliverables/manuscript/manuscript.qmd:291,293` — "a curated grey-literature seed together with the World Bank repository"; "Including the grey literature is deliberate…"
- `deliverables/agentic/agentic-paper.qmd:29,148` — "grey literature repositories"; "OpenAlex, ISTEX, bibCNRS, grey literature, teaching canon"

Both describe the layer this PR renames elsewhere. Outside ticket 0565's declared scope (data paper + corpus report), and manuscript prose is co-edited with the author rather than changed autonomously — so this belongs in its own ticket, not this PR.

**11. `nofollow:` `scripts/_deposit_schema.py:51`** — `DATASET_TITLE` / `DESCRIPTION` / `KEYWORDS` in the shipped `datapackage.json` still say "grey literature". These track the paper title, which the ticket puts out of scope alongside the Zenodo retitle. Same follow-up.

**12. `nofollow:` `tests/test_retrieval_protocol.py`'s `for token in ("istex", "institutional-reports", "asserted")`** is a literal positive pin, the fragility `writing.md`'s polarity rule warns against — but the rule is scoped to `test_manuscript_prose.py`, the pattern already pinned `"grey"` before this PR, and it was merely carried forward. Not introduced here.

---

### Scope and completeness — clean

Every hunk maps to a ticket Action. The `REPO`/`TABLES` path constants are a minimal enabling refactor for the new guard, not a drive-by. The 191 test lines are proportionate: three tables × generator-and-artifact, plus one file holding two adjacent pipe tables to disambiguate. Both exit-criterion checkboxes have a concrete satisfying change. The `sys.path.insert` of `scripts/figures` matches an established pattern in nine other test files with no basename collision. `tab_corpus_sources.md` really is gitignored (`git check-ignore` confirms), so the guard's stated reason for not reading it holds.

The guard reads the label off Table 1 rather than pinning a wording — correct under the polarity rule, and it is what makes a future rename legal so long as every table moves together.

**Recommendation:** fix 1–3 (three one-line comment/docstring edits), rule on 4–5, and open a follow-up ticket for 10–11. 6–9 are worth a note in the ticket log; none needs to block.

MinhHaDuong (commented):
Second review cycle (raid orchestrator, on behalf of the /review-pr panel — correctness, consistency, doc-propagation): label-drift guard correctly fingerprints the renamed row with red/green fang tests; rename applied consistently, no stale instances; numeric columns byte-identical (labels/captions only — frozen-corpus constraint holds). One doc flag checked and confirmed out-of-scope per ticket 0565: scripts/_deposit_schema.py 'Grey Literature' belongs to the author-level Zenodo retitle decision. Nothing blocking. Verdict: approve.

PR #1287, opened 2026-07-28
fix(0362): a corpus re-run refuses to publish dvc.lock instead of pushing to main

MinhHaDuong:
**Ticket:** tickets/0362-run-corpus-pipeline-sh-auto-commits-and.erg

## What

`scripts/run_corpus_pipeline.sh` ended by branching, committing `dvc.lock`,
merging into `main` and running `git push origin main` whenever `dvc.lock` was
the only changed file. That bypassed the branch-and-PR gate every other change
in this repo goes through, and it removed the checkpoint pending corpus changes
rely on: a config or code edit that alters the corpus was executed *and*
published by the next `make corpus`, with no human confirmation.

The tail now delegates to the new `scripts/dvc_lock_gate.sh`, which prints the
`dvc.lock` diff, names the branch + merge-request recipe, cites ticket 0362,
and exits 1. The pre-existing "files other than dvc.lock changed" warning path
is carried over unchanged (still exit 0).

No `--publish` opt-in flag: the raid plan chose the conservative resolution,
and an opt-in is the same hole behind a flag.

## Why a separate file

`run_corpus_pipeline.sh`'s tail is unreachable in a throwaway repo — the
padme / dvc / GROBID / main guards fire first. Extracting the gate is what
makes the behaviour testable rather than only greppable.

## Test

`tests/test_dvc_lock_gate.py` (`@pytest.mark.integration`, 4 tests) drives the
gate as a **real subprocess** against a throwaway git repo, per the bash rules'
never-source-in-shell discipline. The dvc.lock case asserts the behaviour —
exit non-zero, `main` does not advance, no branch created, the change left in
the working tree, the message names `dvc.lock` and cites 0362 — and only
secondarily greps the pipeline script for `git push origin main`.

Red first (commit `test(0362)`), then green.

## Gate

- `make check-fast` — rc 0
- `make lint` — 330 passed, 12 skipped, rc 0
- `make check` (diff touches `scripts/`) — 2262 passed, 8 failed. All 8 are
  `test_corpus_acceptance` / `test_corpus_flow` missing-corpus failures: this
  worktree has not run `make data`. Environmental, unrelated to the diff.
- `/verify-adherence` — **PASS**, no mechanical or semantic findings. `/gaze`
  may skip the mechanical phase.

## Invariants respected

Corpus v2 is frozen: the pipeline was not run and no DVC state was touched.
The warning path was not weakened.

## Reported, not ticketed

The `sed 's/^...//'` single-file detection is carried over unchanged and would
misparse a rename entry (`R  old -> new`). Pre-existing, out of scope, and
below the tooling-lane severity floor — the failure mode is falling into the
WARNING path, which is safe.

---

## Review rounds (3, all fixes applied)

**Round 1** — stale docs the behaviour change orphaned: `README.md:45`,
`Makefile:183,190`. Plus `git diff HEAD` (a staged `dvc.lock` printed an empty
diff), the no-`--publish` decision moved into the script header, and tests for
both.

**Round 2** — the refusal trigger **widened**, deliberately: from "`dvc.lock`
is the only changed file" to "`dvc.lock` is among the changed files". The
reviewers reproduced the hole — a modified `dvc.lock` plus any stray untracked
file, which a real `dvc repro` can leave behind, fell through to the exit-0
warning path, exactly the case the checkpoint exists for. The refusal now lists
the other changed files. This strengthens rather than weakens the warning path
the ticket protects: a change set without `dvc.lock` still takes it,
byte-identical. Also fixed a stale docstring in `test_makefile_contract.py`.

**Round 3** — verdict **approve**, 4/5 approve, no blockers. Red team ran 12
git states against throwaway repos and found no fail-open. Its one actionable
finding is applied: `docs/revision-runbook.md` scenario D now says the exit 1
is the expected outcome of a corpus-changing run, not a broken pipeline.

## Known gaps, disclosed not fixed

- A rename entry (`R  old -> dvc.lock`) is misparsed and routes to the WARNING
  path. Pre-existing parsing, commits nothing either way.
- `dvc push` still runs before the gate, so a refusal leaves the DVC remote
  ahead of `origin/main`'s lock until the human lands the PR. Pre-existing; the
  blobs are additive and unreferenced, so nothing reads a wrong value. Worth a
  follow-up ticket if the raid lead judges it above the severity floor.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,14 +180,16 @@ all: manuscript papers
 # Padme is the data authority. Run the pipeline on padme, pull on doudou.
 #
 # On padme (pipeline run):
-#   make corpus                      # dvc repro + push + commit lock
+#   make corpus                      # dvc repro + dvc push (never commits)
 #
 # On doudou (sync only):
 #   make corpus-sync                 # git pull + dvc pull
 #   make figures && make manuscript  # Phase 2 + 3 (no DVC needed)
 
 # Full pipeline — run on padme only (GPU, API access).
-# After dvc repro + push, auto-commits dvc.lock if it's the only change.
+# After dvc repro + push, this target NEVER commits. If dvc.lock changed it
+# prints the diff and exits 1 — land the lock via branch + merge request
+# (ticket 0362).
 corpus:
 	bash scripts/run_corpus_pipeline.sh
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,8 @@ uv run dvc cache dir /data/projets/dvc-cache/oeconomia
 **Padme is the data authority.** The corpus pipeline (Phase 1) runs on padme. Doudou only pulls data — never pushes.
 
 ```bash
-# On padme — build corpus (auto-pushes data, auto-commits dvc.lock):
+# On padme — build corpus (pushes data to the DVC remote; never commits.
+# A changed dvc.lock exits 1 — land it via branch + merge request):
 make corpus
 
 # On doudou — sync and build:
--- a/docs/revision-runbook.md
+++ b/docs/revision-runbook.md
@@ -75,7 +75,10 @@ Reviewer asks to include 2025 data or additional sources.
 
 1. Create ticket branch: `t{N}-corpus-expansion-{description}`
 2. Update `config/corpus_collect.yaml` (year_max, new queries)
-3. Run full pipeline: `make corpus` (on padme — GPU, API access)
+3. Run full pipeline: `make corpus` (on padme — GPU, API access). It exits 1
+   by design once `dvc.lock` changes — that refusal is the expected outcome of
+   a corpus-changing run, not a failure. Review the printed diff, then commit
+   `dvc.lock` onto the ticket branch yourself (ticket 0362).
 4. Verify: `make corpus-validate` (acceptance tests)
 5. Freeze new v2 config: `config/v2_*`
 6. Regenerate all figures: `make figures`
--- a/scripts/dvc_lock_gate.sh
+++ b/scripts/dvc_lock_gate.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+# dvc_lock_gate.sh — report what a corpus re-run changed; never publish it.
+#
+# Ticket 0362. This used to be the tail of run_corpus_pipeline.sh, where it
+# branched, committed dvc.lock, merged into main and `git push origin main`.
+# That bypassed the branch-and-PR gate every other change in this repo goes
+# through, and it removed the checkpoint a pending corpus change relies on:
+# a config or code edit that alters the corpus would be executed *and*
+# published by the next `make corpus`, with no human confirmation.
+#
+# The refusal lives in its own file so it can be driven directly by a test —
+# the pipeline script's own guards (padme, dvc, GROBID, main) make its tail
+# unreachable in a throwaway repo.
+#
+# Exit status: 0 when nothing changed or when files other than dvc.lock
+# changed (the pre-existing warning path, unchanged); 1 when dvc.lock changed,
+# so `make corpus` fails loudly and the operator commits it via branch + PR.
+#
+# Usage: bash scripts/dvc_lock_gate.sh   (run from the repository root)
+
+set -euo pipefail
+
+changed=$(git status --porcelain)
+
+if [ -z "$changed" ]; then
+    echo "dvc.lock unchanged, nothing to commit."
+elif [ "$(echo "$changed" | sed 's/^...//')" = "dvc.lock" ]; then
+    echo ""
+    echo "REFUSING to publish: the pipeline re-run changed dvc.lock."
+    echo ""
+    git --no-pager diff -- dvc.lock || true
+    echo ""
+    echo "dvc.lock is left uncommitted in the working tree. Publishing a corpus"
+    echo "change is a human decision (ticket 0362): review the diff above, then"
+    echo "land it on a branch and open a merge request —"
+    echo ""
+    echo "    git switch -c data-dvclock-\$(date +%Y%m%d)"
+    echo "    git add dvc.lock && git commit -m 'data: update dvc.lock after pipeline re-run'"
+    echo "    git push -u origin HEAD && gh pr create"
+    echo ""
+    exit 1
+else
+    echo ""
+    echo "WARNING: files other than dvc.lock changed:"
+    echo "$changed"
+    echo "Stage and commit manually."
+fi
--- a/scripts/dvc_lock_gate.sh
+++ b/scripts/dvc_lock_gate.sh
@@ -28,10 +28,19 @@ changed=$(git status --porcelain)
 
 if [ -z "$changed" ]; then
     echo "dvc.lock unchanged, nothing to commit."
-elif [ "$(echo "$changed" | sed 's/^...//')" = "dvc.lock" ]; then
+elif echo "$changed" | sed 's/^...//' | grep -qxF 'dvc.lock'; then
+    # Refuse when dvc.lock is *among* the changed files, not only when it is
+    # the sole one. A real `dvc repro` can leave a stray untracked file behind,
+    # and the old "is the only change" test then downgraded the refusal to the
+    # exit-0 warning path — the one case the checkpoint exists for.
     echo ""
     echo "REFUSING to publish: the pipeline re-run changed dvc.lock."
     echo ""
+    if [ "$(echo "$changed" | sed 's/^...//')" != "dvc.lock" ]; then
+        echo "Other files changed too — review them before committing anything:"
+        echo "$changed"
+        echo ""
+    fi
     # Against HEAD, not the index: an already-staged dvc.lock would otherwise
     # print an empty diff under "review the diff above".
     git --no-pager diff HEAD -- dvc.lock || true
--- a/scripts/dvc_lock_gate.sh
+++ b/scripts/dvc_lock_gate.sh
@@ -12,6 +12,10 @@
 # the pipeline script's own guards (padme, dvc, GROBID, main) make its tail
 # unreachable in a throwaway repo.
 #
+# There is deliberately no --publish opt-in flag (ticket 0362, action 3): an
+# opt-in is the same hole behind a flag, and the operator's next step is two
+# git commands the message spells out.
+#
 # Exit status: 0 when nothing changed or when files other than dvc.lock
 # changed (the pre-existing warning path, unchanged); 1 when dvc.lock changed,
 # so `make corpus` fails loudly and the operator commits it via branch + PR.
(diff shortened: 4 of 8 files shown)


MinhHaDuong (commented):
## `/review-pr` — four-perspective panel (correctness, consistency, red team, doc propagation)

**Verdict: comment** — no blockers. The diff meets ticket 0362's exit criteria: no branch, no commit, no push anywhere in the new path; a `dvc.lock`-only change exits 1 with the diff and a two-command recipe; the pre-existing "other files changed" warning path is carried over byte-identical. Exit code propagates losslessly (the gate is the last statement of `run_corpus_pipeline.sh`) and no caller in the repo chains off `make corpus`, so the new non-zero exit breaks no automation. All 6 tests in `tests/test_dvc_lock_gate.py` pass locally (0.28 s) and 3 of them would fail against plausible broken implementations (substring match, diff-against-index, still-commits).

### Findings

**verifiable: stale docstring — `tests/test_makefile_contract.py:54`** *(consistency + doc propagation, independently, high confidence)*
`"The pipeline logic (guards, dvc repro, auto-commit) lives in a standalone bash script"` — auto-commit is exactly what this PR removed. False documentation sitting in the test file that asserts on the target this PR changed. One-line fix (e.g. `guards, dvc repro, and the publish gate`).

**verifiable: the DVC remote is already published before the gate runs** *(red team, high confidence)*
`run_corpus_pipeline.sh` does `dvc repro || ret=$?`, then **unconditionally** `dvc push`, then checks `ret`, then calls the gate. The blobs are on the remote — pushed even when `dvc repro` failed — before any human sees the refusal. So the checkpoint gates the *git pointer*, not the *data*. Ticket 0362's framing ("a pending corpus change cannot be executed without an explicit human step") is only half-satisfied in practice. Pre-existing, not introduced here, and arguably the pointer is the thing that matters for reproducibility — but the gap is real and deserves a follow-up ticket rather than silence.

**verifiable: a single stray untracked file downgrades the refusal to a warning (exit 0)** *(red team, high confidence, reproduced)*
```
$ echo "changed: true" >> dvc.lock; echo stray > newfile.txt; bash dvc_lock_gate.sh
WARNING: files other than dvc.lock changed:
 M dvc.lock
?? newfile.txt
EXIT: 0
```
A real `dvc repro` can leave an incidental untracked file (log, un-gitignored report), which silently flips `make corpus` from "fails loudly" to "succeeds with a warning". **Explicitly out of scope**: this `elif`/`else` structure is byte-identical pre- and post-PR, and the ticket says "do not weaken the existing warning path". Flagged for awareness; the tighter rule would be *"dvc.lock is among the changed files"* → refuse, rather than *"is the only changed file"*.

**consider: `git diff HEAD -- dvc.lock` prints nothing when the gate runs from a subdirectory** *(red team, reproduced)*
`git status --porcelain` is root-relative so the refusal still fires correctly, but the diff pathspec resolves against cwd and matches nothing — masked by `|| true`, so "review the diff above" refers to an empty region. Not reachable through the pipeline (which `cd`s to `PROJ_ROOT` and passes an absolute path); only bites a standalone invocation. `-- :/dvc.lock` would close it.

**consider: `build/templates/README-datapaper.md:77`** describes `make corpus` as "runs the full DVC pipeline" with no mention of the publish semantics. Not wrong, just silent; it is a template for spun-off repos, arguably out of scope.

**consider: `.claude/rules/architecture.md:181-183`** still frames the dvc.lock-via-PR expectation as convention; the gate now enforces it mechanically. A sentence, not a blocker.

**nofollow:** the `sed 's/^...//'` rename misparse — already declared in the PR body, pre-existing, fails safe into the WARNING path.
**nofollow:** no rename/submodule/porcelain-quoting exposure — `dvc.lock` is a plain regenerated path and the comparison is exact-string.
**nofollow:** shebang, `set -euo pipefail`, header style, `@pytest.mark.integration`, real-subprocess discipline and `tests/test_<module>.py` naming all match repo convention; `script-io.md` does not apply to a bash gate.

### Dissent
None — the four perspectives converged. Correctness read findings 2 and 3 as "outside the targeted failure mode, fine to leave"; red team read finding 2 as worth a follow-up ticket. Both agree it is not this PR's defect. Author's call.

### Suggested action before merge
Fix the `test_makefile_contract.py:54` docstring inline (one line, in scope — it documents the very behaviour this PR removed). The unconditional `dvc push` deserves its own ticket; the untracked-file downgrade can ride along with it.

MinhHaDuong (commented):

**Verdict: approve.** Correctness, Consistency, Scope, and Red team all returned *approve*; Doc propagation returned *comment* on two pre-existing doc gaps this diff makes newly relevant. No blockers.

Gate re-run in a clean worktree on this branch: `make check-fast` rc 0 · `make lint` 330 passed / 12 skipped · `tests/test_dvc_lock_gate.py` + `tests/test_makefile_contract.py` 29 passed.

### Red team — the fail-open question, answered with experiments
The reviewer copied `scripts/dvc_lock_gate.sh` into throwaway repos and exercised clean tree, `M`, `D`, `A`, `MM`, `??`, `UU`, `R`, `sub/dvc.lock`, a trailing-space filename, outside-a-repo, and a repo with no commits.

**No fail-open found.** Every state where `dvc.lock` genuinely changed and porcelain emits an exact `XY dvc.lock` line refuses with exit 1, leaves `main` unmoved, creates no branch, and leaves the change in the working tree.

- `nofollow:` the rename entry (`R  old -> dvc.lock`) routes to the exit-0 WARNING path instead of REFUSE — reproduced, and exactly the defect the PR body already discloses. Not a publish bypass: neither path commits or pushes. Below the tooling-lane floor, as filed.
- `consider:` in a repo with **no commits**, `git diff HEAD` leaks a `fatal: bad revision 'HEAD'` line into the refusal message. Cosmetic; exit code unaffected. (confidence: high)
- `nofollow:` `dvc push` runs *before* the gate, so corpus data can reach the DVC remote while the lock is refused. Pre-existing, unchanged by this diff, and consistent with the documented design.

### Correctness
Gate detects `dvc.lock` across the porcelain states that matter; `set -euo pipefail` interactions are sound (`grep -q` in a condition, `|| true` on the diff); the gate is the last statement so its status propagates. No other Make target depends on `corpus`, so exit 1 fails only `make corpus`. A mutation dropping `exit 1` or restoring the auto-commit block would be caught by `test_dvc_lock_gate.py`. Ticket exit criteria met.

- `consider:` no test covers a **deleted** `dvc.lock`; the red team confirmed the behaviour is correct, but it is unpinned. (confidence: medium)

### Consistency
Shebang, `set -euo pipefail`, header style, and executable bit match sibling `scripts/*.sh`. `@pytest.mark.integration` is the right tier (real subprocess, per the bash never-source-in-shell rule). Repo-wide grep for `auto-commit` / `auto-push` / `housekeeping-dvclock-` found no survivors outside `.venv`. No registration gap — the gate is called only from `run_corpus_pipeline.sh`, and `test_script_hygiene.py` scopes to Python. Nothing found.

### Scope
All three of ticket 0362's actions are addressed, including the explicit decision against a `--publish` opt-in. Extraction into a separate file is justified, not speculative: the pipeline tail is unreachable in a throwaway repo because the padme/dvc/GROBID/main guards fire first. 136 test lines for a data-safety gate is proportionate.

- `consider:` the refusal trigger widened from *`dvc.lock` is the sole change* to *`dvc.lock` is among the changes*. Direction is strictly more conservative and well-commented in the script, so the ticket's "do not weaken the warning path" invariant holds for the branch body — but the *dispatch into it* changed, which "Invariants respected" slightly elides. Worth naming as a deliberate decision rather than an implementation detail. (confidence: medium)

### Doc propagation — the one open item
- `verifiable:` `docs/revision-runbook.md`, Scenario D "Corpus expansion": step 3 says *Run full pipeline: `make corpus` (on padme)* and step 9 jumps to *PR → main*, with nothing between them about the new exit-1 refusal or the manual branch/commit/PR step. This is the operator doc most likely to be read *before* running `make corpus` for a real corpus change, and it does not warn that step 3 now fails by design. Suggest a one-line insert. (confidence: high)
- `consider:` the ticket's Verification line "`make corpus` still succeeds end to end on padme" reads as exit-0, but the shipped design intentionally exits 1 in the common case. A one-line log annotation would keep the record honest. (confidence: medium)
- `nofollow:` `docs/plan-makefile-modularization.md:55` has stale line numbers for the `corpus` target — pre-existing drift, not introduced here.

Neither doc finding is a regression from this diff. The runbook line is worth patching before the next corpus-expansion cycle hits it cold — in this PR or a follow-up, author's call.

MinhHaDuong (commented):
## Agent C review — PR 1287 (ticket 0362)

**Verdict: comment (no blockers found)**

### Correctness
`dvc_lock_gate.sh` refuses (exit 1) whenever `dvc.lock` appears among `git status --porcelain` entries, prints the diff against HEAD, and leaves the tree uncommitted — verified by reading the code and by driving `tests/test_dvc_lock_gate.py` (29/29 pass, including the "sibling untracked file" case that would have downgraded the old logic). `run_corpus_pipeline.sh` no longer contains `git push origin main` or auto-commit/branch/merge logic; it delegates entirely to the gate.

### Consistency
Follows `coding-bash.md`: `set -euo pipefail`, subprocess-based tests (no source-in-shell), tab-delimited pitfalls not applicable here. Script I/O rules (script-io.md) don't apply — this is a control script, not a data-producing one.

### Scope
Diff stays within the ticket's stated blast radius: `Makefile` (comment only), `README.md`/`docs/revision-runbook.md` (doc sync), `scripts/dvc_lock_gate.sh` (new), `scripts/run_corpus_pipeline.sh` (tail replaced), plus matching tests. No unrelated changes.

### Red-team
Reproduced both gaps the PR body discloses, by direct testing in a throwaway repo:
- **verifiable:** a *renamed* dvc.lock (`git mv old dvc.lock`) produces `git status --porcelain` output `R  dvc.lock.old -> dvc.lock`; after `sed 's/^...//'` this is `dvc.lock.old -> dvc.lock`, which does **not** match `grep -qxF 'dvc.lock'`, so the gate falls through to the exit-0 WARNING path instead of refusing. Confirmed by direct repro (not present in the test suite). Low real-world likelihood — `dvc repro` rewrites `dvc.lock` in place rather than renaming it — and it's disclosed in the PR body, so **consider:** add a test pinning this known gap (or a `git status --porcelain` parse that also matches on the `-> dvc.lock` suffix) rather than leaving it only documented in prose.
- **verifiable:** `dvc push` runs unconditionally before the gate is invoked (`run_corpus_pipeline.sh` lines ~87-89), so pipeline-produced data reaches the DVC remote even when the gate then refuses to let `dvc.lock` land on main. Confirmed by reading the script; disclosed in the PR body. **nofollow:** pushing data without the lock pointing at it is a pre-existing property of the old script too (data was always pushed before the git dance), and reverting a `dvc push` isn't meaningfully different in risk from the status quo — not a regression this PR introduces.
- Deletion of dvc.lock (`rm dvc.lock`) is correctly caught (status `D` → sed strips to `dvc.lock` → refuses). Staged-only and mixed untracked+dvc.lock cases are covered by the existing tests and behave correctly.

### Doc propagation
`README.md` and `docs/revision-runbook.md` both replaced "auto-pushes/auto-commits" language with the exit-1 refusal behavior and cite ticket 0362. No other doc references the old auto-commit behavior (grepped CLAUDE.md, AGENTS.md — no hits).

### Housekeeping
`make check-fast` equivalent (`pytest tests/test_dvc_lock_gate.py tests/test_makefile_contract.py`) passes 29/29.

PR #1289, opened 2026-07-28
fix(0530): publish the ISTEX query as a code span so it survives rendering

MinhHaDuong:
## What

`deliverables/_shared/tables/tab_retrieval_protocol.md` publishes the ISTEX
Boolean query verbatim so a referee can reproduce the harvest. Pandoc's
`smart` extension — live in the `markdown` reader Quarto uses, absent from
`gfm` — rewrote its straight quotes as typographic ones:

```
SOURCE   : "climate finance" OR "finance climat" OR "finance climatique"
RENDERED : “climate finance” OR “finance climat” OR “finance climatique”
```

ISTEX does not read `“` as a phrase delimiter, so a reader who copied the
query out of the published table ran one that matches nothing. The
reproducibility claim the table exists to support was false as published.
This is the 0325 defect class: a published recipe that does not work when
copied.

Ticket 0376 recalibrated the escaper from `gfm` to the real reader and caught
three of the four extensions that differ (`@`, `~`, `^`). `smart` is the
fourth, and it needed its own judgment rather than a reflex escape.

## The fix, and the two options not taken

The ISTEX `Query terms` cell is wrapped in a backtick code span at the emitter
call site (`VERBATIM_CELLS`, a set of `(Source, column)` pairs). A code span
suppresses `smart` inside it and tells the reader the value is verbatim.
Escape first, wrap second — the order `export_corpus_table.py` already uses
for the TOTAL row's emphasis, because `_cell` escapes the backtick and
wrapping first would turn the span's own delimiters into literal characters.

Not escaping `smart`'s characters in `markdown_text_cell`: unlike `@ ~ ^`,
`smart` produces *correct* output. Curly quotes and en-dashes in a journal
name or a caption are what a rendered document should have, so a global
escape would degrade every other table to fix one cell. Not disabling `smart`
in the Quarto config either — same objection across ten documents, with the
manuscript at submission stage. The distinction is semantic (a value a reader
executes versus prose) and per-cell, which the escaper cannot see, so the
emitter declares it where the values are known.

The construction lives in `markdown_verbatim_cell`, on the shared escaper
surface. The first draft wrapped `markdown_text_cell` output in backticks at
the call site instead, arguing that one call site does not earn a third
function — and that was wrong for a reason the YAGNI framing could not see.
Getting a code span right for an arbitrary value needs the escaper module's
own knowledge of what survives inside one, and no precondition the emitter
could state would enforce it. Two review rounds found two instances of exactly
that: a backslash-escaped backtick closes the span early, and a
backslash-escaped pipe publishes the backslash. So the ticket's option 2 is
right after all, on evidence the first pass did not have.

## Tests

Three, all driving the real emitter (`build_protocol_rows` / `render_markdown`)
rather than a hand-built frame, so they cover the path that produces the
shipped file.

- `test_published_istex_query_renders_verbatim` (integration, needs pandoc) —
  renders through `tests/_qmd_render.py` and asserts the cell reads back
  byte-equal. **Red before the fix.**
- `test_every_quoted_cell_is_a_code_span` (fast) — the class guard. Keyed on
  the distinguishing feature (a straight quote outside a code span), not on
  the ISTEX row, so a query added for a sixth source inherits it. **Red before
  the fix.**
- `test_prose_cells_keep_smart_typography` (integration) — the control, green
  throughout. A curated report title's apostrophe must still read back curly.
  Without it the verbatim guard would go green on an oracle that had silently
  lost `smart` (the `gfm` regression 0376 fixed) while the shipped table
  stayed broken.

`scripts/_markdown_table.py`'s "Which reader" docstring now names the fourth
extension and why it is deliberately not escaped there.

## Verification

- Regenerated: CSV byte-identical, `.md` differs in exactly the one intended
  cell (old-vs-new diff is one line). The wrap is Markdown-only, so no number
  and no CSV consumer moves — corpus v2 stays frozen.
- `make check-fast` exit 0; `make lint` exit 0; `/verify-adherence` **PASS**
  (mechanical phase can be skipped on the next `/gaze`).
- Full `make check` run (diff touches `scripts/`): 8 failures, all
  `tests/test_corpus_acceptance.py` / `test_corpus_flow.py` reporting
  `data/catalogs/*` absent — the known worktree-without-`make data` class,
  not reachable from a config-only emitter.

**Ticket:** tickets/0530-smart-typography-rewrites-a-published-query.erg

### DIFF
--- a/deliverables/_shared/tables/tab_retrieval_protocol.md
+++ b/deliverables/_shared/tables/tab_retrieval_protocol.md
@@ -5,7 +5,7 @@ All API queries are bounded to publication years 1990--2024. Terms, concept grou
 | Source | Retrieval | Query fields | Query terms | Languages |
 |:---|:---|:---|:---|:---|
 | OpenAlex | Four-tier keyword search (co-occurrence filter, T3 2 / T4 3 of 4 concept groups) | default.search (title, abstract, indexed fulltext) | 52 (T1 15, T2 25, T3 10, T4 2) | 8: Arabic, Chinese, English, French, German, Japanese, Portuguese, Spanish |
-| ISTEX | Boolean phrase search | ISTEX default index | "climate finance" OR "finance climat" OR "finance climatique" | English, French |
+| ISTEX | Boolean phrase search | ISTEX default index | `"climate finance" OR "finance climat" OR "finance climatique"` | English, French |
 | Institutional reports | Curated seed list plus World Bank repository API | seed identifiers; repository full-text search | 17 curated reports, 3 API queries | English |
 | UNFCCC key documents | Curated seed list (config/unfccc_sources.yaml) | document symbol | 232 seed documents | English |
 | OECD DAC key documents | Curated seed list (config/oecd_dac_sources.yaml) | document symbol | 35 seed documents | English |
--- a/scripts/_markdown_table.py
+++ b/scripts/_markdown_table.py
@@ -16,7 +16,9 @@ Three functions, because the callers hold three different input contracts:
 
 ``markdown_cell``
     The input **is** Markdown — a curated description authored in-repo, whose
-    backticks are intentional code spans. Used by the deposit codebook.
+    backticks are intentional code spans. Written for the deposit codebook; no
+    live caller under ``scripts/`` today, and its rule is pinned by
+    ``tests/test_variables_table.py`` (see the ``_MARKDOWN_CODE`` note below).
 ``markdown_text_cell``
     The input is plain text with no markup intent — a journal name out of the
     bibliographic corpus, a source label edited by hand, a language code the
@@ -75,7 +77,9 @@ import re
 # the #1289 review caught it. Left in place rather than fixed: ``markdown_cell``
 # is the only user and has no live caller today, so changing its output would be
 # an unverifiable edit to a dormant path. Corrected here so the next reader does
-# not inherit the same false premise a third time (ticket 0530).
+# not inherit the same false premise a third time (ticket 0530). Whoever does
+# fix it should know that ``tests/test_variables_table.py`` currently *pins* the
+# escaped form, so the guard moves with the behaviour.
 _PROSE = {"\\": r"\\", "|": r"\|"}
 
 _MARKDOWN_TEXT = str.maketrans(_PROSE)
--- a/scripts/_markdown_table.py
+++ b/scripts/_markdown_table.py
@@ -63,10 +63,19 @@ which describes neither the Lua reader nor the extension set it selects.
 
 import re
 
-# The pipe is escaped everywhere; the reader honours ``\|`` inside a code span
-# too, so that one rule holds throughout. The backslash needs the opposite
-# treatment on each side of a span boundary — CommonMark reads it as an escape
-# in prose but literally inside code — so prose and code are escaped separately.
+# In prose the pipe is escaped and the backslash with it, because CommonMark
+# reads a backslash as an escape there.
+#
+# ``_MARKDOWN_CODE`` escapes the pipe *inside* a span, and that rule is wrong —
+# measured, 2026-07-28, against the reader above: ``a \| b`` inside a span
+# renders the backslash literally, and a raw ``a | b`` inside a span parses as
+# one cell, because pandoc's pipe-table splitter respects spans. The comment
+# here used to claim the reader "honours ``\|`` inside a code span too", and
+# that claim is what ``markdown_verbatim_cell``'s first draft inherited before
+# the #1289 review caught it. Left in place rather than fixed: ``markdown_cell``
+# is the only user and has no live caller today, so changing its output would be
+# an unverifiable edit to a dormant path. Corrected here so the next reader does
+# not inherit the same false premise a third time (ticket 0530).
 _PROSE = {"\\": r"\\", "|": r"\|"}
 
 _MARKDOWN_TEXT = str.maketrans(_PROSE)
@@ -155,16 +164,30 @@ def markdown_verbatim_cell(text: str) -> str:
     what a backtick wrap at the call site violates. Nothing in the emitter can
     enforce the precondition, so the construction belongs here.
 
-    Inside a span the only escape that survives is ``\\|``, which the reader
-    honours. A backtick is handled the way CommonMark provides for: the fence
-    is one longer than the value's longest backtick run, and a value that
-    begins or ends with one is padded, since the reader strips a single leading
-    and trailing space from a span's content.
+    **Nothing is escaped**, and that is the whole point of a span: CommonMark
+    processes no backslash escape inside one, so an escape does not protect the
+    value, it *becomes* the value. Measured, because the module comment above
+    asserted the opposite and the first draft of this function inherited it —
+    ``pandoc -f markdown`` renders ``a \\| b`` inside a span as the literal
+    ``a \\| b``, and renders a raw ``a | b`` inside a span as ``a | b`` in one
+    cell. The pipe needs no escape here: the reader's pipe-table splitter
+    respects code spans, so the span itself does the work the backslash does in
+    prose. Escaping it would publish a backslash in the middle of a regex a
+    reader is meant to run (caught by the #1289 review panel, second round).
+
+    A backtick is therefore handled structurally, the way CommonMark provides
+    for: the fence is one longer than the value's longest backtick run, and a
+    value that begins or ends with one is padded, since the reader strips a
+    single leading and trailing space from a span's content.
+
+    Runs of whitespace are collapsed, as everywhere else in these emitters — a
+    pipe-table row is line-delimited, so a newline in the value would end the
+    row outright and no construction can hold it.
 
     Never raises, for ``markdown_text_cell``'s reason: the input is data, not
     markup this repo authored.
     """
-    value = " ".join(str(text).split()).translate(_MARKDOWN_CODE)
+    value = " ".join(str(text).split())
     if not value:
         return ""
     runs = re.findall(r"`+", value)
--- a/scripts/_markdown_table.py
+++ b/scripts/_markdown_table.py
@@ -12,7 +12,7 @@ wrong one: it normalises *corpus* text (encoding artifacts, DOIs, language
 codes) and pulls in pandas, ftfy and langdetect for it. Escaping markup is a
 different concern with no dependencies at all.
 
-Two functions, because the callers hold two different input contracts:
+Three functions, because the callers hold three different input contracts:
 
 ``markdown_cell``
     The input **is** Markdown — a curated description authored in-repo, whose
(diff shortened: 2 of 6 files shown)


MinhHaDuong (commented):
## `/review-pr` — four-perspective panel (correctness, red team, consistency+scope, doc propagation)

**Verdict: comment** — the shipped fix is correct and well-evidenced. Three findings concern the *general mechanism* the PR introduces (`VERBATIM_CELLS`) rather than the live value, and one of the four reviewers rated the first as blocking. Dissent preserved below; the author decides.

The panel independently reproduced the red: with pandoc 3.1.3, an unwrapped straight-quoted cell renders curly, a code-span-wrapped one stays straight, and a sibling prose apostrophe still curls. Both new integration tests and the fast class guard are credible reds. CSV separation is correct **by construction**, not just by the reported diff — `main()` builds the CSV from `build_protocol_rows()` via `save_csv`, never touching `_pipe_table` (`export_retrieval_protocol.py:261`).

### Findings

**1. `verifiable:` (red team, high) — a backtick in a future `VERBATIM_CELLS` value corrupts the span, a failure mode strictly worse than the bug being fixed.**
`_cell` escapes a literal backtick to `` \` ``, correct *outside* a span. `_pipe_table` then wraps the already-escaped string in backticks, and pandoc does not honour `\` inside a code span — it reads the first literal backtick as the closing delimiter. Reproduced:

```
$ printf '| a |\n|:---|\n| `the \`x\` value` |\n' | pandoc -f markdown -t plain
  a
  ----------------
  the \x` value`
```

The span ends mid-value and the remainder leaks into the cell as raw Markdown. No live instance — the ISTEX query has no backtick — but `VERBATIM_CELLS` is a set that invites a second entry, and nothing enforces the precondition. Suggested: assert in `_pipe_table` (or a fast unit test) that a verbatim value contains no backtick, so the next entry cannot ship silently corrupted.

**2. `verifiable:` (correctness, high — DISSENT: this reviewer rated it request-changes) — the class guard's cell splitter is not pipe-aware, unlike its sibling 137 lines above it in the same file.**
`test_every_quoted_cell_is_a_code_span` splits with `line.strip("|").split("|")`, while `_separators()` (`tests/test_retrieval_protocol.py:480`) exists in the same module precisely because that is wrong for a cell containing `\|`:

```
line = '| "a \| b" | prose |'
naive split → ['"a \\', 'b"', 'prose']   # boundary corrupted
```

A future value combining a quote and a pipe — exactly the class the guard's own docstring claims to cover ("so a query added for a sixth source inherits it") — can be mis-split and missed. Dormant today. Cheap fix: unescape (or reuse the `_separators` logic) before splitting.

**3. `verifiable:` (both, medium) — escape-then-wrap leaves a literal backslash in a "verbatim" value containing a pipe.**

```
$ printf '| a |\n|:---|\n| `q \| r` |\n' | pandoc -f markdown -t plain
  q \| r
```

The two reviewers agreed on the behaviour and split on the reading: correctness called it a narrower verbatim guarantee than the docstrings imply; red team called it acceptable because the value survives readably. Either way it is worth one line of docstring caveat. This tension is inherent to the ticket's own option 1, not an oversight in the implementation.

**4. `consider:` (three reviewers, converging) — the class guard keys on `"` only, not `'` / `--` / `...`.**
A future verbatim value using `--` (a CLI flag) or `...` with no quote is rewritten by `smart` and passes the guard silently. The ticket's blast-radius scan found no live data-cell hit for those characters, so this matches the ticket's measured scope — but the guard's docstring ("no future executable value may land in a smart cell") oversells what it covers. Suggest either widening the character class or narrowing the docstring.

**5. `consider:` (scope) — the ticket's Verification item asks for all nine Markdown targets regenerated and byte-compared; the PR body evidences only `tab_retrieval_protocol`.**
Expected (only this table's source changed, and the diff touches no other generated file), but the narrative should say "regenerated all nine, only this one changed" rather than omit the other eight.

### Cleared, with evidence

- `nofollow:` **No second unfixed copy of the query.** Grep for `finance climatique` / `climate finance" OR` across `deliverables/*.qmd`, `papiers/**/*.tex`, `docs/*.md` found only unrelated book-pitch text. The sole consumer, `deliverables/data-paper/data-paper.qmd:84`, references the table by path and does not quote the query inline. (This repo's known "containment guard blind to second file" class does not bite here.)
- `nofollow:` **No PDF layout risk.** `tab_retrieval_protocol.md` is a standalone deposit artifact, never `\input`ed into the manuscript, so the lengthened cell cannot cause an overfull hbox in the paper.
- `nofollow:` **Renamed-`Source` scenario is covered.** The class guard scans emitted Markdown text independently of `VERBATIM_CELLS`, so a stale key (e.g. `ISTEX` → `Istex`) still trips it as long as the value carries a `"`.
- `nofollow:` **Makefile is not drifted.** `Makefile:342-343` already lists `scripts/_markdown_table.py` as a prerequisite of the grouped `.csv .md &:` target; one invocation writes both outputs.
- `nofollow:` **Docstring is technically accurate.** `pandoc --list-extensions=markdown` → `+smart`; `gfm` → `-smart`. No prose-rule violations found (checked against the LLM-tell and Elements-of-Style guard list).
- `nofollow:` **Test tiering correct.** `export_retrieval_protocol.py` pulls `pandas` but none of the conftest's declared heavy deps, so the unmarked fast-tier guard is right.
- `nofollow:` **Escape-then-wrap is consistent** with `export_corpus_table.py`'s TOTAL-row emphasis, including its empty-cell guard. Declining a third `markdown_verbatim_cell` helper for one call site matches the repo's lean-methods preference.
- `nofollow:` **No unrelated changes.** The diff touches exactly the four files the ticket names.

### Note on tests

The panel did not re-run `make check` — the PR body reports `make check-fast` and `make lint` at exit 0 and eight pre-existing `test_corpus_acceptance` / `test_corpus_flow` failures from `data/catalogs/*` being absent in a worktree without `make data`, which is the known class and not reachable from a config-only emitter. Verification above is pandoc-level reproduction plus static analysis of the branch.

MinhHaDuong (commented):
## `/review-pr` — five-agent panel

**Verdict: approve with comments.** No blockers. `make check-fast` exit 0, `make lint` exit 0 (338 passed, 12 skipped) on `90317bd5`. All 68 tests in the two touched test files pass, including six pandoc-driven integration tests (pandoc 3.1.3 present).

The panel independently confirmed the core of the fix, empirically rather than by reading:

- **Red before green** (correctness): reverting `_markdown_table.py` and `export_retrieval_protocol.py` to `main` while keeping the new tests fails collection on `ImportError: markdown_verbatim_cell` — the tests genuinely depend on the fix.
- **The span construction is CommonMark-correct** (red team, 34 tool calls of real `pandoc -f markdown`): single backtick, all-backticks, mixed runs `` a`b``c```d ``, max-run-at-edge, backslash-terminated `C:\path\`, pipe `(alpha|beta)+`, `@cite`/`~sub~`/`^sup^`/`$math$`/`[link](x)`, HTML entities, NBSP/tab/newline collapse, `:---`, leading/trailing `|`, 500-char value — every one round-trips byte-for-byte. The `smart` defect was reproduced live and confirmed suppressed inside the span.
- **Ticket invariants hold** (scope): `@ ~ ^` escaping untouched, `markdown_text_cell` body untouched, `_pipe_table` substitutes only for the one `VERBATIM_CELLS` pair. `git diff main...HEAD -- deliverables/` is exactly one file, one line.
- **CSV is clean** (doc propagation): `save_csv` runs off the raw rows; `markdown_verbatim_cell` is reachable only from `_pipe_table`.
- **Makefile prerequisite is correct**: `Makefile:340` already lists `scripts/_markdown_table.py` under the `tab_retrieval_protocol.csv .md &:` grouped target, so an escaper edit does trigger a rebuild.
- **No downstream render risk**: `data-paper.qmd:84` only *names* this table in prose — unlike its four siblings it is not `{{< include >}}`'d, so there is no LaTeX pipe-table column for a monospace span to overflow.

### Findings

**`verifiable:` `tests/test_markdown_table.py:9` is now stale** — *consistency + doc propagation, high confidence, both agents independently.* The module docstring reads "the boundary between the two — the reason there are two functions", while `scripts/_markdown_table.py:15` was correctly updated to "Three functions" in this same diff. Introduced by this PR, one-line fix.

**`verifiable:` the module docstring contradicts its own new comment** — *consistency, high confidence.* `scripts/_markdown_table.py:19` still says `markdown_cell` is "Used by the deposit codebook"; line 75, added by this PR, says it "has no live caller today". Verified: `grep -rn "markdown_cell" --include=*.py` returns no importer under `scripts/` — only `tests/test_variables_table.py`, `tests/test_markdown_table.py`, `tests/test_markdown_table_shape.py`. The "no live caller" claim is true and the docstring line is the false one. Both regions were edited in this commit without reconciling them.

**`consider:` the class guard covers one emitter, not the class** — *red team, medium confidence.* `test_every_quoted_cell_is_a_code_span` scans only `export_retrieval_protocol.render_markdown()`. Five sibling emitters (`export_core_venues_markdown.py`, `export_corpus_flow.py`, `export_corpus_table.py`, `export_language_table.py`, `export_tab_venues.py`) also call `markdown_text_cell` and are unscanned, while the docstring frames the guard as "no future executable value may land in a smart cell". The ticket's own blast-radius scan found no live offender in those five, so this is a coverage gap, not a live defect — but the docstring's scope claim outruns the implementation. Either narrow the docstring or widen `_pipe_tables` to iterate the emitters.

**`consider:` the `smart` rationale is restated in four places** — *consistency, medium confidence.* Module docstring, module-level comment, `markdown_verbatim_cell`'s docstring, and the `VERBATIM_CELLS` comment in `export_retrieval_protocol.py` each carry a near-verbatim version of the same reasoning (plus the test-class docstrings). Not a correctness issue; a drift risk against the module's own stated philosophy of recording a thing once.

**`nofollow:` `_MARKDOWN_CODE`'s wrong pipe rule, left documented-but-unfixed.** Verified documentation-only: `_PROSE`, `_MARKDOWN_TEXT` and `markdown_cell`'s body are byte-identical before and after. The deferral reasoning (dormant path, unverifiable edit) is sound and the panel does not contest it. Worth recording for whoever picks it up: `tests/test_variables_table.py:314` and `:318` currently *pin* the behaviour the new comment declares wrong (`markdown_cell("a \`x | y\` b") == r"a \`x \\| y\` b"`), so the eventual fix has to move those assertions, not just the code.

**`nofollow:` the guard regex `` (`+)(?:(?!\1).)*?\1 `` diverges from pandoc's maximal-run matching** on a hand-crafted sub-run case. Both correctness and red team tried to weaponize it — a 20-combination sweep plus targeted constructions — and neither could make it hide a real straight quote. Unreachable from any emitter path today, since `markdown_text_cell` always emits an even count of literal backticks and `markdown_verbatim_cell` always sizes its fence minimally.

### On proportionality

The scope reviewer was asked to weigh the 190 added test lines against this repo's standing "permitted is not proportionate" lesson, and puts this squarely on the guard-is-warranted side: that lesson is scoped to *prose* claims with no data signature, where the only instrument is a source-text pattern match. This defect has a hard render-level signature — byte-identity of the rendered cell against the source value, through the real reader — and `test_prose_cells_keep_smart_typography` is the control that stops the guard from going vacuously green if the oracle ever loses `smart` again, which is the exact `gfm` regression 0376 fixed. The three tests cover three distinct defect modes, two of which were found by earlier review rounds on this same PR. No padding found.

The panel also endorses the PR's reversal on option 2. The evidence is in the diff rather than the narrative: two concrete call-site-wrap failures (backslash-escaped backtick closing the span early, backslash-escaped pipe publishing the backslash) each carry a dedicated regression test, and neither is a precondition the emitter could have stated.

